PR #233, opened 2012-11-12
BUGFIX Prevent unbounded recursion in admin/pages

benrwhite:
One of our clients was editing their site tree in this way and caused us to discover and fix this bug, so we are sharing the love:
## To Reproduce:

WARNING: this bug results in unbounded recursion, so ensure you have sensible memory/stack limits
1. Enter the site tree, create and publish a page heirarchy with one the child of the other.
2. Change the staged versions of these pages such that the child page is now the parent page.
3. Attempt to view the cms page /admin/pages, this is where the bug results.
## Resolution

The branch to pull fixes the problem by ensuring that a list of handled IDs is kept, so generating the sitetree won't attempt to render an infinite heirarchy.

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -1132,7 +1132,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 
 					if($groupedByParent) {
 						$groupedByParent = array_diff_key($groupedByParent,$handledIds);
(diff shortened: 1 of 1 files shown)


tractorcow:
This has been fixed in 4.x with our refactor of batch permission checking.

tractorcow:
Also fixed in 3.x with https://github.com/silverstripe/silverstripe-cms/commit/51164768751de4e2c7c931d21f5635714df7bf34

PR #281, opened 2013-01-29
Explicitly added routes (following removal of auto-routing controllers)

chillu:
See https://github.com/silverstripe/sapphire/pull/1142


tractorcow:
I should point out, since these are added AFTER $URLSegment route, none of those will ever match their respective controllers; They'll all just be handled by ModelAsController :)

PR #881, opened 2013-10-23
NEW: Page types no longer require their own controller (fixes silverstripe/silverstripe-framework#1905)

lozcalver:
They will instead inherit the Controller from their parent class.

https://github.com/silverstripe/silverstripe-framework/issues/1905

### DIFF
--- a/code/controllers/ModelAsController.php
+++ b/code/controllers/ModelAsController.php
@@ -16,9 +16,16 @@ class ModelAsController extends Controller implements NestedController {
 	 * @param string $action
 	 * @return ContentController
 	 */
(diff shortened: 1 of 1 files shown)


dhensby:
Sorry to resurrect this but I think this breaks rendering hierarchy because the viewer looks at the current controller instead of the current model. So if you don't need a controller for functionality, you still need it for rendering...

kinglozzer:
@dhensby I believe that was fixed - https://github.com/silverstripe/silverstripe-cms/pull/914

dhensby:
@kinglozzer yep, looks like it was ;)

PR #905, opened 2013-12-02
VirtualPages: Allow whitelisting methods and relations

jedateach:
... that should be called on the referenced page, and not the VirtualPage’s methods/relations.

This is particularly useful for getting images to display that relate to the referenced page.

The whitelist is a just a config of relations or function names, as they are defined in Page/Page_Controller.

eg:

```yaml
---
Name: app
After: "*"
---
VirtualPage:
  virtual_methods:
    - Image1
    - Image2
    - HeroImage
```

### DIFF
--- a/code/model/VirtualPage.php
+++ b/code/model/VirtualPage.php
@@ -417,9 +417,8 @@ class VirtualPage extends Page {
 	public function __call($method, $args) {
 		if(parent::hasMethod($method)) {
 			return parent::__call($method, $args);
-		} else {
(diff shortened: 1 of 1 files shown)


tractorcow:
Sorry, closing, no response in 3 years @jedateach :D

jedateach:
:laughing:

tractorcow:
I'm going to fix this soon anyway... see the linked issue at https://github.com/silverstripe/silverstripe-cms/issues/1386 which I've made to collect duplicates of this problem.

PR #1052, opened 2014-07-16
FIX so that OldPageRedirector::find_old_page uses ORM call

dmeeking:
Cleaning out a TODO item in OldPageRedirector so that we now use ORM syntax to find old versions of pages.


dhensby:
This looks good to go in

dhensby:
merged with ef46dee37f9bd3483729c348cafab0e8104c38d0

PR #1074, opened 2014-08-09
API: Extract siteconfig out to an external module.

wilr:
Keep CMSSettingsController as from experience this is a common instance to 'hook' into updates for. Not sure whether to throw a deprecated warning if people try to add_extension to it. That's probably the most useful approach.

Site config is now at https://github.com/silverstripe/silverstripe-siteconfig


anjaewool6656-png:
사탄들아 악마에 자식들아 인간이길 포기한자들아 하루빨리 객사하길 비나이다 생체실험은 사람한테 하는게 아니고 쥐한테 하는것이다

PR #1091, opened 2014-08-22
BUG: Fixes #3422 revealing controller info via media extensions

kmayo-ss:
Fixes a bug where appending a media file extension to a non-existent sub url results in controller info being leaked

https://github.com/silverstripe/silverstripe-framework/issues/3422

### DIFF
--- a/code/controllers/ContentController.php
+++ b/code/controllers/ContentController.php
@@ -216,9 +216,13 @@ class ContentController extends Controller {
 	public function httpError($code, $message = null) {
 		// Don't use the HTML response for media requests
 		$response = $this->request->isMedia() ? null : ErrorPage::response_for($code);
+		// if we have a broken error and response is null do not reveal controller info
+		if (!$response && ($code < 200 || $code > 302)) {
+			$response = ErrorPage::response_for($code);
+		}
 		// Failover to $message if the HTML response is unavailable / inappropriate
 		parent::httpError($code, $response ? $response : $message);
-		}
+	}
 
 	/**
 	 * Get the project name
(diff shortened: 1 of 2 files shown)


code/controllers/ContentController.php
```diff
@@ -216,9 +216,13 @@ public function handleRequest(SS_HTTPRequest $request, DataModel $model = null)
 	public function httpError($code, $message = null) {
 		// Don't use the HTML response for media requests
 		$response = $this->request->isMedia() ? null : ErrorPage::response_for($code);
+		// if we have a broken error and response is null do not reveal controller info
+		if (!$response && ($code < 200 || $code > 302)) {
+			$response = ErrorPage::response_for($code);
```
tractorcow: This doesn't prevent an error page being returned if the response is a media type. If you have an  it should not attempt to return an ErrorPage. That's what the above condition attempts to prevent.

sminnee:
This change breaks deliberately designed functionality. See my explanation in #3422.

tractorcow:
I suggest


```php
	/**
	 * @uses ErrorPage::response_for()
	 */
	public function httpError($code, $message = null) {
		// Don't use the HTML response for media requests
		if($this->request->isMedia()) $response = $message = null;
		else $response = ErrorPage::response_for($code);

		// Failover to $message if the HTML response is unavailable / inappropriate
		parent::httpError($code, $response ? $response : $message);
	}
````

dhensby:
I assume this is not going to get any loving any time soon - feel free to re-open if it will

PR #1095, opened 2014-08-26
Removing stage reading mode as it will be set in LeftAndMain

dhensby:
This depends on https://github.com/silverstripe/silverstripe-framework/pull/3437


dhensby:
see #1384

Issue #1099, opened 2014-09-11
VirtualPage controller doesn't call init() in linked controller


stnvh:
The ```__call()``` method which handles the method forwarding never handles ```init()``` because the method is recognised (exists on the current / parent controller).
A temporary workaround to this is creating a subclass of VirtualPage and using ```init()``` to manually call ```init()``` on the target controller and the VirtualPage controller:
```php
public function init() {
	parent::init();
	$this->__call('init', array());
}
```
[This](https://github.com/silverstripe/silverstripe-cms/pull/905) could be a potential fix as ```init``` could be added to the list of white-listed methods but it would mean the ```VirtualPage_Controller->init()``` method would never be called. 
A custom handler would need to be implemented which calls both ```VirtualPage_Controller->init()``` and the linked page controller's ```init()``` function to fix this bug.

EDIT:
Adding ```$this->__call('init', array());``` to ```VirtualPage_Controller->init()``` seems to fix this issue. I'll create a pull request when I can.

PR #1102, opened 2014-09-16
VirtualPage call init on target controller

stnvh:
Fixes [#1099](https://github.com/silverstripe/silverstripe-cms/issues/1099)

### DIFF
--- a/code/model/VirtualPage.php
+++ b/code/model/VirtualPage.php
@@ -493,6 +493,7 @@ class VirtualPage_Controller extends Page_Controller {
 			}
 		}
 		parent::init();
+		$this->__call('init', array());
 	}
 
 	public function loadcontentall() {


wilr:
e2d85b2f88193a356311eb7b433710d6342c9273

dhensby:
@wilr this is a bug fix, so should make it's way into 3.1? I suggest this gets merged

wilr:
I'm an api change too

dhensby:
Hmm? It's not an api change that I see? Definitely count it as a bug fix myself

chillu:
I agree its an API change - init() typically has side effects, e.g. `Requirements` inclusion. That's likely desired in most cases, but not all of them - so should go against 4.0

dhensby:
Ok

PR #1109, opened 2014-10-02
Update it.yml

g4b0:
Translation misprint


chillu:
Sorry, this has to be performed via Transifex: https://docs.silverstripe.org/en/3.1/contributing/translations/

Issue #1114, opened 2014-10-15
BUG SiteTree _resampled images URLs don't get rewritten when folders move.


micmania1:
To reproduce:
* Create a new Folder in the 'Files' section.
* Go to any Page and add an image to the Content.
* Resize the image.
* Save & Publish
* Go to the Files section and rename the folder.
* Go back to the page (note that the image path hasn't updated.).

sminnee:
I believe that this has been addressed with the SS4 filesystem rewrite. Can you confirm, @micmania1?

tractorcow:
I can confirm this is fixe in 4.x. Moving files moves all generated files, and there are no more static file paths in HTML areas anymore (since the move to shortcodes).

PR #1115, opened 2014-10-16
FIX Ensure resampled image links get updated when files are moved.

micmania1:
This is a fix for #1114. Still needs tests but would be good to get a second opinion on the solution first.


dhensby:
@micmania1 what happened with this in the end?

micmania1:
Its still on my todo list :)

PR #1137, opened 2014-11-26
ENH adding restore batch action

IgorNadj:
We have a batch action for Delete from Draft Site (CMSBatchAction_Delete) but no corresponding way to undo that deletion. This PR adds it.


IgorNadj:
Has been added to CMS by another change

tractorcow:
Lol, I remember writing this. :D

PR #1140, opened 2014-12-01
PHPDoc improvements for SiteTree and SiteTreeLinkTracking

nikrolls:
- Removed @param/@method references to fields from extensions
- Added @mixin references to extensions (supported by PHPStorm, should also be by Netbeans)
- Fixed incorrect PHPDoc @params and @returns
- Fixed wrapping width on many docblocks
- Added and/or improved method documentation when missing or confusing
- General tweaks and fixes to match latest standards and IDE capabilities
- Fixed typos as found

For the Framework counterpart see https://github.com/silverstripe/silverstripe-framework/pull/3698

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -1,17 +1,16 @@
 <?php
 /**
- * Basic data-object representing all pages within the site tree.
- * This data-object takes care of the heirachy. All page types that live within the hierarchy should inherit from this.
- * In addition, it contains a number of static methods for querying the site tree.
+ * Basic data-object representing all pages within the site tree. All page types that live within the hierarchy should
+ * inherit from this. In addition, it contains a number of static methods for querying the site tree and working with
+ * draft and published states.
  * 
  * <h2>URLs</h2>
- * A page is identified during request handling via its "URLSegment" database column. 
- * As pages can be nested, the full path of a URL might contain multiple segments.
- * Each segment is stored in its filtered representation (through {@link URLSegmentFilter}).
- * The full path is constructed via {@link Link()}, {@link RelativeLink()} and {@link AbsoluteLink()}.
- * You can allow these segments to contain multibyte characters through {@link URLSegmentFilter::$default_allow_multibyte}.
+ * A page is identified during request handling via its "URLSegment" database column. As pages can be nested, the full
+ * path of a URL might contain multiple segments. Each segment is stored in its filtered representation (through
+ * {@link URLSegmentFilter}). The full path is constructed via {@link Link()}, {@link RelativeLink()} and
+ * {@link AbsoluteLink()}. You can allow these segments to contain multibyte characters through
+ * {@link URLSegmentFilter::$default_allow_multibyte}.
  *
- * @property integer ID ID of the SiteTree object.
  * @property string URLSegment
  * @property string Title
  * @property string MenuTitle
@@ -21,17 +20,17 @@
  * @property string ShowInMenus
  * @property string ShowInSearch
  * @property string Sort Integer value denoting the sort order.
- * @property string HasBrokenFile
- * @property string HasBrokenLink
  * @property string ReportClass
  * @property string CanViewType Type of restriction for viewing this object.
  * @property string CanEditType Type of restriction for editing this object.
  *
- * @method ManyManyList LinkTracking() List of site pages linked on this page.
- * @method ManyManyList ImageTracking() List of Images linked on this page.
- * @method ManyManyList ViewerGroups() List of groups that can view this object.
- * @method ManyManyList EditorGroups() List of groups that can edit this object.
- * @method ManyManyList BackLinkTracking() List of site pages that link to this page.
+ * @method ManyManyList ViewerGroups List of groups that can view this object.
+ * @method ManyManyList EditorGroups List of groups that can edit this object.
+ * @method ManyManyList BackLinkTracking List of site pages that link to this page.
+ *
+ * @mixin Hierarchy
+ * @mixin Versioned
+ * @mixin SiteTreeLinkTracking
  *
  * @package cms
  */
@@ -162,8 +161,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	/**
 	 * If this is false, the class cannot be created in the CMS by regular content authors, only by ADMINs.
 	 * @var boolean
-	* @config
-	*/
+	 * @config
+	 */
 	private static $can_create = true;
 
 	/**
@@ -228,7 +227,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	private static $enforce_strict_hierarchy = true;
 
 	/**
-	 * The value used for the meta generator tag.  Leave blank to omit the tag.
+	 * The value used for the meta generator tag. Leave blank to omit the tag.
 	 *
 	 * @config
 	 * @var string
@@ -317,8 +316,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	 * Note that if no model can be found, this method will fall over to a extended alternateGetByLink method provided
 	 * by a extension attached to {@link SiteTree}
 	 *
-	 * @param string $link
-	 * @param bool $cache
+	 * @param string $link  The link of the page to search for
+	 * @param bool   $cache True (default) to use caching, false to force a fresh search from the database
 	 * @return SiteTree
 	 */
 	static public function get_by_link($link, $cache = true) {
@@ -388,9 +387,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	}
 	
 	/**
-	 * Return a subclass map of SiteTree
-	 * that shouldn't be hidden through
-	 * {@link SiteTree::$hide_ancestor}
+	 * Return a subclass map of SiteTree that shouldn't be hidden through {@link SiteTree::$hide_ancestor}
 	 *
 	 * @return array
 	 */
@@ -413,7 +410,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 			}
 		}
 
-		// If any of the descendents don't want any of the elders to show up, cruelly render the elders surplus to requirements.
+		// If any of the descendents don't want any of the elders to show up, cruelly render the elders surplus to
+		// requirements
 		if($kill_ancestors) {
 			$kill_ancestors = array_unique($kill_ancestors);
 			foreach($kill_ancestors as $mark) {
@@ -429,10 +427,10 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	/**
 	 * Replace a "[sitetree_link id=n]" shortcode with a link to the page with the corresponding ID.
 	 *
-	 * @param array $arguments
-	 * @param mixed $content
-	 * @param object|null $parser
-	 * @return string|void
+	 * @param array      $arguments
+	 * @param string     $content
+	 * @param TextParser $parser
+	 * @return string
 	 */
 	static public function link_shortcode_handler($arguments, $content = null, $parser = null) {
 		if(!isset($arguments['id']) || !is_numeric($arguments['id'])) return;
@@ -453,14 +451,14 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 			return $link;
 		}
 	}
-	
+
 	/**
 	 * Return the link for this {@link SiteTree} object, with the {@link Director::baseURL()} included.
 	 *
-	 * @param string $action Optional controller action (method). 
-	 *  Note: URI encoding of this parameter is applied automatically through template casting,
-	 *  don't encode the passed parameter.
-	 *  Please use {@link Controller::join_links()} instead to append GET parameters.
+	 * @param string $action Optional controller action (method).
+	 *                       Note: URI encoding of this parameter is applied automatically through template casting,
+	 *                       don't encode the passed parameter. Please use {@link Controller::join_links()} instead to
+	 *                       append GET parameters.
 	 * @return string
 	 */
 	public function Link($action = null) {
@@ -482,9 +480,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	}
 	
 	/**
-	 * Base link used for previewing. Defaults to absolute URL,
-	 * in order to account for domain changes, e.g. on multi site setups.
-	 * Does not contain hints about the stage, see {@link SilverStripeNavigator} for details.
+	 * Base link used for previewing. Defaults to absolute URL, in order to account for domain changes, e.g. on multi
+	 * site setups. Does not contain hints about the stage, see {@link SilverStripeNavigator} for details.
 	 * 
 	 * @param string $action See {@link Link()}
 	 * @return string
@@ -500,7 +497,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	/**
 	 * Return the link for this {@link SiteTree} object relative to the SilverStripe root.
 	 *
-	 * By default, it this page is the current home page, and there is no action specified then this will return a link
+	 * By default, if this page is the current home page, and there is no action specified then this will return a link
 	 * to the root of the site. However, if you set the $action parameter to TRUE then the link will not be rewritten
 	 * and returned in its full form.
 	 *
@@ -529,9 +526,12 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 
 		return Controller::join_links($base, '/', $action);
 	}
-	
+
 	/**
 	 * Get the absolute URL for this page on the Live site.
+	 *
+	 * @param bool $includeStageEqualsLive Whether to append the URL with ?stage=Live to force Live mode
+	 * @return string
 	 */
 	public function getAbsoluteLiveLink($includeStageEqualsLive = true) {
 		$oldStage = Versioned::current_stage();
@@ -549,7 +549,9 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	}
 	
 	/**
-	 * @return String
+	 * Generates a link to edit this page in the CMS.
+	 *
+	 * @return string
 	 */
 	public function CMSEditLink() {
 		return Controller::join_links(singleton('CMSPageEditController')->Link('show'), $this->ID);
@@ -566,7 +568,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	}
 	
 	/**
-	 * Returns TRUE if this is the currently active page that is being used to handle a request.
+	 * Returns true if this is the currently active page being used to handle this request.
 	 *
 	 * @return bool
 	 */
@@ -575,8 +577,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	}
 	
 	/**
-	 * Check if this page is in the currently active section (e.g. it is either current or one of it's children is
-	 * currently being viewed.
+	 * Check if this page is in the currently active section (e.g. it is either current or one of its children is
+	 * currently being viewed).
 	 *
 	 * @return bool
 	 */
@@ -587,9 +589,9 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	}
 	
 	/**
-	 * Check if the parent of this page has been removed (or made otherwise unavailable), and 
-	 * is still referenced by this child. Any such orphaned page may still require access via
-	 * the cms, but should not be shown as accessible to external users.
+	 * Check if the parent of this page has been removed (or made otherwise unavailable), and is still referenced by
+	 * this child. Any such orphaned page may still require access via the CMS, but should not be shown as accessible
+	 * to external users.
 	 * 
 	 * @return bool
 	 */
@@ -621,8 +623,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	}
 	
 	/**
-	 * Return "link", "current" or section depending on if this page is the current page, or not on the current page but
-	 * in the current section.
+	 * Return "link", "current" or "section" depending on if this page is the current page, or not on the current page
+	 * but in the current section.
 	 *
 	 * @return string
 	 */
@@ -639,8 +641,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	/**
 	 * Check if this page is in the given current section.
 	 *
-	 * @param string $sectionName Name of the section to check.
-	 * @return boolean True if we are in the given section.
+	 * @param string $sectionName Name of the section to check
+	 * @return bool True if we are in the given section
 	 */
 	public function InSection($sectionName) {
 		$page = Director::get_current_page();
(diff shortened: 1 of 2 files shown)


code/model/SiteTree.php
```diff
@@ -191,7 +190,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	
 	private static $searchable_fields = array(
 		'Title',
-		'Content',
+		'Content'
```
dhensby: Please keep the trailing comma

nikrolls: Do you want that fixed to approve the PR, or just a general comment for whoever approves it?

tractorcow:
`@mixin` isn't a standard PHPDoc syntax, although I like the sentiment. Can you show me any documentation on how this feature is used?

You should keep the round braces `()` when using the `@method` directive. http://phpdoc.org/docs/latest/references/phpdoc/tags/method.html

You should also use the `$` symbol when defining `@property` http://phpdoc.org/docs/latest/references/phpdoc/tags/property.html

tractorcow:
Ah I found http://code4x.com/php/decorator-pattern-in-php-with-autocomplete-in-phpstorm/ :)

nikrolls:
Yep, that's the one. Also the announcement here: http://blog.jetbrains.com/phpstorm/2013/11/phpstorm-7-1-eap-133-51/
It effectively tells the IDE to autocomplete the target class as if it were a `trait`. This means we can keep the `@property` and `@method` definitions on the class they belong to, and then use `@mixin` where there are out-of-the-box extensions. No need to update `SiteTree`'s documentation when `Versioned` changes, for example. `@mixin` is currently supported by PhpStorm and in-progress (if not already implemented) for NetBeans.
I'll fix the rest of the things you mentioned and restore the comma @dhensby asked for tonight, and push it to this branch. Are you ok with two commits in this PR, considering you usually ask for them to be squashed?

tractorcow:
I'll test it out on my netbeans later. Will let you know.

tractorcow:
Nope it doesn't work in netbeans. =(

nikrolls:
Hmm, not as far along as I thought: https://netbeans.org/bugzilla/show_bug.cgi?id=241740

dhensby:
@assertchris - any thoughts on this one too :)

code/model/SiteTree.php
```diff
@@ -317,8 +316,8 @@ static public function get_create_default_pages() {
 	 * Note that if no model can be found, this method will fall over to a extended alternateGetByLink method provided
 	 * by a extension attached to {@link SiteTree}
 	 *
-	 * @param string $link
-	 * @param bool $cache
+	 * @param string $link The link of the page to search for
+	 * @param bool $cache True (default) to use caching, false to force a fresh search from the database
```
assertchris: Should allow `null`.

code/model/SiteTree.php
assertchris: Good idea, in general!

code/model/SiteTree.php
```diff
@@ -430,9 +428,9 @@ static public function page_type_classes() {
 	 * Replace a "[sitetree_link id=n]" shortcode with a link to the page with the corresponding ID.
 	 *
 	 * @param array $arguments
-	 * @param mixed $content
-	 * @param object|null $parser
-	 * @return string|void
+	 * @param string $content
```
assertchris: Should allow `null`. As should the rest...

code/model/SiteTree.php
```diff
@@ -430,9 +428,9 @@ static public function page_type_classes() {
...
+	 * @param TextParser $parser
```
assertchris: Documented `TextParser`, hinted `mixed`.

code/model/SiteTree.php
```diff
@@ -453,14 +451,14 @@ static public function link_shortcode_handler($arguments, $content = null, $pars
...
-	
+
 	/**
 	 * Return the link for this {@link SiteTree} object, with the {@link Director::baseURL()} included.
 	 *
-	 * @param string $action Optional controller action (method). 
-	 *  Note: URI encoding of this parameter is applied automatically through template casting,
-	 *  don't encode the passed parameter.
-	 *  Please use {@link Controller::join_links()} instead to append GET parameters.
+	 * @param string $action Optional controller action (method).
```
assertchris: Should allow `null`.

code/model/SiteTree.php
```diff
@@ -529,9 +526,12 @@ public function RelativeLink($action = null) {
 
 		return Controller::join_links($base, '/', $action);
 	}
-	
+
 	/**
 	 * Get the absolute URL for this page on the Live site.
+	 *
+	 * @param bool $includeStageEqualsLive Whether to append the URL with ?stage=Live to force Live mode
```
assertchris: Should allow `null`.

code/model/SiteTree.php
```diff
@@ -714,13 +712,12 @@ public function duplicateAsChild($id) {
 	}
 	
 	/**
-	 * Return a breadcrumb trail to this page. Excludes "hidden" pages
-	 * (with ShowInMenus=0).
+	 * Return a breadcrumb trail to this page. Excludes "hidden" pages (with ShowInMenus=0) by default.
 	 *
 	 * @param int $maxDepth The maximum depth to traverse.
```
assertchris: Should allow `null`. As should the rest...

code/model/SiteTree.php
```diff
@@ -804,10 +799,9 @@ public function NestedTitle($level = 2, $separator = " - ") {
 	 *
 	 * @uses DataObjectDecorator->can()
```
assertchris: Should we be using these? Can be inferred, leads to lots of work...

code/model/SiteTree.php
```diff
@@ -804,10 +799,9 @@ public function NestedTitle($level = 2, $separator = " - ") {
...
 	 *
-	 * @param string $perm The permission to be checked, such as 'View'.
-	 * @param Member $member The member whose permissions need checking.
-	 *        Defaults to the currently logged in user.
-	 * @return boolean True if the the member is allowed to do the given action.
+	 * @param string $perm The permission to be checked, such as 'View'
+	 * @param Member $member The member whose permissions need checking. Defaults to the currently logged in user.
```
assertchris: Should allow `null`.  Also: documented `Member`, hinted `mixed`.

code/model/SiteTree.php
```diff
@@ -874,8 +864,8 @@ public function canAddChildren($member = null) {
 	 * @uses DataExtension->canView()
 	 * @uses ViewerGroups()
 	 *
-	 * @param Member|int|null $member
-	 * @return boolean True if the current user can view this page.
+	 * @param Member|int $member
```
assertchris: Should allow `null`.

code/model/SiteTree.php
```diff
@@ -940,9 +930,9 @@ public function canView($member = null) {
 	 * 
 	 * @todo Implement in CMS UI.
 	 * 
-	 * @param String $stage
+	 * @param string $stage
```
assertchris: Should allow `null`. As should the rest...

code/model/SiteTree.php
```diff
@@ -940,9 +930,9 @@ public function canView($member = null) {
...
 	 * @param Member $member
```
assertchris: Documented `Member`, hinted `mixed`.

code/model/SiteTree.php
```diff
@@ -955,21 +945,20 @@ public function canViewStage($stage = 'Live', $member = null) {
...
+	 * Denies permission if any of the following conditions is true:
+	 * - canDelete() returns false on any extension
+	 * - canEdit() returns false
+	 * - any descendant page returns false for canDelete()
 	 * 
 	 * @uses canDelete()
 	 * @uses SiteTreeExtension->canDelete()
 	 * @uses canEdit()
 	 *
 	 * @param Member $member
```
assertchris: Documented `Member`, hinted `mixed`. Should allow `null`.

code/model/SiteTree.php
```diff
@@ -1007,7 +995,7 @@ public function canDelete($member = null) {
 	 * @uses DataExtension->canCreate()
 	 *
 	 * @param Member $member
```
assertchris: Should allow `null`. Also: documented `Member`, hinted `mixed`.

code/model/SiteTree.php
```diff
@@ -1025,23 +1013,24 @@ public function canCreate($member = null) {
...
+	 *   CMS_Access_CMSMAIN permission code
 	 * - "CanEditType" directive is set to "OnlyTheseUsers" and user is not in the given groups
 	 * 
 	 * @uses canView()
 	 * @uses EditorGroups()
 	 * @uses DataExtension->canEdit()
 	 *
-	 * @param Member $member Set to FALSE if you want to explicitly test permissions without a valid user (useful for unit tests)
-	 * @return boolean True if the current user can edit this page.
+	 * @param Member $member Set to false if you want to explicitly test permissions without a valid user (useful for
```
assertchris: Should allow `null`. Also: documented `Member`, hinted `mixed`.

code/model/SiteTree.php
```diff
@@ -1069,18 +1058,17 @@ public function canEdit($member = null) {
...
-	 * Denies permission if any of the following conditions is TRUE:
-	 * - canPublish() on any extension returns FALSE
-	 * - canEdit() returns FALSE
+	 * Denies permission if any of the following conditions is true:
+	 * - canPublish() on any extension returns false
+	 * - canEdit() returns false
 	 * 
 	 * @uses SiteTreeExtension->canPublish()
 	 *
 	 * @param Member $member
```
assertchris: Should allow `null`. Also: documented `Member`, hinted `mixed`.

code/model/SiteTree.php
```diff
@@ -1117,13 +1107,13 @@ public function getSiteConfig() {
...
-	 * This method will use the static can_(perm)_multiple method for efficiency.
-	 * 
-	 * @param string $permission The permission: edit, view, publish, approve, etc.
-	 * @param array $ids An array of page IDs
-	 * @param callback|null $batchCallback The function/static method to call to calculate permissions.  Defaults
-	 * to 'SiteTree::can_(permission)_multiple'
+	 * Pre-populate the cache of canEdit, canView, canDelete, canPublish permissions. This method will use the static
+	 * can_(perm)_multiple method for efficiency.
+	 *
+	 * @param string          $permission    The permission: edit, view, publish, approve, etc.
```
assertchris: Should allow `null`. As should the rest...

code/model/SiteTree.php
```diff
@@ -1135,26 +1125,27 @@ static public function prepopulate_permission_cache($permission = 'CanEditType',
...
+	 *
+	 * Caches based on $typeField data. To invalidate the cache, use {@link SiteTree::reset()} or set the $useCached
+	 * property to FALSE.
+	 *
+	 * @param array  $ids              Of {@link SiteTree} IDs
+	 * @param int    $memberID         Member ID
+	 * @param string $typeField        A property on the data record, e.g. "CanEditType".
+	 * @param string $groupJoinTable   A many-many table name on this record, e.g. "SiteTree_EditorGroups"
+	 * @param string $siteConfigMethod Method to call on {@link SiteConfig} for toplevel items, e.g. "canEdit"
+	 * @param string $globalPermission If the member doesn't have this permission code, don't bother iterating deeper
```
assertchris: Should allow `null`. As should the rest...

code/model/SiteTree.php
```diff
@@ -1265,15 +1257,15 @@ static public function batch_permission_check($ids, $memberID, $typeField, $grou
...
-	 * 
-	 * @param array $ids An array of IDs of the SiteTree pages to look up.
-	 * @param int $memberID ID of member.
-	 * @param bool $useCached Return values from the permission cache if they exist.
-	 * @return array A map where the IDs are keys and the values are booleans stating whether the given
-	 * page can be edited.
+	 *
+	 * @param array $ids       An array of IDs of the SiteTree pages to look up
+	 * @param int   $memberID  ID of member
+	 * @param bool  $useCached Return values from the permission cache if they exist
```
assertchris: Should allow `null`.

code/model/SiteTree.php
```diff
@@ -1281,9 +1273,10 @@ static public function can_edit_multiple($ids, $memberID, $useCached = true) {
 
 	/**
 	 * Get the 'can edit' information for a number of SiteTree pages.
-	 * @param array $ids An array of IDs of the SiteTree pages to look up.
-	 * @param int $memberID ID of member.
-	 * @param bool $useCached Return values from the permission cache if they exist.
+	 *
+	 * @param array $ids       An array of IDs of the SiteTree pages to look up
+	 * @param int   $memberID  ID of member
+	 * @param bool  $useCached Return values from the permission cache if they exist
```
assertchris: Should allow `null`.

code/model/SiteTree.php
```diff
@@ -1368,13 +1357,12 @@ public function collateDescendants($condition, &$collator) {
...
-
 	/**
 	 * Return the title, description, keywords and language metatags.
-	 * 
+	 *
 	 * @todo Move <title> tag in separate getter for easier customization and more obvious usage
-	 * 
-	 * @param boolean|string $includeTitle Show default <title>-tag, set to false for custom templating
+	 *
+	 * @param boolean $includeTitle Show default <title>-tag, set to false for custom templating
```
assertchris: Should allow `null`.

code/model/SiteTree.php
```diff
@@ -1411,29 +1399,23 @@ public function MetaTags($includeTitle = true) {
...
+	 * Returns the object that contains the content that a user would associate with this page.
 	 *
-	 * Ordinarily, this is just the page itself, but for example on
-	 * RedirectorPages or VirtualPages ContentSource() will return the page
-	 * that is linked to.
+	 * Ordinarily, this is just the page itself, but for example on RedirectorPages or VirtualPages ContentSource() will
+	 * return the page that is linked to.
 	 *
-	 * @return SiteTree The content source.
+	 * @return SiteTree The content source
```
assertchris: `@return self` is more robust.

code/model/SiteTree.php
```diff
@@ -2146,9 +2124,8 @@ public function getSettingsFields() {
 	}
 	
 	/**
-	 *
-	 * @param boolean $includerelations a boolean value to indicate if the labels returned include relation fields
-	 * @return array|string
+	 * @param bool $includerelations A boolean value to indicate if the labels returned should include relation fields
```
assertchris: Should allow `null`.

code/model/SiteTree.php
```diff
@@ -2434,7 +2409,8 @@ public function doRevertToLive() {
 	
 	/**
 	 * Restore the content in the active copy of this SiteTree page to the stage site.
-	 * @return The SiteTree object.
+	 *
+	 * @return self The SiteTree object
```
assertchris: `@return self` is enough. IT IS ENOUGH! :)

code/model/SiteTree.php
```diff
@@ -2559,20 +2530,18 @@ protected function getClassDropdown() {
...
 	}
 
-
 	/**
-	 * Returns an array of the class names of classes that are allowed
-	 * to be children of this class.
+	 * Returns an array of the class names of classes that are allowed to be children of this class.
 	 *
-	 * @return array
+	 * @return string[]
```
assertchris: Standard pending.

code/model/SiteTree.php
```diff
@@ -2643,19 +2609,18 @@ public function setMenuTitle($value) {
...
 	 * Example (simple): 
-	 * "deletedonlive" => "Deleted"
+	 *   "deletedonlive" => "Deleted"
 	 * 
 	 * Example (with optional title attribute): 
-	 * "deletedonlive" => array('text' => "Deleted", 'title' => 'This page has been deleted')
+	 *   "deletedonlive" => array('text' => "Deleted", 'title' => 'This page has been deleted')
 	 *
-	 * @param Boolean $cached
+	 * @param bool $cached Whether to serve the fields from cache; false regenerate them
```
assertchris: Should allow `null`.

tractorcow: Why? The method doesn't seem to have any special consideration for null values of $cached. (same question on some of the above). It feel unnecessarily complicating, although maybe you know a reason I don't. :)

assertchris: The parameter has a default value which means null is a valid call value for that parameter:

```php
getStatusFlags() is the same as getStatusFlags(null)
```

"Why bother?" I hear you thinking. Because I can hear your thoughts. PHP supports parameters [with default values] between others that don't have default values:

```php
function getSomeValue($type, $cached = true, $shouldReturn)...
```

The `null` type hint tells the type checker that `getSomeValue("field", null, true)` is a valid call.

tractorcow: > The parameter has a default value which means null is a valid call value for that parameter:

The default value is true not null. getStatusFlags is getStatusFlags(true) not (null)

assertchris: > getStatusFlags is getStatusFlags(true) not getStatusFlags(null)

That would only be true if `true == null` which is not the case. Where it is called; `getStatusFlags()` is `getStatusFlags(null)`. The engine wires the default values up inside the function. I guess that's semantics though. 

The important thing is that [static analysis] engines view nullable type hints differently to nullable type hints. Could you give `null` as a value to that parameter? Then `null` is an allowed type. Whether or not it will be given a default value (inside the function) or will be coerced or whatever. The type hints (in this case the type annotations) tell a static analyser what types are allowed, not what they will become. That is why `null` should be hinted/annotated for all PHP parameters with default values.

dhensby: @assertchris are you saying that passing `null` as a parameter to a function means it uses the default value? Because I've had that discussion before with another dev and tested the code and it's not the case.

Therefore, `getStatusFlags(null)` is not the same as `getStatusFlags()`

Here's a pretty rudementary proof .

PS: sorry if I have the wrong end of the stick

assertchris: You're right. I got a bit confused there. I've adjusted the comment, and fortunately it still makes sense. :)

dhensby: Ok, so does it come down to:

1. Whether strict type checking is used in the function (null will be falsey and so is technically accepted param type)
2. ~~Whether the type is "hinted" in the function declaration eg: `function getStatusFlags(bool $cached = true)`~~ in fact, it seems you can't hint at booleans at all.

Isn't `null` (and any other type) technically valid even if you comply with 1 (it'll just always fail a type comparison) but will pass a loose comparison?

(I'm asking to seek education that to be provocative)

code/model/SiteTree.php
nikrolls: How about `@return static`? That would ensure the correct type hinting for subclasses (considering the function returns `$this`). `@return self` implies `SiteTree`. In fact, I usually document it as `@return $this` which works well in PhpStorm but I'm not sure if it's standard.

code/model/SiteTree.php
assertchris: There are no scalar type hints (yet), so all scalar types are loose comparisons. The annotations are our way of hinting at strict, scalar types. We could leave the types or hint mixed. If we're being specific (this is `int`, this is `string etc.) then we should be as specific as possible.

The [PSR recommends](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#details) being as detailed as possible.

The reason for all the detail is so that a static analyser can predict errors before runtime. If you don't annotate `null|bool` on an optional parameter then PHPStorm (or whatever static analyser you use) will warn you that you could be passing null to function that doesn't expect it.

assertchris: The more we discuss it, the less convicted I feel about adding null to these types. *sigh*

nikrolls: I'm of the opinion that documentation in loosely typed languages should represent semantics and not how the function itself actually works.

Can you pass `null` as a parameter? Usually always. Should you document `null` as a parameter type? Only if `null` is one of the intended usages of that method.

For instance:

```
public function doSomething($input = null) {
    if ($input = null) {
        // Set $input to something calculated that can't be represented as a default value
```

In this case I would not document `null` as being an option because you would never need to pass `null` to the function. Instead I would document `$input` as an optional override and state the expected type.

However, if there was a second parameter I *would* document `null` because you should use that to access the second parameter while keeping the default value for the first.

However if the default value was, for instance, `false`, and there was a second parameter, I would not document `null` as an option even if `null` would be coerced to `false` within the function.

Basically, my rule of thumb is that documentation should be for the user, and not the documentation library or the IDE. The latter can both read the method signature and identify the technically legal types. The documentation should always reflect the *intended* uses rather than the *possible* uses.

assertchris: So basically ignore all my comments to allow `null`. :)

Good talk!

dhensby: > Can you pass `null` as a parameter? Usually always. Should you document `null` as a parameter type? Only if `null` is one of the intended usages of that method. / The documentation should always reflect the *intended* uses rather than the *possible* uses.

Yes, this is what I'm thinking at the moment

code/model/SiteTree.php
assertchris: The spec allows all three, but I prefer `self` for `return $this` and `static` for instances of the same type/subtype that are not `$this`.

nikrolls: However `$this` will always be the subtype, just like `static`. `self` is always the parent type.

Eg, if you have an instance of `Page` this function will actually return a `Page` object, but with `@return self` the IDE will only autocomplete for `SiteTree`. If you use `@return static` then the IDE will autocomplete for `Page`, and if you use `@return $this` then the IDE can actually infer the current instance state for the object returned.

chillu:
@nikrolls could you please rebase, and determine what's blocking this PR? Discussion got a bit long :)

nikrolls:
@chillu Sorry, my fault -- I changed jobs and it got a bit hectic. More time to get back to this now it's settled down.
In summary, we came to the conclusion that we should document according to the methods' intended uses rather than their possible uses, and I've actioned the agreed upon feedback. 

There are a bunch of items 'Documented N, hinted mixed' which are there because I intended this to be a documentation-only PR so it could be merged as quickly as possible without having to worry about a minor or major version increment. I'd love to go through and add type-hinting everywhere but that would require a lot of testing and possibly even a major version increase to safely merge them.

I think apart from that it's good to go. I'm not sure what the Scrutinizer error is since there are no code changes.

code/model/SiteTree.php
chillu: Yeah agreed, `@uses` seems like unnecessary busywork. Nick?

chillu:
There's three outstanding comments from Chris, but otherwise its good to go (at least from cursory review). Thanks for seeing this through Nik! Any chance you could have a look over other core classes as well? :D

code/model/SiteTree.php
harmoney-nikr: Agreed. I would only use it where there are less-than-obvious side-effects, for example `@uses DataObject->delete()` on `DataList#remove`. But not in this instance.

nikrolls: I would only use it where there are less-than-obvious side-effects, for example `@uses DataObject->delete()` on `DataList#remove`. But not in this instance.

nikrolls:
@chillu Absolutely! I'm thinking of Requirements next because it has quite a few broken doc blocks. And then I'll see where to from there, looking at most frequently used classes first.

PR #1149, opened 2015-01-08
FIX: Fixes #1054 By preventing errors in the CMS _only_.

### LINKED ISSUE #1054: Compare mode on History tab on draft-deleted object produces Internal Server Error
Hi,

I've deleted an item from the draft site. 
image

Whenever I go to the history tab, select compare mode and select two version that are there in the system, I get Internal Server Error. 

Corresponding log entry is following:

    [Thu Jul 17 13:05:17 2014] [error] [client 80.31.52.242] PHP Fatal error:  Call to a member function compareVersions() on a non-object in /home/sites/cnk1/mpp1cms/www2/cms/code/controllers/CMSPageHistoryController.php on line 381

It's not a big issue at this moment, but definitely does not make the site look very professional for our users.

The page I am trying to see is a subclass of Page, but nothing special really apart from a few extra fields.

phptek:
Because of the way we're fetching pages via the ORM, "deleted" pages are
never fetched, so the comparison/diff logic cannot therefore run either.

This patch therefore simply addresses errors being thrown in the CMS
when attempting to view the history of a "deleted" page.


tractorcow:
You'll probably need to rebase this to get the tests passing.

Can you please fix this is 3.3 branch, since you'll have to re-base it anyway?

dhensby:
Bump

tractorcow:
Bamp.

dhensby:
@phptek are you going to be able to close and re-open against 3.2/3.3?

phptek:
@dhensby sure..give a day or so.

dhensby:
Thanks, @phptek please do :)

phptek:
@dhensby issued against 3.3.0 with some minor mods, hopefully they don't confuse the original work (Added missing `@return` types, missing braces and used `exists()` and `isset()` as necessary. Feedback welcome :-)

tractorcow:
@phptek if you rebase on 3.3 you'll need to create a new pull request; You can't change the target branch of open pull requests.

Issue #1128, opened 2014-11-06
$description not working in page.php


Friksel:
When changing the $description in page.php (like so: private static $description = 'Base Page';), the description of sitetree.php remains ("Generic content Page") and so doesn't get overwritten by page.php. Neither after dev/build. For all other pages setting the description works like charm, but not for page.php itself. Looks like I'm not the only one having this problem, so it seems like it's a bug.

Using Silverstripe 3.1.6

Friksel:
Sorry for my late reply
 @cam-findlay because I'm not that into the framework and there's probably a good reason why it's not build this way in the first place. I don't have an overview on dependencies and what it can influence for example. Seems like a small change for people that work on the framework itself on a 'daily' basis

tractorcow:
Fixed with the SiteTree::classDescription() method

https://github.com/silverstripe/silverstripe-cms/blob/63f930c4e990ae2502c6b7f0e141dec59982e1b4/code/Model/SiteTree.php#L2804

PR #1152, opened 2015-01-13
FIX. Convert Page to SiteTree for correct localization lookups

uniun:
Convert 'Page' to 'SiteTree' for correct localization lookups.
Similar to line 493.

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -350,7 +350,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 	protected function getPageTypes() {
 		$pageTypes = array();
 		foreach(SiteTree::page_type_classes() as $pageTypeClass) {
-			$pageTypes[$pageTypeClass] = _t($pageTypeClass.'.SINGULARNAME', $pageTypeClass);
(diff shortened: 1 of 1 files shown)


dhensby:
There is, in fact, a bug report open complaining about line 493.

I don't think this is behaviour we want to have as it removes control from the developer.

Perhaps for the Page class we can use SiteTree if the translation for Page doesn't exist, but I don't think it's right to assume that someone who has created a `Page` class wants to have the descriptions or names of `SiteTree`

dhensby:
see #1128 - consensus seems to be that the current behaviour is wrong

uniun:
Ok, I agree that current behavior is wrong and should be fixed.

Page class isn't created by someone, it's created by SS installer and it should be translated :) Usually, there is no need to have separate translation for Page and SiteTree. That's why this lookup is useful.
Also, it's impossible to translate Page object in Transifex. It doesn't have necessary entities there:

https://www.transifex.com/projects/p/silverstripe-cms/translate/#lt/$/14803080?key=Page.SINGULARNAME
https://www.transifex.com/projects/p/silverstripe-cms/translate/#lt/$/14803077?key=Page.DESCRIPTION

It can be fixed by adding these entities to the base language file. And remove that lookup then :)

PR #1153, opened 2015-01-13
BUG. Missing translation entity.

uniun:
For some reason this entity is absent on Transifex.


chillu:
It's present now. Sorry for the late response :(

uniun:
Great! Thanks.

PR #1168, opened 2015-02-20
Fix auto-setting of URL from localized page titles

chillu:
Replace the hardcoded check for english locale ("new-page") with a localized version. Tested in IE8 and Chrome. See https://github.com/silverstripe/silverstripe-translatable/issues/180

### DIFF
--- a/tests/model/SiteTreeTest.php
+++ b/tests/model/SiteTreeTest.php
@@ -20,7 +20,7 @@ class SiteTreeTest extends SapphireTest {
 		'SiteTreeTest_NotRoot',
 		'SiteTreeTest_StageStatusInherit',
 	);
-
+	
 	/**
 	 * Ensure any current member is logged out
 	 */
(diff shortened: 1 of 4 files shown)


tests/model/SiteTreeTest.php
```diff
@@ -653,7 +653,62 @@ public function testIsSection() {
 		$this->assertTrue($staff->isSection());
 		$this->assertTrue($ceo->isSection());
 	}
-	
+
+	public function testURLSegmentAutoUpdate() {
+		Config::nest();
```
dhensby: Doesn't this need to be unnested at the end of the test?

chillu: Good spotting - was a remnant of the test code split out to `testURLSegmentAutoUpdateLocalized()`. Turns out you can't control `i18n::$current_locale` via the Config API...

chillu:
Rebased and incorporated feedback, @dhensby can you please review again?

bueckl:
Hey guys, any chance this change will make it into the master branch anytime soon?

... would come in handy.  Thanks

dhensby:
Ok, looks good - I'll try to merge it by hand and get rid of all that white spacing you added ;)

bueckl:
Yayyyy, great :)

Issue #1167, opened 2015-02-20
Not able to specify ::$allowed_parents, only ::$allowed_children on pages.


patricknelson:
Not sure if this has affected anyone else, but I've noticed that there's no way to intuitively define which parents a child can have, if you need to. I've had a situation where I need all the flexibility/functionality of a page, but due to it's own functionality, should only be able to be added to a certain kind of parent. You can define the inverse relationship (that a certain parent can only have certain children) but you cannot restrict those same children from being added elsewhere without manually defining the relationship on ever single type of page in the CMS, which can be prohibitive and error prone from a development standpoint (e.g. if you forget to update one of the pages).

I found a pretty simple solution to this on my own by overriding the `->allowedChilren()` method and then using the following code:

```php
	public function allowedChildren() {
		// Default.
		$children = parent::allowedChildren();

		// Define allowed parents on a per page basis.
		$allowedParents = [];
		$pageTypes = SiteTree::page_type_classes();
		foreach($pageTypes as $pageType) {
			// Only worry about it if this page type has the private static ::$allowed_parents configured.
			$curParents = singleton($pageType)->stat("allowed_parents");
			if (!$curParents) continue;
			$allowedParents[$pageType] = $curParents;
		}

		// See if any one of the current allowed children exist in the array of pages above that restrict their parents.
		$restrictedChildren = array_intersect($children, array_keys($allowedParents));
		if (empty($restrictedChildren)) return $children; // ... nothing more to do.

		// Go through each of the children with restricted parents and remove it from the list of children if this class
		// DOES NOT appear in its list of allowed parents.
		$thisPage = get_class($this);
		foreach($restrictedChildren as $childPage) {
			if (!in_array($thisPage, $allowedParents[$childPage])) {
				$childIndex = array_search($childPage, $children);
				unset($children[$childIndex]);
			}
		}

		// Return processed array.
		return $children;
	}
```

Is this something that you guys would be interested in rolling out into the CMS somehow? I'm thinking that other people might find this useful as well, so maybe it could be implemented on `SiteTree` or as "yet another module".

dhensby:
I do believe this is something that the CMS is missing and needs to be addressed.

If you could open a PR of this against 3 branch, it'd be a way to get good feedback on it.

neilcreagh:
Any progress on Allowed Parents in Silverstripe? 

Using allowed_children is there a way to allow Page but not some of the pages that extend Page?

dhensby:
[...]

PR #1171, opened 2015-02-25
NEW: Adding feature #1167 for ::$allowed_parents to function like ::$allowed_children

patricknelson:
An implementation for #1167 on `3`, per @dhensby request.

### DIFF
--- a/tests/model/SiteTreeTest.php
+++ b/tests/model/SiteTreeTest.php
@@ -917,6 +917,40 @@ class SiteTreeTest extends SapphireTest {
 		);
 	}
 
+	public function testAllowedParents() {
+		$page = new Page();
+		$parent1 = new SiteTreeTest_SpecificParent1();
+		$parent2 = new SiteTreeTest_SpecificParent2();
+
+		// Ensure that none of the "Specific" parents and children show up as allowed children on a generic Page object.
+		$pageChildren = $page->allowedChildren();
+		$this->assertNotContains("SiteTreeTest_SpecificParent1", $pageChildren);
+		$this->assertNotContains("SiteTreeTest_SpecificParent2", $pageChildren);
+		$this->assertNotContains("SiteTreeTest_SpecificChild1", $pageChildren);
+		$this->assertNotContains("SiteTreeTest_SpecificChild2", $pageChildren);
+
+		// Ensure that only SpecificChild1 and SpecificChild2 show up as children for SpecificParent1 (SiteTreeTest_SpecificChild2 inherits from SiteTreeTest_SpecificChild1).
+		$this->assertEquals(array("SiteTreeTest_SpecificChild1", "SiteTreeTest_SpecificChild2"), $parent1->allowedChildren());
+
+		// Ensure that only SpecificChild2 shows up for SpecificParent2 (via inheritance in the child's allowed parents).
+		$this->assertEquals(array("SiteTreeTest_SpecificChild2"), $parent2->allowedChildren());
+
+		// Ensure that parent 1 can still be in the root, while parent 2 cannot.
+		$parent1->ParentID = 0;
+		$parent1->write();
+		$parent2->ParentID = 0;
+		$isDetected = false;
+		try {
+			$parent2->write();
+		} catch(ValidationException $e) {
+			$this->assertContains('is not allowed on the root level', $e->getMessage());
+			$isDetected = true;
+		}
+
+		if(!$isDetected) $this->fail('Fails validation with $can_be_root=false');
+
+	}
+
 	public function testAllowedChildrenValidation() {
 		$page = new SiteTree();
 		$page->write();
@@ -1225,4 +1259,29 @@ class SiteTreeTest_Extension extends DataExtension implements TestOnly {
 		return false;
 	}
 
+}
+
+class SiteTreeTest_SpecificParent1 extends Page implements TestOnly {
+	// Cannot have any parents at all.
+	private static $allowed_parents = array();
+	private static $allowed_children = array('SiteTreeTest_SpecificChild1');
+}
+
+class SiteTreeTest_SpecificParent2 extends SiteTreeTest_SpecificParent1 implements TestOnly {
+	// Defines no specific allowed_parents/allowed_children, only inherits that functionality. However, it cannot be in
+	// the root either. Essentially, this cannot exist anywhere and would be expected functionality, if set.
+	private static $can_be_root = false;
+}
+
+class SiteTreeTest_SpecificChild1 extends Page implements TestOnly {
+	// Only explicitly allow SiteTreeTest_SpecificParent1 (not its children).
+	private static $allowed_parents = array('*SiteTreeTest_SpecificParent1');
+	private static $allowed_children = array();
+}
+
+class SiteTreeTest_SpecificChild2 extends SiteTreeTest_SpecificChild1 implements TestOnly {
+	// Should also allow SpecificParent2 as a parent by proxy of inheritance.
+	// Should continue to show up as an allowed child of SpecificParent1 by proxy of inheritance.
+	private static $allowed_parents = array('SiteTreeTest_SpecificParent1');
+	private static $allowed_children = array();
 }
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -2696,7 +2696,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	}
 
 	/**
-	 * Returns an array of the class names of classes that are allowed to be children of this class.
+	 * Returns an array of the class names of classes that are allowed to be children of this class by proxy of each
+	 * class' defined ::$allowed_children and ::$allowed_parents properties.
 	 *
 	 * @return string[]
 	 */
@@ -2717,7 +2718,45 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 				}
 			}
 		}
-		
+
+		// Define allowed parents on a per page basis.
+		$allowedParents = array();
+		$pageTypes = SiteTree::page_type_classes();
+		foreach($pageTypes as $pageType) {
+			// Only worry about it if this page type has the private static ::$allowed_parents configured.
+			$curParents = singleton($pageType)->stat("allowed_parents");
+			if ($curParents === null) continue;
+			$allowedParents[$pageType] = $curParents;
+		}
+
+		// See if any one of the current allowed children exist in the array of pages above that restrict their parents.
+		$restrictedChildren = array_intersect($allowedChildren, array_keys($allowedParents));
+		if (!empty($restrictedChildren)) {
+			// Fetch all parent classes up to (and including) the final "Page" classes to see if this class will qualify
+			// as one of possible parents in the ::$allowed_parents array, which will include the specified classes and
+			// children thereof (notwithstanding any prefixed with the special "*").
+			$thisPage = get_class($this);
+			$candidateParents = array_values(ClassInfo::ancestry($thisPage));
+			$candidateParents = array_slice($candidateParents, array_search("Page", $candidateParents));
+
+			// Now add the "*" prefixed version of this class (only) so that it too will classify as a hit. This works
+			// because if a parent of this page is included but with the asterisk, it will not appear in this array; only
+			// if the current class is explicitly defined (since only the current class is defined both with/without the prefix)
+			$candidateParents[] = "*$thisPage";
+
+			// Go through each of the children with restricted parents and remove it from the list of children if this
+			// class and none of its parents appear in its list of allowed parents. Performed via quick intersection.
+			foreach($restrictedChildren as $childPage) {
+				if (count(array_intersect($candidateParents, $allowedParents[$childPage])) === 0) {
+					$childIndex = array_search($childPage, $allowedChildren);
+					unset($allowedChildren[$childIndex]);
+				}
+			}
+
+			// Re-key the array since some items may now have been removed.
+			$allowedChildren = array_values($allowedChildren);
+		}
+
 		return $allowedChildren;
 	}
 


tractorcow:
Would it be ok to have some tests with this please? :D

patricknelson:
Sure, you're referring to some demo code right? Here's what I had setup for myself:

**mysite/code/GenericPage.php**
```php
<?php

class GenericPage extends Page {
	// noop.
}
```

**mysite/code/SpecificParent.php**
```php
<?php

class SpecificParent extends Page {
	private static $allowed_parents = [];
	private static $allowed_children = ['SpecificChild'];
}
```

**mysite/code/SpecificChild.php**
```php
<?php

class SpecificChild extends Page {
	private static $allowed_parents = ['SpecificParent'];
	private static $allowed_children = [];
}
```

Or were you referring to unit tests?

dhensby:
We need unit test coverage and possibly even behat tests too

patricknelson:
Ok. I might need some guidance on how to do that for something like this. 

1. Would I be dynamically mocking up classes that extend `Page` and then somehow set their private static fields fields in the unit test itself? I suppose via fixtures (looking at some existing code) but didn't find any documentation or examples from what I've seen so far.

2. You'd want the tests in `cms/tests/model/SiteTreeTest.php` correct?

Thanks.

patricknelson:
I'm confused on how this is possible since it goes against intuition (maybe magic behind the scenes via `ReflectionClass`?) but since I know private statics are accessible externally via `->stat()` does that mean I'd just need to use the fixtures YML, setup my test to get an object from that fixture and then set the static fields via `->set_stat()` (I just found)? If so then I think I know what to do.

dhensby:
the stat functions should generally be avoided and `Config` used instead.

You want to:

1. Create some classes that have the `private static` vars set as you need them - these are typically defined at the end of the test file
2. Add a test functions to fetch those pages and attempt to relate them as parent-child as expected and ensure it works (no error when writing).
3. Add a test to check that pages that shouldn't work don't and throw an error as expected
4. Add a test to check edge cases, like this works on ancestor classes and so on.

patricknelson:
Good point @dhensby. I've revisited this and realized that it should obviously follow similar syntax and logic as the `::$allowed_children` array. So the functionality will be that if you specify a class in this array (without a `*` prefix) it will include that class and any of its children as being allowed as parents to the current class. However, if you do provide the `*` prefix (e.g. `*ThisClassOnly`) it will only allow just that class and none of its children. I'll update the code I've got now to match that and then setup test cases to cover:

1. Ensuring any page is allowed as a parent if no `::$allowed_parents` array is provided.
2. Ensuring that a specified parent page **will** have another page as a child if that candidate child *does* list either:
   * The parent page (explicitly with the `*` prefix), or
   * The parent page or one of it's children (with no prefix).
3. Ensuring that **no** page is allowed as a parent to a class if the `::$allowed_parents` array is provided and is empty. Practically, this means that the `->allowedChildren()` method will never include this class in its returned results.

patricknelson:
Ok, I've setup the test `->testAllowedParents()` on `SiteTreeTest` and so far it checks out. Let me know if you can think of any other scenarios that I may have missed (hopefully Travis CI checks out as well).

patricknelson:
Ping @dhensby or @tractorcow I don't want this to get stale :)

tests/model/SiteTreeTest.php
```diff
@@ -858,6 +858,25 @@ public function testAllowedChildren() {
...
 	}
 
+	public function testAllowedParents() {
+		$page = new Page();
+		$parent1 = new SiteTreeTest_SpecificParent1();
+		$parent2 = new SiteTreeTest_SpecificParent2();
+
+		// Ensure that none of the "Specific" parents and children show up as allowed children on a generic Page object.
+		$pageChildren = $page->allowedChildren();
+		$this->assertEquals(false, in_array("SiteTreeTest_SpecificParent1", $pageChildren));
```
dhensby: use `assertContains("SiteTreeTest_SpecificParent1", $pageChildren)` please

dhensby:
Thanks for the test coverage, that was my main concern, I think.

I'm really busy at the moment, but will try to check this out (literally) and give it a test... Though can't promise how soon that'll be

tests/model/SiteTreeTest.php
patricknelson: Ok, I'll use `assertNotContains()` since I'm trying to ensure it's not in the list.

patricknelson:
No problem, thanks for following up. I've tweaked the unit test to use the more aesthetic and practical
`$this->assertNotContains(..., ...)` method as well.

dhensby:
hey @patricknelson please can you rebase this and ping me when that's done? then I'll take a look :)

patricknelson:
Alright, @dhensby done. Sorry for the delay in basically everything else. I've been super busy and preparing for a long vaca starting Friday :D.

dhensby:
OK - I've checked this out locally and I think it's working as expected... but I was a bit confused by the following behaviour.

When I had a page with `allowed_parents` it could be a root page. I personally think this should STOP the page being added as a root page; there is of course the `can_be_root` config option (default true) but I feel this is a bit too un-expected.

We could get around this by demanding that `can_be_root` is explicitly set to true on the class in question.

@patricknelson The other thing we really need here is docs and perhaps more tests to assert this case and how we handle it? Behat tests would really be great too...

patricknelson:
Right, I understand the confusion. I think documentation would be called for here. The logic behind allowing it to still exist in the root even though `::$allowed_parents = []` is because technically the root is not a parent (since in this case literally `ParentID = 0`). Also, from a standpoint of separation of concerns, if I say "This page cannot have parents," then that's what I want, it doesn't mean I cannot save it in the root; just that I cannot have parents. 

Also note that this would *actually* create a conflict with `::$can_be_root` since if an empty array for `::$allowed_parents` means you cannot save it in the root, then it literally cannot exist anywhere, creating more work when all I wanted to do is prevent it from having parents. In this case, it makes more sense to me to explicitly say `::$can_be_root = false` and at that point, if I as a developer have an issue finding a way to save my page, it's easier to resolve (because then I've explicitly said it cannot be root as well as defining that it cannot have parents). 

Does that make sense?

patricknelson:
@dhensby  Also, side note: I have no clue how to use Behat and no prior knowledge on how to setup docs so neither of those will be today for sure (I'd be learning on the job on those). I can work on the other tests though. I'll ensure that you can still add `::$allowed_parents = []` to the root and maybe an alternate test case.

tests/model/SiteTreeTest.php
```diff
@@ -1226,3 +1260,28 @@ public function augmentValidURLSegment() {
 	}
 
 }
+
+class SiteTreeTest_SpecificParent1 extends Page implements TestOnly {
+	// Cannot have any parents at all.
+	private static $allowed_parents = array();
```
tractorcow: I think we allow and test `$allowed_parents = 'none'`. We added an explicit non-falsey 'none' value in order to overcome the behaviour inherit in config at not being good at distinguishing falsey values.

SiteTree adds a default `private static $allowed_children = array("SiteTree");` so perhaps we should be consistent and add the same for allowed_parents?

This would mean your code would always expect a value for allowed_parents, rather than treating empty values as an implicit 'None', and reduce the risk of getting the 'wrong' empty values for root-only pages. :)

tractorcow:
Sorry to prolong something that should have been solved a while back. I wish I got back to you earlier.

chillu:
Looks useful, although I'm a bit worried about complicating any logic which needs to be run for every tree node. How much data does this add to each `data-allowedchildren` HTML attribute?

I assume this works the same way than `allowed_children` in that it only applies to the direct parent in the tree? So you can still place a child into a structure where a grand parent node might not match `allowed_parents`.

Also, did you check that the tree drag/drop interactions still work as expected? From memory, I've used some jstree APIs to deny dropping nodes based on the `data-allowedchildren` values.

How about docs? I know `allowed_children` aren't well documented, but we shouldn't continue this trend with `allowed_parents`, right? ;) https://docs.silverstripe.org/en/3.1/developer_guides/customising_the_admin_interface/how_tos/customise_cms_tree/

dhensby:
@patricknelson sorry - I think I wasn't entirely clear.

I'm talking about when a page DOES have `allowed_parents` defined.

eg: `BlogPost` has `$allowed_parents = array('BlogHolder');` but blog page can still be at the root. Does allowed parents mean "these or root" or does it mean "can only be under these pages".

I think there are some scenarios we need to clear up:

1. Page has no allowed_parents defined (empty array/falsey value)
2. Page has allowed_parents set to "none" (proposal by @tractorcow)
3. Page has allowed_parents defined to one or more classes

For each scenario can the page be a root page and which pages can they be placed under?

I'd *personally* expect that if I define and "allowed_parent" then it means it has to be under one of those (ie: root page is not allowed).

I agree that if the allowed_parents is false then it'd be implicitly allowed at root level but that meaning it MUST be root, is different.

Personally, I'd prefer to add a 'must_be_root` option over implicitly forcing it through "none" value for `allowed_parents`

tractorcow:
>  Personally, I'd prefer to add a 'must_be_rootoption over implicitly forcing it through "none" value forallowed_parents`

Well, for child pages, `none` means `must_be_leaf`, so it's just being consistent. :)

dhensby:
Sure, I'm just saying I don't like this "none" value having a special meaning. I'd prefer that a config always has to be a single datatype

tractorcow:
Then what should be done with `allowed_children` to make it consistent? I'd still rather have an API that's consistent with the rest of the API. Too many special rules will make it hard to use. `None can be used with allowed_children, but for allowed_parents you need to set an empty array` is arbitrary.

dhensby:
> I suggest supporting both for both parameters, and in 4.x we can look at another solution.

agreed

JorisDebonnet:
Looks interesting! I actually just needed this :)

Does this PR need more work, or is it perhaps rebased somewhere else?

dhensby:
Hmm, we haven't come to a conclusion about inferred "can_be_root" or not...

I think keep this simple - if `can_be_root` is true, then it can be root even with defined `allowed_parents`.

Empty array or falsey value for `allowed_parents` means "can go anywhere"

`allowed_parents = "none"` means can't be a child page.

This work doesn't need to go against a different branch.

chillu:
@patricknelson ping

patricknelson:
Sorry about this -- I'll have to follow up on this in the coming weeks (not days); got a lot going on at the moment. Just letting you know so you'll be warned.

To be sure -- we're going to want to go with the implementation outlined by @dhensby here, correct? https://github.com/silverstripe/silverstripe-cms/pull/1171#issuecomment-148397150 If so I can move forward I think without any further instruction.

dhensby:
> To be sure -- we're going to want to go with the implementation outlined by @dhensby here, correct? https://github.com/silverstripe/silverstripe-cms/pull/1171#issuecomment-148397150 If so I can move forward I think without any further instruction.

@chillu @tractorcow what do you think?

tractorcow:
I like @dhensby 's solution yeah. ;D

Just be careful of falsey array values. I think allowed_children has a default value (defined in SiteTree) so it doesn't normally ever have the "is falsey check". It's ok if the apis differ a little there. ;)

I'm happy because we have the explicit "none" I was hoping for :D

tractorcow:
@patricknelson any update on this feature? I believe we were looking at 'none' option for 'allowed_parents'?

neilcreagh:
Bump - looks like this got so close!

JorisDebonnet:
I guess the question is whether @patricknelson is willing to finish this with the final root/none/... remarks. If not, perhaps someone else needs to make that final commit.

tractorcow:
If / when he comes back I'll re-open. ;P

patricknelson:
Could you re-open this, please? I believe it made mention of merge conflicts which is not visible now (cannot recall 100%).

patricknelson:
Oh ok I see why this has sat for so long. Some **TODO's** just for my reference (since there's such a huge backlog of comments and has sat for 3yrs):

### TODO:

- Fix merge conflicts
- Move away from `->stat()`
- Migrate to `$allowed_parents = 'none'` to follow current `$allowed_children = 'none'` syntax
- Possibly additional `->write()` tests to reiterate pages that shouldn't work still do not (already tested in retrieval in `->allowedChildren()`, but per @dhensby's [comment here](https://github.com/silverstripe/silverstripe-cms/pull/1171#issuecomment-76526993), maybe was missed?
- Verify and also initialize a `null` (or also empty, i.e. empty array?) definition of  `$allowed_parents` with `array("SiteTree")` if it's not defined


@dhensby @tractorcow @chillu  Please let me know if you happen to remember anything I've missed and I'll edit the list above.

---

**p.s.**

**Wrapping up on `$can_be_root`:**

@dhensby At this point I'm also reaffirming that no effort re: `$can_be_root` will be performed, in terms of code in this PR (per bullets above). I understand the confusion to a degree, but I know logically these two things (`$can_be_root` vs `$allowed_parents`) are unrelated in terms of implementation. Logically they come into conflict in a particular edge case scenario (i.e. `$can_be_root = false` while `$allowed_parents = "none"`), that requires no effort here. Just that we're aware that it could happen, particularly in cases of inheritance of these configs. 

> I think keep this simple - if can_be_root is true, then it can be root even with defined allowed_parents.

As I think this doesn't impact `->allowedChildren()` I presume I do nothing 😄

**Documentation:**

The best documentation I could find (that was documentation, not just a tutorial) was https://docs.silverstripe.org/en/3.0/reference/sitetree/ - What about v4? The search results [returned this hit ](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/how_tos/customise_cms_pages_list/) as the top hit for `allowed_children` for v4 and I don't find that as informative as the prior link.

Also re: documentation, @JorisDebonnet check out the weird hack from the link above:

```php
class Page extends SiteTree {
  // Don't let BlogEntry pages be underneath Pages.  Only underneath Blog holders.
  static $allowed_children = array("*Page,", "BlogHolder");
}
```

Coincidentally I think that would have been an ugly (but probably functional) workaround for a lack of `$allowed_parents` in the interim. Have you been using that? 😀

JorisDebonnet:
> Check out the weird hack from the link above: `static $allowed_children = array("*Page,", "BlogHolder");` 
> Have you been using that? 😀

I have used that to e.g. create `Blog` and `BlogPost` extensions named `ActivityOverview` and `Activity`, and make sure that 'regular' `Blog`s can only have plain `BlogPost`s, while `ActivityOverview`s can only have `Activity`s. That was the most important bit to me to make sure I have pretty green buttons "Add BlogPost" and "Add Activity" on top of their GridFields, instead of a dropdown. 

Wasn't obvious though because the yaml config didn't seem to work and I had to use the following to restrict Blogs to only have plain BlogPosts:
```php
Config::inst()->remove('Blog', 'allowed_children');
Config::inst()->update('Blog', 'allowed_children', ['*BlogPost']);
```

But I could still create `BlogPost`s and `Activity`s in various other undesired places though. So I have not yet used it on the `Page` class itself, but I figure that would indeed work fine if overwritten where needed. It would potentially require a long list in `Page` though, if you have many PageTypes, since they'd all need to be listed except for those with an "allowed_parent". Something like `class Page { static $allowed_children = ["Gallery", "Blog", "ActivityOverview", "ContactFormPage","*Page"]; }`. I guess that's why you'd call it an ugly workaround :) 

Anyway, I have only used this on SilverStripe 3.x, not in SS4 yet.

maxime-rainville:
Unfortunately we're only doing security updates to SilverStripe 3 from now on, so I'll have to close this one.

If you feel this is still a worthwhile feature, please refactor it so it works with SilverStripe 4.

PR #1172, opened 2015-03-02
NEW Update functionality around setMenuTitle

micmania1:
At the moment, if Title is the same as MenuTitle then MenuTitle will be set to null which I think is really odd behaviour and isn't what is expected. It also means you can't (easily) query using the MenuTitle column.

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -2674,11 +2674,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	 * @param string $value
 	 */
 	public function setMenuTitle($value) {
-		if($value == $this->getField("Title")) {
-			$this->setField("MenuTitle", null);
-		} else {
-			$this->setField("MenuTitle", $value);
-		}
+		Deprecation::notice(4.0, 'SiteTree::setMenuTitle is deprecated from 4.0');
+		$this->setField("MenuTitle", $value);
 	}
 	
 	/**
--- a/tests/model/SiteTreeTest.php
+++ b/tests/model/SiteTreeTest.php
@@ -900,25 +900,6 @@ class SiteTreeTest extends SapphireTest {
 		$this->assertContains('inherited-class', $treeTitle);
 	}
 
-	public function testMenuTitleIsUnsetWhenEqualsTitle() {
-		$page = new SiteTree();
-		$page->Title = 'orig';
-		$page->MenuTitle = 'orig';
-		$page->write();
-		
-		// change menu title
-		$page->MenuTitle = 'changed';
-		$page->write();
-		$page = SiteTree::get()->byID($page->ID);
-		$this->assertEquals('changed', $page->getField('MenuTitle'));
-
-		// change menu title back
-		$page->MenuTitle = 'orig';
-		$page->write();
-		$page = SiteTree::get()->byID($page->ID);
-		$this->assertEquals(null, $page->getField('MenuTitle'));
-	}
-
 	public function testMetaTagGeneratorDisabling() {
 		$generator = Config::inst()->get('SiteTree', 'meta_generator');
 


Zauberfisch:
generally +1, 
but not sure if deprecating setMenuTitle is the right step.

especially because (as far as I know) set{$fieldName} will be automatically called for every field by the CMS, so the CMS will still call setMenuTitle for MenuTitle, and I am not sure there is a sane way to tell the CMS to stop using setMenuTitle.

So I think we should either do 1 of the following 3 things:
- make setMenuTitle just call setField without the deprecation notice
- just remove setMenuTitle and catch calls to setMenuTitle in __call and throw a deprecation notice there
- just remove it and hope nobody notices? :P

micmania1:
The reason i've called deprecation is more so that when 4.0 comes around, there will be alerts so that devs will no to completely remove the setMenuTitle function. I guess I could make that change now against master but i'd rather wait and see what happens here.

micmania1:
I "fixed" the test by deleting it :nail_care: 

I've scoured the codebase and can't find anywhere in which this functionality is used, apart from the test. If this is accepted I'll just remove the method from the master branch.

It doesn't break backwards compatibility as nowhere that I can see relies on the MenuTitle where it is null.

dhensby:
I quite liked this functionality myself as it stood redundant storage of data. 

However, I accept that its a problem for querying on menu title. 

I don't think a deprecation notice should be used because the method is not deprecated and can still legitimately be used or defined. We are simply adjusting its functionality. 

Removing the function would have to be in master but we could remove that "feature" in 3. 

This is basically a method that is used to abstract the DB layer and so if we start storing it in the DB and removing this feature, it shouldn't have a big impact. 

One problem will be the fact that legacy DB will have null menu titles and without migrations could cause a problem... Perhaps we can only remove this from master

dhensby:
An alternative wpuld be to manipulate the sql query when searching against menu title with some kind of conditional logic.

Zauberfisch:
I have discussed the deprecation notice at length in irc with @micmania1 because I initially also thought its a bad idea, but I have come to accept it. it serves more as a reminder to remove it in 4.0 (because we will get deprecation notices all over the place when we ship 4). 
So effectively, the notice will never be thrown, because as soon as we see it, we will remove the whole method.
But for people browsing the api docs, it would be visible that its deprecated.

In fact, @micmania1 could you change the message to "will be removed in 4.0" and add a `@deprecated` phpdoc comment?

micmania1:
@dhensby the deprecation notice wouldn't do anything on the 3 branch. This would only trigger on the 4 (master) branch which is as you've said. I've only changed the functionality in the 3 branch and intend to remove it in the 4 branch, depending on feedback here.

The deprecation notice should scream at everyone on 4.0 if I forget to remove it which should serve as a pretty clear reminder.

Also, the getMenuTitle method still returns the Title field if no MenuTitle is set so you shouldn't need the conditional logic. The only case I can think of where you'd need conditional logic is with the current setup, and even then that could be wrong if the user has intentionally left the field blank for some reason.

I like less magic :)

dhensby:
> So effectively, the notice will never be thrown, because as soon as we see it, we will remove the whole method.

that's not the point of a deprecation notice. We can remove it NOW from master if we don't want people using it - if someone's building a site on dev-master then adding this deprecation notice won't help them, they'll either need to submit a PR to take the function out or they'll have to manually override the function on `Page`.

> But for people browsing the api docs, it would be visible that its deprecated.

That can be done in a docblock comment - but my point really is that this function is not being deprecated at all - it's still a legitimate function to use - it's just the behaviour of the function that is changing.

---

IMO, this is a change that can ONLY go into master and needs a section in the upgrade notes about it. People need to have a migration script too so that this can be fixed. If this kind of change were to get into 3 or 3.1 the it'd need to be a much less intrusive change (ie: changing the query so that `MenuTitle = MenuTitle || Title`)

micmania1:
I can make this change against 4.0 but I disagree about having a migration script. 4.0 allows for breaking backwards compatibility and i've already come across multiple instances where people are suggesting migration scripts - we're going to end up with more migration scripts than actual usable code - this is not the point of semantic versioning.

Also, I still don't get your point about calling `MenuTitle || Title`. That would only be applicable at the moment and in fact the getMenuTitle() method still falls back to Title if MenuTitle is empty.

+ At the moment, if I search MenuTitle for 'My Page' then I need to search the MenuTitle column and Title column because I can't guarantee MenuTitle has a value.
+ With this change, I only need to search MenuTitle because it will have a value. If it doesn't, then this is intentional and the result of the query will still be correct.

dhensby:
My point is that existing sites will be upgraded and need to have a clean way of going from a DB with no MenuTitle to one that has to have it.

At the very least, upgrade docs are required.

If you have a site now that your query is failing on, this change won't make it work as soon as you upgrade, it'll still break and you're site will be full of empty menu items. You'll need to migrate your data or you'll need to have some other work around; this case needs to be mitigated or handled.

micmania1:
In a migration script, how would we know whether the MenuTitle is empty because it was set to the same value as Title, or its empty because the user wants it to be?

Guidance (docs) on this is a must, but definitely no migration script.

I'll make the PR to master and include docs.

dhensby:
`UPDATE "SiteTree" SET "MenuTitle" = "Title" WHERE "MenuTitle" IS NULL;`

In an upgrade from current code it wouldn't be possible to have a MenuTitle that is empty because the user "wants it to be" it'd be the same as being the title. If the dev had overridden this behaviour intentionally, then they wouldn't run the migration.

Anyway, seems a bit academic now, but please include a snippet to that affect in your upgrade notes.

PR #1173, opened 2015-03-13
Updating deprecated functions

dhensby:
See https://github.com/silverstripe/silverstripe-framework/pull/3799

### DIFF
--- a/code/model/SiteTreeFileExtension.php
+++ b/code/model/SiteTreeFileExtension.php
@@ -31,8 +31,17 @@ class SiteTreeFileExtension extends DataExtension {
 			$rememberSubsiteFilter = Subsite::$disable_subsite_filter;
 			Subsite::disable_subsite_filter(true);
 		}
+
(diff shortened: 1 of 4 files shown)


dhensby:
@chillu any ideas on the behat errors?

tractorcow:
@dhensby did you mean to merge this into master? I don't think we were changing the api for those methods in 3.2.

tractorcow:
I've restarted this build not sure what's going on.

chillu:
The build was broken because there was a pretty obvious fatal error when opening `admin/assets` - easy to reproduce locally based on the Behat output, although it would be nicer to have seen it in a screenshot. For some reason the travis build tooling has lost the AWS key config, so is no longer uploading screenshots to S3.

kinglozzer:
> pretty obvious fatal error when opening `admin/assets`



Thanks @chillu

dhensby:
@chillu I think secure env variables aren't available to PRs / forks...

PR #1175, opened 2015-03-18
FIX: Prevent OldPageRedirector 404 redirect loop

wilr:
Redirect should check to ensure the new redirect location does not match the provided one. This was triggered when a website homepage was unpublished. Instead of throwing a 404 page the method previously caused a redirect loop. @cc ss23 #cwp

### DIFF
--- a/code/controllers/OldPageRedirector.php
+++ b/code/controllers/OldPageRedirector.php
@@ -10,21 +10,24 @@ class OldPageRedirector extends Extension {
 	 * @throws SS_HTTPResponse_Exception
 	 */
 	public function onBeforeHTTPError404($request) {
+
 		// Build up the request parameters
 		$params = array_filter(array_values($request->allParams()), function($v) { return ($v !== NULL); });
-
+		
 		$getvars = $request->getVars();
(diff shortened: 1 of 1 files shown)


ss23:
@wilr Tests are failing (looks like a legitimate failure)

code/controllers/OldPageRedirector.php
```diff
@@ -10,21 +10,24 @@ class OldPageRedirector extends Extension {
...
+		
 		$getvars = $request->getVars();
-		unset($getvars['url']);
-
+		
 		$page = self::find_old_page($params);
-
-		if ($page) {
+		
+		if ($page && $page !== $getvars['url']) {
```
tractorcow: Won't a baseurl of anything other than '/' break this check?

tractorcow:
@wilr did you have a new PR to open for this? :)

tractorcow:
Ah, I see this is a local branch on this repo. Would you like to back it up to your fork before I remove it?

dhensby:
Was fixed with #1515

PR #1177, opened 2015-03-19
BUG Fix SiteTree / SiteConfig permissions (3.0 backport version)

tractorcow:


code/model/SiteConfig.php
```diff
@@ -192,48 +205,70 @@ static public function make_site_config() {
 	 * called if a page is set to Inherit, but there is nothing
 	 * to inherit from.
 	 *
-	 * @param mixed $member 
+	 * @param Member $member
 	 * @return boolean
 	 */
-	public function canView($member = null) {
```
hafriedlander: You remove canView here, but don't later replace it. Is the default from DataObject enough? (You do replace canEdit later)

code/controllers/CMSPageAddController.php
```diff
@@ -21,7 +21,7 @@ public function AddForm() {
 		$pageTypes = array();
 		foreach($this->PageTypes() as $type) {
 			$html = sprintf('<span class="page-icon class-%s"></span><strong class="title">%s</strong><span class="description">%s</span>',
-				$type->getField('Title'),
+				$type->getField('ClassName'),
```
hafriedlander: What's the purpose of this change? Doesn't seem related to the rest of the patch?

tractorcow: It was an error which the later fixes to the .js relies on. ;)

code/model/SiteConfig.php
tractorcow: canView has been reinstated with a logical replacement.

PR #1180, opened 2015-03-30
Disable 'Add' button when the input field is empty

flashbackzoo:
Disable the 'add' button so users can't create empty rows. This requires https://github.com/silverstripe/silverstripe-framework/pull/4045

### DIFF
--- a/javascript/CMSMain.EditForm.js
+++ b/javascript/CMSMain.EditForm.js
@@ -347,6 +347,37 @@
 			}
 		});
 
+		/**
+		 * Don't allow users to submit empty values in grid field auto complete inputs.
+		 */
+		$('.cms-edit-form .ss-gridfield-buttonrow .add-existing-autocompleter').entwine({
+			onmatch: function () {
+				var self = this;
+
+				this.toggleDisabled();
+
+				this.find('input[type="text"]').bind('keyup', function () {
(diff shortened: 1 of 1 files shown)


kinglozzer:
Hmm, would this be better solved entirely at a GridField level in framework, rather than specific to this one form? I can’t imagine any scenario in which submitting an empty value for the autocomplete would be desirable

javascript/CMSMain.EditForm.js
```diff
@@ -347,6 +347,37 @@
...
+				this.toggleDisabled();
+
+				this.find('input[type="text"]').bind('keyup', function () {
+					self.toggleDisabled();
+				});
+			},
+			onunmatch: function () {
+				this.find('input[type="text"]').unbind('keyup');
+			},
+			toggleDisabled: function (value) {
```
dhensby: The `value` param doesn't seem to be used

dhensby:
I agree adding this to the cms edit form is to tightly coupled to a single use of the autocompleter

javascript/CMSMain.EditForm.js
flashbackzoo: Yup - I'l remove that.

flashbackzoo:
Yeah good call - I'll move it. Cheers guys.

Issue #1150, opened 2015-01-09
BUG Save and delete buttons showing on 'Files' section and throwing errors.


camfindlay:
I have noticed that if you go to the Files section of the CMS you initially get presented with a Save and Delete button which throw errors when pressed (a regression error here perhaps?).

![silverstripe files](https://cloud.githubusercontent.com/assets/367847/5687208/1493751e-98af-11e4-8746-4bad64e6b700.png)

If you then drill down into the File structure and back up (with the 'assets/show/#' action present - the buttons disappear (which I assume is the intend UI for the top level of the Files section?).

![silverstripe-files-2](https://cloud.githubusercontent.com/assets/367847/5687228/4ff5ebbe-98af-11e4-9411-7a2a24e5cba1.png)

micmania1:
Does this happen on a base install (ie no modules)?

camfindlay:
I'll test it and check @micmania1 :+1:

camfindlay:
@micmania1 yes it is a bug on a fresh checkout of 3.1.8

chillu:
Fixed by https://github.com/silverstripe/silverstripe-cms/pull/1186

camfindlay:
:+1:

PR #1186, opened 2015-04-16
Issue #1150: Removed Save and Delete buttons from regular edit form.

colintucker:
I think this should fix the issue; there is no need for those buttons to be there anyway except for on Root.DetailsView?  Can someone confirm whether or not this change will cause any unforeseen issues?


chillu:
Yes, that's the correct solution for https://github.com/silverstripe/silverstripe-cms/issues/1150, thanks Colin! The Behat failure seems to be unrelated, so merging as-is.

colintucker:
Sweet, thanks Ingo!

PR #1189, opened 2015-04-19
Description styles added to cms

A7DC:


tractorcow:
@A7DC you seem to have comitted the silverstripe-installer module code to cms. Would you like to make a new PR please?

PR #1190, opened 2015-04-20
added padding to the cms

A7DC:


tractorcow:
The accordion field is actually in the framework module not CMS so it probably should go there.

dhensby:
Please move this to the framework.

Thanks

PR #1191, opened 2015-04-21
Refactor ErrorPage to have static file building in a separate method

ss23:
Also changes to file_put_contents and single quotes for readability in some places

### DIFF
--- a/code/model/ErrorPage.php
+++ b/code/model/ErrorPage.php
@@ -227,34 +227,34 @@ class ErrorPage extends Page {
 	public function doPublish() {
 		parent::doPublish();
 
+		return $this->writeStaticPage();
+	}
+
+	/**
+	 * Write out the published version of the page to the filesystem
+	 *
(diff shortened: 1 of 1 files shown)


ss23:
scrutinizer error wasn't me pls merge

chillu:
Why is that necessary? 

BTW there's some discussion around removing the `ErrorPage` class from core on the mailinglist (for 4.x): https://groups.google.com/forum/#!topic/silverstripe-dev/VFXrusjAnKE

ss23:
When doing static publishing, the base URL is different, I need to re-trigger this particular function when the static publisher is called. This is the easiest way I can think of doing it.

With that in mind, it may need to be public, not protected... any objections to merging if I make it public?

chillu:
Good enough of a reason :) Changed to public and pushed as 857ade7d1d37c360cfc1f78cfe7c68a817343f92

PR #1192, opened 2015-04-23
Fix for when you access a CMS admin page using the "show" action

oilee80:
When accessing the CMS page via the "show" action the body class is "ArrayData" instead of having "CMSMain" this means the selector for "Add Page" doesn't match and the "onclick" event doesn't trigger.


chillu:
It's unclear what commits are actually relevant here, and we no longer maintain 2.4. Closing for now.

PR #1193, opened 2015-04-24
Update from the original

mediaclinic:


chillu:
Please read http://doc.silverstripe.org/en/contributing/code/

PR #1194, opened 2015-04-27
Check $dataRecord existance in ContentController stage tests

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/3131.

Submitting against 3.1 because this is a fairly common feature ("anonymous" page controllers). There's a latent risk of causing a security issue due to wrong permission checks, so would appreciate some peer review.

Steps to reproduce:

1. Create controller as follows:

```php
class My_Controller extends ContentController {
    static $allowed_actions = array('mymethod');

    public function mymethod() {
        die('here');
    }
}
```

2. Login to CMS and switch a preview to draft mode

3. Open `/My_Controller/mymethod`

You'll see an "access denied" message.

### DIFF
--- a/code/controllers/ContentController.php
+++ b/code/controllers/ContentController.php
@@ -89,6 +89,8 @@ class ContentController extends Controller {
 	public function init() {
 		parent::init();
 
+		$recordExists = ($this->dataRecord && $this->dataRecord->exists());
+
 		// If we've accessed the homepage as /home/, then we should redirect to /.
 		if(
 			$this->dataRecord
@@ -128,7 +130,8 @@ class ContentController extends Controller {
 
 		// Draft/Archive security check - only CMS users should be able to look at stage/archived content
 		if(
-			$this->URLSegment != 'Security' 
+			$recordExists
+			&& $this->URLSegment != 'Security' 
 			&& !Session::get('unsecuredDraftSite') 
 			&& (
 				Versioned::current_archived_date() 
--- a/code/controllers/ContentController.php
+++ b/code/controllers/ContentController.php
@@ -95,12 +95,11 @@ class ContentController extends Controller {
 		if(
 			$recordExists
 			&& RootURLController::should_be_on_root($this->dataRecord)
-			&& (!isset($this->urlParams['Action']) || !$this->urlParams['Action'])
-			&& !$_POST
-			&& !$_FILES
+			&& !$this->request->param('Action')
+			&& !$this->request->postVars()
 			&& !$this->redirectedTo()
 		) {
-			$getVars = $_GET;
+			$getVars = $this->getRequest()->getVars();
 			unset($getVars['url']);
 			if($getVars) {
 				$url = "?" . http_build_query($getVars);
--- a/code/controllers/ContentController.php
+++ b/code/controllers/ContentController.php
@@ -93,8 +93,7 @@ class ContentController extends Controller {
 
 		// If we've accessed the homepage as /home/, then we should redirect to /.
 		if(
-			$this->dataRecord
-			&& $this->dataRecord instanceof SiteTree
+			$recordExists
 			&& RootURLController::should_be_on_root($this->dataRecord)
 			&& (!isset($this->urlParams['Action']) || !$this->urlParams['Action'])
 			&& !$_POST
@@ -113,7 +112,7 @@ class ContentController extends Controller {
 		}
 
 		// Run extension hooks
-		if($this->dataRecord) {
+		if($recordExists) {
 			$this->dataRecord->extend('contentcontrollerInit', $this);
 		} else {
 			singleton('SiteTree')->extend('contentcontrollerInit', $this);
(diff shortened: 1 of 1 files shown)


nathanbrauer:
cc @tylerkidd

code/controllers/ContentController.php
```diff
@@ -87,42 +88,59 @@ public function Page($link) {
...
+		$recordExists = ($this->dataRecord && $this->dataRecord->exists());
+
 		// If we've accessed the homepage as /home/, then we should redirect to /.
-		if($this->dataRecord && $this->dataRecord instanceof SiteTree
-			 	&& RootURLController::should_be_on_root($this->dataRecord) && (!isset($this->urlParams['Action']) || !$this->urlParams['Action'] ) 
-				&& !$_POST && !$_FILES && !$this->redirectedTo() ) {
+		if(
+			$recordExists
+			&& RootURLController::should_be_on_root($this->dataRecord)
+			&& (!isset($this->urlParams['Action']) || !$this->urlParams['Action'])
```
kinglozzer: `$this->request->param('Action');`?

code/controllers/ContentController.php
```diff
@@ -87,42 +88,59 @@ public function Page($link) {
...
+			&& !$_POST
+			&& !$_FILES
```
kinglozzer: Could we just check `$this->request->postVars()`? IIRC `$_FILES` is merged into `$_POST` for this. Appreciate you’re just tidying, not refactoring though ;)

code/controllers/ContentController.php
```diff
@@ -87,42 +88,59 @@ public function Page($link) {
...
 			$getVars = $_GET;
 			unset($getVars['url']);
-			if($getVars) $url = "?" . http_build_query($getVars);
-			else $url = "";
+			if($getVars) {
+				$url = "?" . http_build_query($getVars);
+			} else {
+				$url = "";
+			}
 			$this->redirect($url, 301);
```
kinglozzer: Could we utilise `SS_HTTPRequest::getURL()` here? As that strips out `$_GET['url']`

chillu: It strips out `'url'`, but leaves in the path, so can't use it.

chillu:
@kinglozzer Implemented two of your three suggestions. Good to merge (once build passes)?

chillu:
@tractorcow Want to peer review to see if you can identify any security risks?

stevie-mayhew:
Rather than `$this->request` can we call `$this->getRequest()`. There may be instances where a developer has over written the `getRequest` function for whatever reason and if we have getters and setters we should use them :smile:

chillu:
There's 68 matches for `$this->request` in core, and 2 for `$this->getRequest()`. You won't have much look overwriting getters on this already. We could make this possible, but that should be a larger push across framework rather than piecemeal in this pull request.

stevie-mayhew:
Fair enough - I'll open a PR elsewhere for this change

code/controllers/ContentController.php
tractorcow: Could we use `$getVars = $this->getRequest()->getVars()` instead of accessing `$_GET` directly?

code/controllers/ContentController.php
```diff
@@ -87,42 +88,58 @@ public function Page($link) {
 
 	public function init() {
 		parent::init();
-		
+
+		$recordExists = ($this->dataRecord && $this->dataRecord->exists());
```
tractorcow: On the security page, we create a new `Page` (not `SiteTree`) record, and don't save it to the database. This will cause $recordExists to be false.

Have you checked that this doesn't interfere with any actions on the Security controller? I'm not saying that there are regressions, but there's a chance for them to occur here.

code/controllers/ContentController.php
```diff
@@ -87,42 +88,58 @@ public function Page($link) {
...
-		
-		if($this->dataRecord) $this->dataRecord->extend('contentcontrollerInit', $this);
-		else singleton('SiteTree')->extend('contentcontrollerInit', $this);
 
-		if($this->redirectedTo()) return;
+		// Run extension hooks
+		if($recordExists) {
+			$this->dataRecord->extend('contentcontrollerInit', $this);
+		} else {
+			singleton('SiteTree')->extend('contentcontrollerInit', $this);
```
tractorcow: See my note above on the Security controller. Do you think extending `SiteTree` over `Page` is a good idea? Since we are in the CMS module it's probably ok to rely on there being a Page class.

tractorcow:
I left a few comments, but in general nothing that I feel is a security issue or blocker to this. I wouldn't complain if this was merged as is (after a squash that is. ^^)

code/controllers/ContentController.php
tractorcow: Honestly I can't see where it could, after looking through this all again. I think it's fine, but check my note below on extensions on the dataRecord. :D

code/controllers/ContentController.php
chillu: Done

chillu:
@tractorcow I've confirmed `/Security` operation, login and logout work as expected. Don't think we need a squash, all these commits make sense individually (no noise from whitespace and alignment changes in the actual feature commit here). So ready to merge?

sminnee:
I attempted to rebase this to 3.4 but it seems to apply to code that doesn't exist anymore because of 5353ac5315703240540c9cde0f5c8eeb5571bc19.

sminnee:
The underlying issue appears to have been resolved, so I'm closing this.

PR #1195, opened 2015-04-28
Improve UX of asset management

flashbackzoo:
This is an improvement to the UX of asset management, specifically the 'used on' field.

used-on-page-closed

User clicks the 'info' icon...
used-on-page

### DIFF
--- a/code/model/SiteTreeFileExtension.php
+++ b/code/model/SiteTreeFileExtension.php
@@ -10,11 +10,39 @@ class SiteTreeFileExtension extends DataExtension {
 	);
 
 	public function updateCMSFields(FieldList $fields) {
-		$fields->insertAfter(new ReadonlyField('BackLinkCount', 
-			_t('AssetTableField.BACKLINKCOUNT', 'Used on:'), 
-			$this->BackLinkTracking()->Count() . ' ' . _t('AssetTableField.PAGES', 'page(s)')), 
+		$fields->insertAfter(
+			ReadonlyField::create(
+				'BackLinkCount', 
+				_t('AssetTableField.BACKLINKCOUNT', 'Used on:'), 
+				$this->BackLinkTracking()->Count() . ' ' . _t('AssetTableField.PAGES', 'page(s)'))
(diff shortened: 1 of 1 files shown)


code/model/SiteTreeFileExtension.php
```diff
@@ -10,14 +10,43 @@ class SiteTreeFileExtension extends DataExtension {
...
+	public function BackLinkHTMLList() {
+		$html = '<em>This list shows all pages where the file has been added through a WYSIWYG editor.</em>';
+
+		$html .= '<ul>';
+
+		foreach ($this->BackLinkTracking() as $backLink) {
+			$listItem = '<li>';
+
+			// Add the page link
+			$listItem .= '<a href="' . $backLink->Link() . '" target="_blank">' . $backLink->MenuTitle . '</a> &ndash; ';
```
tractorcow: You need to XML encode this title, otherwise you could get HTML entities in your text.

code/model/SiteTreeFileExtension.php
```diff
@@ -10,14 +10,43 @@ class SiteTreeFileExtension extends DataExtension {
...
+
+			// Add the CMS link
+			$listItem .= '<a href="' . $backLink->CMSEditLink() . '">Edit</a>';
```
tractorcow: Can you please make edit translatable?

```
_t(`SiteTreeFileExtension.EDIT`, 'Edit')
```

code/model/SiteTreeFileExtension.php
```diff
@@ -10,14 +10,43 @@ class SiteTreeFileExtension extends DataExtension {
...
 		);
 	}
 
 	/**
+	 * Generate an HTML list which provides links to where a file is used.
+	 *
+	 * @return String
+	 */
+	public function BackLinkHTMLList() {
+		$html = '<em>This list shows all pages where the file has been added through a WYSIWYG editor.</em>';
```
tractorcow: As below, please make this string translatable.

Issue #1197, opened 2015-05-01
RedirectorPage still redirects even if URL is invalid


patricknelson:
For example, if you setup a redirector page at `/base-section` which is intended to redirect to `/base-section/overview`, but a user visits an invalid page URL such as `/base-section/old-page` it will _always_ redirect that request to `/base-section/overview`, even if that URL is `/base-section/very/long/with/many/slashes`. This was a problem with my site (and fixed pretty quickly once I tracked down what was causing it) because this base section had a lot of pages underneath it that all shifted to be one level deeper, making this quirk more apparent. Please note that this is just one example, but I suspect that there may be other situations where this result would not be quite as helpful

Looking deeper, this may be due to the [Redirector::init()](https://github.com/silverstripe/silverstripe-cms/blob/3.1/code/model/RedirectorPage.php#L164) method not checking to see if an action was provided. In my opinion, if you setup a redirector page at a specific URL and the user enters a URL that doesn't match that page's URL, it should continue past ::init() and process as usual, therefore resulting in an expected 404.

maxime-rainville:
Confirmed fixed on Silverstripe 4.

Unfortunately, [Silverstripe CMS 3 has entered limited support in June 2018](https://www.silverstripe.org/blog/update-on-silverstripe-5-x/). This means we'll only be fixing critical bugs and security issues for Silverstripe CMS 3 going forward.

You can read the [Silverstripe CMS Roadmap](https://www.silverstripe.org/software/roadmap/) for more information on our support commitments.

PR #1198, opened 2015-05-01
FIX for #1197 Ensuring the RedirectorPage_Controller does not inadvertently redirect URL's which don't actually match it.

patricknelson:
Proof of concept which addresses #1197.

### DIFF
--- a/code/model/RedirectorPage.php
+++ b/code/model/RedirectorPage.php
@@ -161,7 +161,9 @@ class RedirectorPage_Controller extends Page_Controller {
 	public function init() {
 		parent::init();
 
-		if($link = $this->redirectionLink()) {
+		// Redirect this request, but only if no additional action (slug) has been provided.
+		$action = $this->request->param('Action');
+		if($action === null && $link = $this->redirectionLink()) {
 			$this->redirect($link, 301);
 			return;
 		}


chillu:
You're assuming here that the `action` parameter is a nested page, right? What if its a method on the `Page_Controller`? I'd expect that to still redirect. Based on the outcome of this, the pull request might get classified as a feature change rather than a bugfix.

Also, we'll need some tests for this new behaviour :)

patricknelson:
Yes, I was. The idea was that I didn't have anything setup on that URL (no "allowed actions here") and yet it was still redirecting. We only had it setup for one URL yet it was inexplicably redirecting everything underneath it. 

What you're saying suggests to me that this instead should ensure that the specified action *does* exist and, if it does,  first check to see if it's on a parent controller and only then if it is, explicitly redirect, and if it's instead on itself or a child controller, then allow it to proceed (to facilitate overriding behavior via inheritance). It may be best to take that approach rather than overriding `->redirectionLink()` since you would want to keep with the standard SilverStripe'isms already put into place (for much needed consistency). 

The problem here is that if you go to *any other type of page* and you requested the `foobar` action and that action weren't setup anywhere nor was it a `URLSegment` on a child page, then it would proceed to a 404. In the case of a `RedirectorPage` that doesn't happen and that's a bothersome inconsistency to me. If I wanted a special redirector that redirected *everything* underneath a URL then I'd set it up to do so! And that *could* be useful but I do not think it should be default in in way! So, even though this is current functionality, I think it's true that SS users may currently be using it (while I consider it an inconsistency). 

**That** being said: If we make this a feature, can I set it up as above and instead upgrade `RedirectorPage` to allow the option to redirect all sub-URL's with the default being to only redirect on an exact match? Or could we roll this out as a `3` or even `3.1` change with the *inverse* default (i.e. default to current state of "redirect all the things!")?

chillu:
By limiting redirects to "action-less" URLs, you introduce new edge cases (see below). As you mention, these edge cases force you to determine the actions present on the redirection target, so complicating the logic. How about we keep the redirection unconditional, and pass through action in `redirectLink()`? In this case the redirection target can decide how to handle the action (such as showing a child page or a 404). That seems like a bugfix-level behaviour change.

Current behaviour:
 * Redirect without action (`page/` -> `redirected/`)
 * Drop child page on redirect (`page/child-page/` -> `redirected/`)
 * Drop action on redirect (`page/action/` -> `redirected/`)
 * Drop non-existing action on redirect (`page/non-existing-action/` -> `redirected/`)

Pull request behaviour:
 * Redirect without action (`page/` -> `redirected/`)
 * Don't redirect with child page (`page/child-page/` -> shows child page)
 * Don't redirect with existing action (`page/action/` -> shows action on RedirectorPage_Controller, which doesn't make sense)
 * Don't redirect on non-existing action (`page/non-existing-action/` -> page not found)

Suggested behaviour:
 * Redirect without action (`page/` -> `redirected/`)
 * Redirect to child page (`page/child-page/` -> `redirected/child-page`)
 * Redirect to new page with action (`page/action/` -> `redirected/action`)
 * Redirect to new page with non-existing action, but show "page not found" (`page/non-existing-action/` -> `redirected/non-existing-action`)

patricknelson:
Actually, as it turns out, a child page attached to a redirector page works fine as expected (bullet 2 above). That is, if you attached a child page to a redirector page, it will show the child page. Side note: This is a very good thing (e.g. "Our Company" redirecting to "Overview": `/our-company/` going to `/our-company/overview/`) but I digress.

On the pull request behavior:

    Don't redirect with existing action (page/action/ -> shows action on RedirectorPage_Controller, which doesn't make sense)

That was part of the original issue with my initial PR since I didn't think about the impact on actions (just that it was my route to fixing my particular issue at the time). However, thinking this through it gets a bit more complicated. Here's a set of possible circumstances and how I wouldrecommend they should result, depending on the initial conditions (let me know what you think):

* Given I have a redirector `page/` pointing to `redirected/` or `https://www.google.com` and user visits `page/segment` **and** `page/segment` is not an actual child page (or another redirector):
  * If `page/` points to a **remote** URL: Return 404 error, as no actions are allowed on the default redirector (as suggested)
  * If `page/` points to a **local** URL: Always redirect to destination and include any subsequent trailing segments. 

Personally, I think this is the most sensible default functionality (instead of just redirecting all requests to the same URL, no matter what). This is because:

* Always redirecting to the remote URL (as it does currently) will take a user off site, possibly unintentionally, which may be undesirable and at that point you lose control.
* Redirecting to a remote site while including trailing segments is potentially unsafe.
* Redirecting to a local page while always including the trailing segments will allow for the proper desired execution of either actions or loading of a child page at the remote location.

What do you think?

chillu:
Yep, didn't consider remote URLs. Your logic make sense to me.

patricknelson:
Ok. Which branch should this be based on?

chillu:
Sorry for the radio silence, Patrick! If you are planning to implement my "Suggested behaviour" items plus the remote URL case you laid out, I think this could be treated as a bugfix. It's a deep enough change that I wouldn't want this in 3.1 though, so please re-raise a PR against 3.2. You'll need to update the 3.2.0 changelog (briefly describe the altered redirect behaviour), describe the different cases in PHPDoc on `RedirectorPage`, and add decent test coverage. Ideally we'd have some markdown docs on the behaviour as well (e.g. in http://docs.silverstripe.org/en/developer_guides/controllers/redirection/), but that's not required to get the PR merged.

@stevie-mayhew Roping you in to this discussion, keen to hear your take :) Do you think this would cause any upgrade pains for projects you've been working on when going to 3.2?

patricknelson:
Bump this, since I just realized this was still open. Forgot it since I think I was able to workaround the bug using the injector. Now that `3.2` is released, I'll need advice on which branch I should reopen a PR on. In this case maybe `3.2` still but probably best now to do it against `3` or `master`.

 @chillu @stevie-mayhew

stevie-mayhew:
Sorry didn't even notice that mention months ago!

We have quite a lot of complex actions in some of our applications but we never have a need to redirect to that action, only the base page so I think this is fine.

I would put this against 3.

dhensby:
Bump?

code/model/RedirectorPage.php
```diff
@@ -161,7 +161,9 @@ class RedirectorPage_Controller extends Page_Controller {
 	public function init() {
 		parent::init();
 
-		if($link = $this->redirectionLink()) {
+		// Redirect this request, but only if no additional action (slug) has been provided.
+		$action = $this->request->param('Action');
+		if($action === null && $link = $this->redirectionLink()) {
```
tractorcow: The same effect could be achieved much easier by putting this into the index() method.

patricknelson: I'll post a comment in the main thread - stand by.

patricknelson:
So @tractorcow in response to your comment here: https://github.com/silverstripe/silverstripe-cms/pull/1198#discussion_r57825075

What we wanted to do was more advanced than what you just commented on. But yes, I think you're right, this could probably be done instead in `index()` however the trailing information may still be useful (see per the discussion above) and I think that information would be lost (404'ed?) if you just put the redirect code specifically in the `index()` but I'm not positive about that. I believe I rewrote the code already (many moons ago) but never got around to posting an updated PR (and still don't really have time to get my environment updated to do so, even now). 

Anyway, what I ended up doing in my own code in the interim was use the `Injector` to create my own instance of the RedirectorPage_Controller`. Here's what I have now at the moment:

```php
class CustomRedirectorPage_Controller extends Page_Controller {

	public function init() {
		parent::init();

		// Redirect this request.
		if($link = $this->redirectionLink()) {
			// See if the link is external or not. If it isn't, try to retain trailing information in the current request URI.
			if (Director::is_site_url($link)) {
				// Check regular link to verify it matches the base of the requested URL before parsing out trailing info.
				$regularLink = trim($this->regularLink(), "/");
				$requestedLink = trim($this->request->getURL(true), "/");
				if (substr($requestedLink, 0, strlen($regularLink)) === $regularLink) {
					// Get trailing information.
					$trailing = trim(substr($requestedLink, strlen($regularLink)), "/");

					// Ensure current redirect link is also root relative and then suffix trailing info.
					$link =  trim(Director::makeRelative($link), "/");
					$link = "/$link/$trailing";
				}
			}

			// Perform redirect after any modifications (if applicable).
			$this->redirect($link, 301);
			return;
		}
	}

}
```

See what I mean? So per my points located here (https://github.com/silverstripe/silverstripe-cms/pull/1198#issuecomment-98421579) this is the agreed logic thus far:

 Given I have a redirector `page/` pointing to `redirected/` or `https://www.google.com` and user visits `page/segment` **and** `page/segment` is not an actual child page (or another redirector):
  * If `page/` points to a **remote** URL: Return 404 error, as no actions are allowed on the default redirector (as suggested)
  * If `page/` points to a **local** URL: Always redirect to destination and include any subsequent trailing segments. 

**However,** what I already implemented above doesn't yet happen to incorporate any 404 logic I believe. I didn't worry about that in my personal implementation since I didn't need that per se (plus I don't know how to implement that properly off hand). 

**Questions:**
- Do you know off hand how I can properly throw a 404 error from the `init()` method in this controller so I can roll that into my new PR?
- Do you think my new PR should still be based on `3` per @stevie-mayhew (https://github.com/silverstripe/silverstripe-cms/pull/1198#issuecomment-151719987)?

tractorcow:
This is very confusing.

patricknelson:
Hah, ok :joy: I can agree. So think of it this way. 

Say you currently have this site tree

- products
  - feline-toys
    - yarn
    - laser
  - pooch-toys
    - bone
    - ball

And you want to redirect `feline-toys` to `cat-toys` and `pooch-toys` to `dog-toys` for SEO reasons. So you manually setup a few redirects like so (assuming the CMS didn't already do this for you, just for demonstration sake):

- products
  - feline-toys (redirector page which to `cat-toys` page below)
  - cat-toys (renamed)
    - yarn
    - laser
  - pooch-toys  (redirector page which points to `http://www.pooch-toys-are.us/`)

The current functionality I believe is to redirect both requests going directly to the old redirected page at `products/feline-toys` as well as other sub-URL's such as `products/feline-toys/yarn` to `products/cat-toys`. What I'm suggesting is that the new functionality be the following:

- Requests to `products/feline-toys` redirect to `products/cat-toys` as expected.
- Requests to `products/dog-toys` would redirect to `http://www.pooch-toys-are.us/` as expected.
- Requests to `products/feline-toys/yarn` redirect to `products/cat-toys/yarn` (new). This is done by taking the entire trailing segment past the base redirector page URL and appending that to the new destination URL for **local redirects** only. If you requested `products/feline-toys/ball` (which never existed anyway) it would redirect the client browser to `products/dog-toys/ball` which *itself* would then 404.
- Requests to `products/dog-toys/yarn` would 404. The intent here is to prevent these pages from becoming an [open redirect](https://www.owasp.org/index.php/Open_redirect).

Does that make it a little less confusing, at least? (It was a lot to type so I hope so :smile:)

tractorcow:
>  Requests to products/feline-toys/yarn redirect to products/cat-toys/yarn (new). This is done by taking the entire trailing segment past the base redirector page URL and appending that to the new destination URL for local redirects only. If you requested products/feline-toys/ball (which never existed anyway) it would redirect the client browser to products/dog-toys/ball which itself would then 404.

What about making that yarn redirection only occur if cat-toys/yarn doesn't exist? As in, keep trailing url sections, but only if there is no other page matching.

I would expect sub-pages of redirector pages to work as per normal, but I wouldn't mind if non-existant sub-pages were treated as url segments to append to the forwarded url.

This behaviour seems very use-case specific though... I wonder if there should be a flag for allowing forwarding of subpages as well as the current page.

patricknelson:
@tractorcow  Yes, good point. I didn't mention that scenario (i.e. if the requested URL actually does exist, it shouldn't attempt to redirect). In that case, it should just load the actual page at that URL. In fact, we have nested redirects on our site as well and this works fine. I've already implemented much of this functionality (except for any of the 404 logic).

patricknelson:
@tractorcow It's not entirely use-case specific in my opinion, only partially. That is, what's **currently** happening is not good. For example:

If I setup a redirect at `products/feline-toys`, I currently expect it to NOT redirect products/feline-toys/some/sub/page/20/levels/deep` but instead simply 404. I didn't setup a redirect for that page. I setup a redirect just for `products/feline-toys`!

If, however, the system redirected to the new (local) location and bundled that sub-URL along with it, that would 404 as well! Which is fine, because that suits both scenarios. i.e. On one hand I may not have any URL underneath the page I was originally redirecting that matched the aforementioned URL (likely more common)  but, on the other hand, I may also have had many other sub-URLs under that page that I'd like to redirect (e.g. maybe the CMS isn't redirecting it properly or I setup a brand new website and want to do a broad sweeping redirect).

In either case -- if you're suggesting the redirect functionality which includes the sub-URL (or action or what-have-you) is too use-case specific, I'd suggest going with that checkbox since, for me, that'd be super useful but I could see how that may rustle jimmies of other users/devs. As long as it's an option.

Also, importantly, I think we should never allow external redirects to include sub-URL's just for security purposes (instead require a dev to code that out his/herself due to the implications).

patricknelson:
Relevance here is also an important issue. If indeed a user did try to visit some specific sub-page that didn't exist anymore (or maybe never did) and will never exist, I don't think it makes sense to continue with current functionality to redirect that user to the same destination page for that longer sub-URL as the root version of the source URL's destination.

patricknelson:
Similar to what was mentioned [here](https://github.com/silverstripe/silverstripe-framework/pull/4096#issuecomment-305272481): In _this case_, I do have some workarounds in place which I think would be useful in core (my mods are via extensions).

I still think the discussion associated with this PR should be resolved. Currently I've got my own modifications which suit my needs, but there may be a relatively simple answer:

- Add radio buttons to either disallow or allow enabling/disabling of redirecting in case trailing information is detected on the source URL (e.g. `redirect/` but URL visited is `redirect/trailing?abc=123`) with two options:
  - Allow trailing but don't retain trailing information (e.g. via appending to destination URL)
  - Allow trailing and retain at the destination.
- Add YAML configuration directives for:
  - Change default selection for the trailing radio/option when creating new `RedirectorPage`'s. Default (for backward compatibility) should be consistent with current functionality (to allow trailing information but not retain trailing it when redirecting).
  - Allow/disallow functionality for URL's that are not pointing to the local site (e.g. remote URL's, in case this opens up to some sort of redirect proxy vulnerability). Default value here I think should be to disallow.

That way we allow developers the flexibility to control what happens without dictating it to them under the complexity of each individuals' implementation.

tractorcow:
Was auto-closed due to removal of 3.4 branch. Re-opened.

dhensby:
It would be good to get this over the line - is there any motivation to do this?

maxime-rainville:
Unfortunately we're only doing security updates to SilverStripe 3 from now on, so I'll have to close this one.

If you feel this is still a worthwhile feature, please refactor it so it works with SilverStripe 4.

maxime-rainville:
I've had a look if this could be rebase in SS4. RedirectPageController has changed too much for the PR.

PR #1199, opened 2015-05-02
Code improvements to various classes to remove unused and unnecessary code

Matthew-Bonner:
Code improvements to various classes to remove unused and unnecessary code

### DIFF
--- a/code/controllers/AssetAdmin.php
+++ b/code/controllers/AssetAdmin.php
@@ -7,15 +7,6 @@
  * @subpackage assets
  */
 class AssetAdmin extends LeftAndMain implements PermissionProvider{
-
-	private static $url_segment = 'assets';
-	
-	private static $url_rule = '/$Action/$ID';
-	
-	private static $menu_title = 'Files';
-
-	private static $tree_class = 'Folder';
-	
 	/**
 	 * Amount of results showing on a single page.
 	 *
@@ -23,28 +14,6 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider{
 	 * @var int
 	 */
 	private static $page_length = 15;
-	
-	/**
-	 * @config
-	 * @see Upload->allowedMaxFileSize
-	 * @var int
-	 */
-	private static $allowed_max_file_size;
-	
-	private static $allowed_actions = array(
-		'addfolder',
-		'delete',
-		'AddForm',
-		'DeleteItemsForm',
-		'SearchForm',
-		'getsubtree',
-		'movemarked',
-		'removefile',
-		'savefile',
-		'deleteUnusedThumbnails' => 'ADMIN',
-		'doSync',
-		'filter',
-	);
 	
 	/**
 	 * Return fake-ID "root" if no ID is found (needed to upload files into the root-folder)
@@ -270,8 +239,7 @@ JS
 			)->addExtraClass('cms-content-toolbar field'),
 			$gridField
 		));
-		
-		$treeField = new LiteralField('Tree', '');
+
 		// Tree view
 		$fields->addFieldsToTab('Root.TreeView', array(
 			clone $actionsComposite,
@@ -431,7 +399,6 @@ JS
 	}
 	
 	public function AddForm() {
-		$folder = singleton('Folder');
 		$form = CMSForm::create( 
 			$this,
 			'AddForm',
@@ -591,7 +558,6 @@ JS
 			array('count' => $count)
 		);
 		$this->response->addHeader('X-Status', rawurlencode($message));
-		return;
 	}
 	
 	/**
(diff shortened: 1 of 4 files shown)


Matthew-Bonner:
Rather than squashing the commits, it makes sense in my view to keep the commits as because I'm working through the files one by one.

kinglozzer:
Hey @Matthew-Bonner. It looks like these `private static` properties aren’t being used, but they’re actually picked up by the Config system and used in a few places. To take the first example from the test failures: [this property](https://github.com/silverstripe/silverstripe-cms/blob/3.1/code/controllers/CMSMain.php#L26) is accessed [here](https://github.com/silverstripe/silverstripe-cms/blob/3.1/code/controllers/CMSMain.php#L512). The `$allowed_actions` are used by `RequestHandler` to determine which URLs can be accessed on a Controller too.

The other bits of tidying up (removing unused variables from methods, deleting commented-out code etc) seem fine, though. It might be easier to close this, start afresh and open a new pull request, rather than undo those `private static` deletions. But I’ll leave that up to you :)

code/controllers/CMSMain.php
```diff
@@ -1196,16 +1150,23 @@ public function BatchActionList() {
 	
 	public function buildbrokenlinks($request) {
 		// Protect against CSRF on destructive action
-		if(!SecurityToken::inst()->checkRequest($request)) return $this->httpError(400);
+		if(!SecurityToken::inst()->checkRequest($request)) {
+      $this->httpError(400);
+      return;
+    }
```
kinglozzer: We actually don’t even need the `return` statement here as `RequestHandler::httpError()` throws an `SS_HTTPResponse_Exception`, so it’s redundant. We could keep this on one line and just have:
```php
if(!SecurityToken::inst()->checkRequest($request)) $this->httpError(400);
```

code/controllers/CMSMain.php
```diff
@@ -1196,16 +1150,23 @@ public function BatchActionList() {
...
 		
 		increase_time_limit_to();
 		increase_memory_limit_to();
-		
+
+    $newPageSet = array();
```
kinglozzer: Tabs for indentation instead of spaces please :)

code/controllers/CMSMain.php
```diff
@@ -1348,7 +1315,10 @@ public function duplicate($request) {
 
 	public function duplicatewithchildren($request) {
 		// Protect against CSRF on destructive action
-		if(!SecurityToken::inst()->checkRequest($request)) return $this->httpError(400);
+		if(!SecurityToken::inst()->checkRequest($request)) {
+      $this->httpError(400);
+      return null;
+    }
```
kinglozzer: Same as above for these `httpErrors()`

Matthew-Bonner:
No problem, I'll change the settings on the IDE and check what is being used then I'll submit a new pull request. Thanks for spotting that some are used, I'll try and be a little more careful in working out what is used and what isn't.

code/controllers/AssetAdmin.php
```diff
@@ -25,28 +16,6 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider{
 	private static $page_length = 15;
 	
 	/**
-	 * @config
-	 * @see Upload->allowedMaxFileSize
-	 * @var int
-	 */
-	private static $allowed_max_file_size;
```
chillu: That removes the config hint which we might start parsing into something useful soon

code/controllers/CMSMain.php
```diff
@@ -1196,16 +1150,23 @@ public function BatchActionList() {
...
+
+    $newPageSet = array();
+
 		if($this->urlParams['ID']) {
 			$newPageSet[] = DataObject::get_by_id("Page", $this->urlParams['ID']);
 		} else {
 			$pages = DataObject::get("Page");
-			foreach($pages as $page) $newPageSet[] = $page;
+			foreach($pages as $page) {
+        $newPageSet[] = $page;
```
chillu: Indentation as well

code/controllers/AssetAdmin.php
```diff
@@ -7,15 +7,6 @@
...
  */
 class AssetAdmin extends LeftAndMain implements PermissionProvider{
-
-	private static $url_segment = 'assets';
-	
-	private static $url_rule = '/$Action/$ID';
-	
-	private static $menu_title = 'Files';
-
-	private static $tree_class = 'Folder';
```
chillu: Are you sure this isn't used?

code/controllers/AssetAdmin.php
```diff
@@ -25,28 +16,6 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider{
...
-		'DeleteItemsForm',
-		'SearchForm',
-		'getsubtree',
-		'movemarked',
-		'removefile',
-		'savefile',
-		'deleteUnusedThumbnails' => 'ADMIN',
-		'doSync',
-		'filter',
-	);
```
chillu: So by removing the whitelist, we allow all public methods? And `deleteUnusedThumbnails` becomes available to everybody with access to the controller?

dhensby:
I'm not sure this kind of work (large scale removal of variables) should go on 3.1

If it's just redundant code removal, rather than declared class variables, then fine, otherwise it's 4.0

PR #1201, opened 2015-05-06
Update default home content

scott1702:


dhensby:
Nice, though I believe the YAML changes need to be done via transifex

tractorcow:
Yeah but the source of the master strings come from PHP, so this is the right place to change it.

dhensby:
h'ok

PR #1202, opened 2015-05-11
FIX: don't use magic accessors to call explicit get functions

stevie-mayhew:
I was fixing a bug on a site where getStatusFlags() is used to determine what information to show in a GridField and these get functions weren't being accessed properly from that code - possibly due to the misrepresentation of the functionality within SiteTree. This is a minor change but improves readability and documentation-ability.

Let me know if this should be further upstream, but as no API is being changed here this should be good for 3.1?

### DIFF
--- a/code/batchactions/CMSBatchActions.php
+++ b/code/batchactions/CMSBatchActions.php
@@ -68,7 +68,6 @@ class CMSBatchAction_Delete extends CMSBatchAction {
 			// if it doesn't remove the tree node
 			$liveRecord = Versioned::get_one_by_stage( 'SiteTree', 'Live', "\"SiteTree\".\"ID\"=$id");
 			if($liveRecord) {
-				$liveRecord->IsDeletedFromStage = true;
 				$status['modified'][$liveRecord->ID] = array(
 					'TreeTitle' => $liveRecord->TreeTitle,
 				);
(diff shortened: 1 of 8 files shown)


tractorcow:
Thanks, but there is also:

CMSBatchActions.php
CMSMain.php
CMSSiteTreeFilter.php
SiteTreeTest.php 

I'm happy for this to go into 3.1 since it improves the quality of the code without actually changing the api in any way.

stevie-mayhew:
Will update all occurrences in the code base :)

stevie-mayhew:
As discussed with @tractorcow , I've also remove removed two calls in `CMSBatchActions` which were setting `IsDeletedFromStage` on the record but were not used in the code.

sminnee:
I don't see a good reason to put code cleanup stuff into a bugfix-only branch, but this seems good for the the 3 branch.

sminnee:
It also broke a behat test but I'm re-running that to see if it was a false positive.

sminnee:
@stevie-mayhew can you rebase this onto the 3 branch?

sminnee:
FYI - I re-ran the behat test, and as expected, it passed this time...

stevie-mayhew:
Hey @sminnee,

That makes sense, I'll open a PR later this morning (after coffee) against 3. I ran the test suite before making the full change,  should have said so to save you the hassle. 

Out of interest, is behat just being finicky or is there a test in there that needs some loving? I've been playing with behat locally so fixing anything which is failing now and then would be of interest to me at the moment.

PR #1203, opened 2015-05-15
API Deprecate delete in favour of archive

tractorcow:
Remove "delete from live" duplicate action in favour of existing "unpublish" which is more consistent with current terminology
Add pop-up verification to destructive actions
Fix bug in reporting publishing of error pages

See https://github.com/silverstripe/silverstripe-framework/pull/4189 for full details


code/model/SiteTree.php
```diff
@@ -2725,9 +2763,9 @@ public function getStatusFlags($cached = true) {
 						'title' => _t('SiteTree.REMOVEDFROMDRAFTHELP', 'Page is published, but has been deleted from draft'),
 					);
 				} else {
-					$flags['deletedonlive'] = array(
-						'text' => _t('SiteTree.DELETEDPAGESHORT', 'Deleted'),
-						'title' => _t('SiteTree.DELETEDPAGEHELP', 'Page is no longer published'),
+					$flags['archived'] = array(
+						'text' => _t('SiteTree.ARCHIVEDPAGESHORT', 'Archived'),
+						'title' => _t('SiteTree.ARCHIVEDPAGEHELP', 'Page is is removed from draft and live'),
```
tractorcow: OOps typo. `is is`

tractorcow: ok fixed this :)

tractorcow:
Added tests for CMS Batch actions.

code/controllers/CMSMain.php
```diff
@@ -88,8 +97,24 @@ public function init() {
 
 		CMSBatchActionHandler::register('publish', 'CMSBatchAction_Publish');
 		CMSBatchActionHandler::register('unpublish', 'CMSBatchAction_Unpublish');
-		CMSBatchActionHandler::register('delete', 'CMSBatchAction_Delete');
-		CMSBatchActionHandler::register('deletefromlive', 'CMSBatchAction_DeleteFromLive');
+		
+
+		// Check legacy actions
+		$legacy = $this->config()->enabled_legacy_actions;
```
micmania1: What's the purpose of this? If devs can call CMSBatchActionHandler::register() from anywhere before this point (_config.php, extensions) then this added API isn't necessary right?

code/model/SiteTree.php
```diff
@@ -2461,6 +2485,22 @@ public function doRevertToLive() {
...
+
+	/**
+	 * Determine if this page references a parent which is archived, and not available in stage
+	 *
+	 * @return bool True if there is an archived parent
+	 */
+	protected function isParentArchived() {
+		if($parentID = $this->ParentID) {
+			$parentPage = Versioned::get_latest_version("SiteTree", $parentID);
+			if(!$parentPage || $parentPage->IsDeletedFromStage) {
```
micmania1: IsDeletedFromStage should be a method.

code/model/SiteTree.php
```diff
@@ -2497,6 +2542,37 @@ public function doRestoreToStage() {
...
+	 */
+	public function canArchive($member = null) {
+		$member = $member ?: Member::currentUser();
+		
+		// Standard mechanism for accepting permission changes from extensions
+		$extended = $this->extendedCan('canArchive', $member);
+		if($extended !==null) return $extended;
+
+		// Check both stages
+		return $this->canDeleteFromLive($member) && $this->canDelete($member);
```
micmania1: Is canDeleteFromLive always needed? If a page is only staged, then a user who has permission to delete from stage but not from live should be able to delete (or archive :p ).

code/controllers/CMSMain.php
tractorcow: See the docs on https://github.com/silverstripe/silverstripe-framework/pull/4189; This is a config to prevent the core api (CMSMain) from automatically calling register on any of the batch handlers. It provides a standardised way to control which deprecated actions should be enabled.

See SiteTree.php for the other place this is used; If we enable the old delete mechanism, then the "archive" popup action will also be replaced with the old "delete" action in order to be consistent.

This is just a temp API until 4.0, where we will remove it, so it's only necessary for the 3.x branch onwards.

code/model/SiteTree.php
tractorcow: It is a method; See SiteTree::getIsDeletedFromStage(). It's used (correctly) as a property accessor all throughout the codebase.

code/model/SiteTree.php
tractorcow: Ok I guess. :D

tractorcow: I've updated this PR; canDeleteFromLive is only checked if the page is published; Otherwise it only checks canDelete.

code/model/SiteTree.php
micmania1: Yeah, but why should it rely on __get() and a whole load of silverstripe magic when you can just call $parentPage->getIsDeletedFromStage() directly.

code/controllers/CMSMain.php
micmania1: There's a config parameter which already exists on CMSBatchActionHandler. 

eg.
$registered = Config::inst()->get('CMSBatchActionHandler', 'batch_actions');
if(!array_key_exist('delete', $registered)) { // enable archive }

Then you can add the deprecation warning to the CMSBatchAction_Delete(FromLive) classes instead 

In 4.0 they can all just be added to the static config and live happily ever after \o/

code/model/SiteTree.php
sminnee: I don't think this blocks merge. We don't have a clear standard one way or the other and I don't think we need to litigate this kind of thing on every change request.

@micmania1 what you might want to do is recommend that for 4.0 we update our coding standard to avoid magic properties where possible. Best for an ss-dev discussion, I think.

code/controllers/CMSMain.php
sminnee: @tractorcow @assertchris does this block merge?

code/model/SiteTree.php
tractorcow: For whatever reason, the entirity of the code base uses the property accessor. I agree with your complaint, but this PR isn't really the place to change that standard. Let's make that a separate PR, and for now just keep this consistent.

stevie-mayhew: See https://github.com/silverstripe/silverstripe-cms/pull/1202 ;)

code/controllers/CMSMain.php
sminnee: The problem with the aforementioned config option is that it doesn't constrain itself only to apply to CMSMain, so this is needed.

code/model/SiteTree.php
sminnee: #1202 looks like a good idea but doesn't block merge of this.

code/controllers/CMSMain.php
tractorcow: Oh right, I remember now.

I tested this, but the problem is the config is global; We'd need to tell OTHER LeftAndMain subclasses which also use batch actions (AssetAdmin) not to add these SiteTree specific batches. I tried to add a "for controller class" config but it ended up as a huge mess and I gave up.

The problem is that the config in place has no allowance for instance specific batch actions. It wouldn't be a problem if the config was on the CONTROLLER, but it's on BatchActionHandler, which is shared for all controllers.

Hell right?

PR #1204, opened 2015-05-19
Saving some memory for echo.

dseguy:


kinglozzer:
Thanks @dseguy, but I’ll close this as it looks like it has been submitted against the wrong branch.

I’m not sure if this level if micro-optimisation is worth it anyway: the `echo` here is only triggered once per request - I don’t think this makes any sort of measurable difference to speed or memory.

nicktacular:
See https://github.com/serbanghita/Mobile-Detect/pull/403

PR #1206, opened 2015-05-21
Fix duplicate not having the correct permission check

ss23:
These checks aren't required -- you can duplicate a page regardless of whether you can edit it.

I secretly presume these are required for something, but if not, here's a nice PR to remove them. If they are, why?

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -1316,7 +1316,9 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 		
 		if(($id = $this->urlParams['ID']) && is_numeric($id)) {
 			$page = DataObject::get_by_id("SiteTree", $id);
-			if($page && (!$page->canEdit() || !$page->canCreate())) return Security::permissionFailure($this);
+			if($page && (!$page->canEdit() || !$page->canCreate(null, array('Parent' => $page->Parent())))) {
+				return Security::permissionFailure($this);
+			}
 			if(!$page || !$page->ID) throw new SS_HTTPResponse_Exception("Bad record ID #$id", 404);
 
 			$newPage = $page->duplicate();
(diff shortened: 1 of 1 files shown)


code/controllers/CMSMain.php
```diff
@@ -1316,7 +1316,6 @@ public function duplicate($request) {
 		
 		if(($id = $this->urlParams['ID']) && is_numeric($id)) {
 			$page = DataObject::get_by_id("SiteTree", $id);
-			if($page && (!$page->canEdit() || !$page->canCreate())) return Security::permissionFailure($this);
```
micmania1: I think this should still include a canView check and a canCreate check. If not then a user could potentially navigate to the /duplicate URL and duplicate pages they don't have permission to see or create.

ss23: Would a canView check be sufficient?

micmania1: Logic being:
A user cannot duplicate a page they can't see.
A user cannot duplicate to a place they can't create.
A user can still edit a page they've been given access to.

tractorcow: canCreate shoud have all the necessary logic to perform this check.

```php
$member = Member::currentUser();
$page = $this;
$parent = $page->Parent();
$context = array('Parent' => $parent);
$canDuplicate = singleton($page->ClassName)->canCreate($member, $context);
```

You could even add `'Duplicate' => $page` to the context if you wanted, although this would be 3.2 land.

ss23:
I have changed this pull request to not break the old permissions system, but just fix the bug where it wasn't working correctly. I believe this can (and should) be merged for 3.1 given it's a bug and not a BC break in itself.

code/controllers/CMSMain.php
```diff
@@ -1316,7 +1316,9 @@ public function duplicate($request) {
...
+			if($page && (!$page->canEdit() || !$page->canCreate($member, array('Parent' => $page->Parent())))) {
```
tractorcow: `$member` variable isn't assigned.

PR #1209, opened 2015-05-29
FIX: SiteTree hints not escaped correctly

kinglozzer:
Looks like this was overlooked in #1177. 3.1’s default cast is `Text`, but not 3.0 so we have to manually escape this. Marked as blocker because it breaks the entire “Pages” section of the CMS.

This has probably been waiting in the wings for a while as SiteTree hints are cached, so no one would spot this unless that cache was wiped (which you can only do by manually removing the files, or visiting `admin/pages/add?flush=1` - it’s not wiped by a regular flush/build in 3.0).

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -422,7 +422,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
 			$this->extend('updateSiteTreeHints', $def);
 
-			$json = Convert::raw2json($def);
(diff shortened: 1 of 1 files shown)


kinglozzer:
Can’t recreate this, I think I’m going mad :disappointed_relieved:. Apologies!

kinglozzer:
https://github.com/silverstripe/silverstripe-cms/compare/3.0.11...3.0.12#diff-36477367c769c0acf8dadd46fefb5be1R22 :disappointed:

PR #1213, opened 2015-06-10
Enforced the published/draft selection for SS reports, which was bein…

nglasl:
…g ignored due to previous code refactoring.

So the first problem is that most reports are currently returning incorrect result sets based on the published/draft selection. This selection is currently being ignored due to previous code refactoring, resulting in `isset($_REQUEST['CheckSite'])` and `isset($_REQUEST['OnLive'])` being false, and therefore always returning the published result set. I believe this problem is in relation to the following, which is through almost every version of 3.0 and 3.1.

https://github.com/silverstripe/silverstripe-cms/commit/57c615d6b232deb00e0275227ae6e5d109b625d0

The second problem is now that this correctly falls back to checking the draft pages, a `DataObject::get` is being called, as opposed to checking the specific stage. If a user has recently viewed a page with `&stage=Live` in the URL, this result set will display all published pages (when clearly it should still be displaying all draft pages).

Something I also thought worth bringing up, is there any reason the individual reports are duplicating the broken links report? They don't seem do anything on top of what's already there. I found this only causes confusion, and therefore I have opted to hide these side reports from the CMS. Would these happen to be for legacy support?

Cheers :)

### DIFF
--- a/code/reports/BrokenLinksReport.php
+++ b/code/reports/BrokenLinksReport.php
@@ -29,9 +29,8 @@ class BrokenLinksReport extends SS_Report {
 				$sort = '';
 			}
 		}
-		if (!isset($_REQUEST['CheckSite']) || $params['CheckSite'] == 'Published') $ret = Versioned::get_by_stage('SiteTree', 'Live', '("SiteTree"."HasBrokenLink" = 1 OR "SiteTree"."HasBrokenFile" = 1)', $sort, $join, $limit);
-		else $ret = DataObject::get('SiteTree', '("SiteTree"."HasBrokenFile" = 1 OR "HasBrokenLink" = 1)', $sort, $join, $limit);
-		
+		$ret = Versioned::get_by_stage('SiteTree', ((!isset($params['CheckSite']) || ($params['CheckSite'] == 'Published')) ? 'Live' : 'Stage'), '("SiteTree"."HasBrokenLink" = 1 OR "SiteTree"."HasBrokenFile" = 1)', $sort, $join, $limit);
+
 		$returnSet = new ArrayList();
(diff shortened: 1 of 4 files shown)


dhensby:
This looks like it fixes a legitimate issue.

We'll need to merge it into https://github.com/silverstripe-labs/silverstripe-reports too

However, there are no tests (it seems) so either we need some, or it needs manually verifying

> The second problem is now that this correctly falls back to checking the draft pages, a DataObject::get is being called, as opposed to checking the specific stage. If a user has recently viewed a page with &stage=Live in the URL, this result set will display all published pages (when clearly it should still be displaying all draft pages).

Sorry, can you confirm if this is still an issue with this patch or if it's resolved?

nglasl:
This patch corrects both the first and second problem, I'm sorry I didn't make that very clear :)

dhensby:
Thanks for confirming.

Are you able to create some tests for this?

nglasl:
There we go, @dhensby. I have now created a test for each different report, and have confirmed the results against the latest version of the CMS (where they all end up failing).

dhensby:
@nglasl you're my hero! I'll take a look when I'm at a computer

nglasl:
Hmm, the build seems to have failed on PHP 5.5 because..

```
$ composer self-update || true
Updating to version d0ff01698de0eb74e45ebe379b6ac8ccd5a6b0db.
Downloading: connection...

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

It might be worth manually triggering another build.

nglasl:
I've manually triggered a build (by closing and opening the PR), and everything now looks to be a success. Over to you, @dhensby :)

PR #1214, opened 2015-06-15
Remove 'Filtered Tree' notification when clearing the filter

scott1702:
Previously, when resetting the filter the 'Filtered Tree.' notification would not disappear
image
It now disappears when clearing the filter.


tractorcow:
What if the filter includes a string component, and CMSSiteTreeFilter_Search is the selected filter? You need to check if CMSSiteTreeFilter_Search is the ONLY filter.

tractorcow:
Tests are failing but not due to this PR... cms behat tests were broken prior to this.

PR #1215, opened 2015-06-16
API Revert DataObject::validate to 3.1 method signature (protected)

tractorcow:
To go with https://github.com/silverstripe/silverstripe-framework/pull/4286


tractorcow:
Merging to follow changes in https://github.com/silverstripe/silverstripe-framework/pull/4286

PR #1216, opened 2015-06-16
Remove the default search text from Searchform

Rhym:
Stop the tyranny of the default value in SearchForm templates. Rise up against our string oppressors!


chillu:
Given that the `TextField` here doesn't have a label, I don't think that's a good default. The extension is straightforward enough to copy across to your controller rather than applying it in order to achieve your level of customisation. Hence I'm going to close this pull request.

dhensby:
can we fix this in 4.0 ;)

PR #1218, opened 2015-06-17
Increase click-size of multi-selection button

scott1702:
Allows the entire multi-selection button to be clicked rather than just the text and the checkbox

### DIFF
--- a/templates/Includes/CMSPagesController_ContentToolbar.ss
+++ b/templates/Includes/CMSPagesController_ContentToolbar.ss
@@ -1,8 +1,8 @@
 <div class="cms-content-batchactions">
-	<div class="view-mode-batchactions-wrapper">
+	<label class="view-mode-batchactions-wrapper">
 		<input id="view-mode-batchactions" name="view-mode-batchactions" type="checkbox" />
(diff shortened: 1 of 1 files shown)


templates/Includes/CMSPagesController_ContentToolbar.ss
```diff
@@ -1,8 +1,8 @@
 <div class="cms-content-batchactions">
-	<div class="view-mode-batchactions-wrapper">
+	<label class="view-mode-batchactions-wrapper">
 		<input id="view-mode-batchactions" name="view-mode-batchactions" type="checkbox" />
-		<label for="view-mode-batchactions"><% _t("CMSPagesController_ContentToolbar_ss.MULTISELECT","Multi-selection") %></label>
-	</div>
+		<span><% _t("CMSPagesController_ContentToolbar_ss.MULTISELECT","Multi-selection") %></span>
```
dhensby: see https://github.com/silverstripe/silverstripe-framework/pull/4293#discussion-diff-32611791

tl;dr Add a class to the span (`view-mode-batchactions-label` perhaps?)

PR #1219, opened 2015-06-17
API make DataObject::validate public

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/4294


dhensby:
Should we update the changelog now or will we use the commit messages to find them?

chillu:
See https://github.com/silverstripe/silverstripe-framework/pull/4286 for context.

PR #1221, opened 2015-06-18
UX improvements

flashbackzoo:
This pull request contains the first part of the UX improvements coming out of the below research:
- http://www.silverstripe.org/blog/using-feedback-to-drive-progress-working-title/
- https://projects.invisionapp.com/share/T53A9E7QA#/screens/83191969

There are UI changes to search filters and bulk actions.

The 'Filters' panel has been removed, and replaced by a search menu, on the top right.

ux-improvements-01

When expanded - you have the same fields as the old filters panel. There is work scheduled to improve the layout of these fields.

ux-improvements-02

The multi-select UI component has been updated and is now called 'bulk actions'.

ux-improvements-03

It is also available in the SiteTree view.

ux-improvements-04

### DIFF
--- a/code/controllers/AssetAdmin.php
+++ b/code/controllers/AssetAdmin.php
@@ -308,7 +308,7 @@ JS
 		$fields->setForm($form);
 		$form->setTemplate($this->getTemplatesWithSuffix('_EditForm'));
 		// TODO Can't merge $FormAttributes in template at the moment
-		$form->addExtraClass('cms-edit-form cms-panel-padded center ' . $this->BaseCSSClasses());
+		$form->addExtraClass('cms-edit-form ' . $this->BaseCSSClasses());
 		$form->setAttribute('data-pjax-fragment', 'CurrentForm');
 		$form->Fields()->findOrMakeTab('Root')->setTemplate('CMSTabSet');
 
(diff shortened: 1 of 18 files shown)


chillu:
BAM! Very cool.

jonom:
:clap:

tractorcow:
Excellent work @flashbackzoo 

These buttons don't seem to be the same height;

image

Also the spacing on the date filter seems a bit inconsistent with the rest of the fields; It would be great if we could make use of the horizontal space a bit more effectively and put the date from - to on the sameline. You could put the from / to into placeholder text.

E.g.

Content [ ... ]
Date [from ] [to ]
Pages [ ... ]
Page Type [ ... ]


As we did with the "lock" pin in the bottom left, it would be great if there was a caption (displayed on hover) beside the search icon.

image


This is a pre-existing issue, but it would be good if the page title didn't say "Pages / Home" when viewing the whole tree. It shouldn't say "Home" if the home page isn't selected.


I understand that in a future version (once the heading bar is updated) we'll be making the search filter available even when editing a page; Just noting that here. :)

Another pre-existing issue but one that would be nice to fix here; Can we change it so that the first option in the bulk actions isn't "Actions". It's fine for it to be displayed on the field when nothing is selected, but it probably shouldn't be on the dropdown list beneath. It would be ok if it was replaced with an empty option, if it can't be completely removed.

image

tractorcow:
You will need to rebase this against the 3 branch; I've just merged up all 3.2 changes into 3 to fix the test issues.

flashbackzoo:
Thanks @tractorcow 

I'll tidy up the styling of the bulk actions button and remove the redundant 'Actions' option. 

RE: The spacing on date filters. @scott1702 has an open pull request on my fork which will fix that.

@clarkepaul what do you think about having an additional label next to the search icon? There is a 'title' you see on hover already.

RE: Breadcrumbs including 'Pages' at the start. That is part of work I have in progress locally, it will be included in a pull request this/next week, along with the 'three column layout' changes. As seen here https://projects.invisionapp.com/share/T53A9E7QA#/screens/83191969

It's not worth doing a stop-gap solution to the search appearing in the SiteTree view. The actual solution is only a week or two away :)

scott1702:
Regarding the additional label, I think it would be unnecessary. The current icon of a magnifying glass is intuitive enough for a user to understand that it means 'Search'.

tractorcow:
Also a comment at https://github.com/silverstripe/silverstripe-framework/pull/4308#issuecomment-113981474 on the icon typography.

clarkepaul:
Re label, there isn't space on the collapsed site tree anyway and that is the view people use most often - I don't think it's needed personally.

clarkepaul:
@tractorcow I'm not sure placeholder is enough info especially if you've already focused also its not  available to ie8 browsers.

tractorcow:
Fair enough; A proper label is probably more robust a solution. :D

scott1702:
Just found a bug, this doesn't allow drag n drop until after the bulk actions are selected then unselected.

clarkepaul:
@scott1702 I think that's built like that on purpose (that you can't drag and drop when multi-select is turned on), I'm not sure if it makes sense to the user though.

flashbackzoo:
Updated the bulk actions styles and placeholder.

flashbackzoo:
buld-actions-button

tractorcow:
Ok that looks fine, I think :D My pixel eyes can't spot anything obviously wrong heh.

flashbackzoo:
Panel styling is now included in the framework PR. Thanks @scott1702 

filter-panel

tractorcow:
I've merged the framework changes; I'll restart the tests on CMS in a few minutes to prove that behat isn't broken.

tractorcow:
Please fix the behat issues and push up again, @flashbackzoo :D

flashbackzoo:
updated behat tests to fix breakages caused by this commit

chillu:
Ah damn, did we just miss the window for pulling in those changes into 3.2, with the first beta released?

flashbackzoo:
yeah it's going to be 3.3 now.

tractorcow:
Behat failures are limited to the pre-existing one.

We have a task to fix that insert-file failure, but it shouldn't block this merge.

PR #1228, opened 2015-06-22
Increase CMSSettingsController:url_priority to suppress SiteConfigLeftAndMain

JayDevlin:
In 3.2, there are two SiteConfigs present in the CMS:
- CMSSettingsController which extends SiteConfigLeftAndMain
- SiteConfigLeftAndMain as a own module

While SiteConfigLeftAndMain's menu item is removed in _config.php, the controller is still active and will be loaded.

Bug:
- In the CMS click the 'Settings' menu item -- its id is "li#Menu-CMSSettingsController"
- xhr is getting a response from SiteConfigLeftAndMain instead
- LeftAndMain.Menu.updateMenuFromResponse() likes to mark the menu item as selected
- ...but fails to do so because its looking for "#Menu-SiteConfigLeftAndMain" instead
- "li#Menu-CMSSettingsController" stays grey

This PR slightly increases the url_priority value of CMSSettingsController, so CMSSettingsController will be used instead of SiteConfigLeftAndMain.

### DIFF
--- a/code/controllers/CMSSettingsController.php
+++ b/code/controllers/CMSSettingsController.php
@@ -7,4 +7,10 @@
  */
 class CMSSettingsController extends SiteConfigLeftAndMain {
 
+	/**
+	 * @config
+	 * @var int
+	 */
+	private static $url_priority = 60;
+
 }


dhensby:
Sorry, I'm a bit confused as to what this is solving. I've just provisioned a 3.2 site and can't see "two" siteconfigs.



Can you elaborate a bit please?

JayDevlin:
This is correct. Because the second menu item is disabled in the menu via CMSMenu::remove_menu_item("SiteConfigLeftAndMain"); in cms/_config.php.

But if you click on the Settings menu item, you'll notice that SiteConfigLeftAndMain is loaded instead of CMSSettingsController. And the menu item will not be marked as 'opened'.

See #1229

I'll have to notice, that is a little bit unusual to increase a property of a soon to be deprecated class. But in my opinion this is the cheapest solution for 3.2 without breaking current API.

Another option would be to change $url_segment to something like 'cmssettings'... I don't know if that'd be considered a API change.

Personally I think, the best solution would be to remove CMSSettingsController, but this is not for me to decide.

dhensby:
Hmmm, interesting. Would a better option be to make use of injector to replace the new class with the old one?

dhensby:
OK. I think I understand the problem. It seems there should be a more elegant solution than this, but after a bit of investigating, I'm not sure what it is!

Let's take this as a patch for now.

THANKS!

PR #1231, opened 2015-06-26
FIX: RedirectorPage_Controller shouldn't attempt redirection if the response is finished (fixes #1230)

### LINKED ISSUE #1230: Unable to redirect in Controller::init() if current page is a RedirectorPage
Example:

```php
class Page_Controller extends ContentController {
	public function init() {
		parent::init();

		if (!$this->someCondition) {
			Security::permissionFailure();
		}
	}
}
```

`RedirectorPage_Controller` will [still attempt its redirect](https://github.com/silverstripe/silverstripe-cms/blob/51c8e8639e9b5f1c258f21f640df2a6cf2af12df/code/model/RedirectorPage.php#L165) even though we’ve already got a redirection, resulting in:

`Already directed to /Security/login?BackURL=%2F; now trying to direct to /my-redirector-page/`

kinglozzer:


kinglozzer:
UGH whitespace changes :disappointed: I can remove those...

PR #1233, opened 2015-06-29
Update button styles

scott1702:
Needed for https://github.com/silverstripe/silverstripe-framework/pull/4353


scott1702:
@chillu This one is also needed for https://github.com/silverstripe/silverstripe-framework/pull/4353 :)

chillu:
Note that the current Behat failure on the 3 branch isn't caused by this commit, but rather https://github.com/silverstripe/silverstripe-cms/pull/1246

PR #1234, opened 2015-06-29
Add three column layout to Page content view

flashbackzoo:
Updated view button styles for new layout

3-col-1

Update breadcrumbs for new layout

3-col-2

Included content filters and view buttons in the SiteTree panel.

3-col-3

SiteTree panel collapsed

3-col-4

### DIFF
--- a/tests/behat/features/insert-a-link.feature
+++ b/tests/behat/features/insert-a-link.feature
@@ -17,7 +17,7 @@ So that I can link to a external website or a page on my site
     Given I select "awesome" in the "Content" HTML field
     And I press the "Insert Link" button
     When I select the "Page on the site" radio button
-    And I fill in the "Page" dropdown with "Home"
+    And I fill in the "internal" dropdown with "Home"
     And I fill in "my desc" for "Link description"
     And I press the "Insert" button
     # TODO Dynamic DB identifiers
@@ -29,7 +29,7 @@ So that I can link to a external website or a page on my site
     Given I select "awesome" in the "Content" HTML field
     And I press the "Insert Link" button
     And I select the "Page on the site" radio button
-    And I fill in the "Page" dropdown with "Details"
+    And I fill in the "internal" dropdown with "Details"
     And I wait for 1 second
     And I select "youranchor" from "Form_EditorToolbarLinkForm_AnchorSelector"
     And I press the "Insert link" button
@@ -52,9 +52,9 @@ So that I can link to a external website or a page on my site
     Given I select "awesome" in the "Content" HTML field
     When I press the "Insert Link" button
     When I select the "Download a file" radio button
-    And I fill in the "File" dropdown with "file1.jpg"
+    And I attach the file "testfile.jpg" to "file[Uploads][]" with HTML5
     And I press the "Insert link" button
-    Then the "Content" HTML field should contain "<a href="[file_link,id=1]" target="_blank">awesome</a>"
+    Then the "Content" HTML field should contain "<a href="[file_link,id=3]" target="_blank">awesome</a>"
     # Required to avoid "unsaved changes" browser dialog
     Then I press the "Save draft" button
 
--- a/tests/behat/features/manage-files.feature
+++ b/tests/behat/features/manage-files.feature
@@ -36,7 +36,7 @@ Feature: Manage files
     And I click on "file1" in the "folder1" table
     And I fill in "renamedfile" for "Title"
     And I press the "Save" button
-    And I press the "Back" button
+    And I follow "folder1"
     Then the "folder1" table should not contain "testfile"
     And the "folder1" table should contain "renamedfile"
 
--- a/css/screen.css
+++ b/css/screen.css
@@ -17,12 +17,13 @@
 .cms-content-tools #cms-content-treeview .cms-content-toolbar { border-bottom: none; box-shadow: none; margin-bottom: 8px; }
 .cms-content-tools #cms-content-treeview .cms-tree-expand-trigger { display: block; float: left; margin: 0 0 2px 0; }
 .cms-content-tools #cms-content-treeview .cms-tree-expand-trigger span.ui-button-text { padding-right: 8px; }
+.cms-content-tools #cms-content-treeview .cms-tree .badge { display: none; }
 .cms-content-tools #cms-content-treeview .cms-tree a:hover > .text > .badge, .cms-content-tools #cms-content-treeview .cms-tree .jstree-clicked > .text > .badge { display: inline-block; }
 
 /** -------------------------------------------- The CMS Content Toolbar. Associated behaviour in CMSMain.js -------------------------------------------- */
 .cms-content-toolbar button.ss-ui-button { padding: .4em .75em; vertical-align: baseline; }
 .cms-content-toolbar button.ss-ui-button:hover { -webkit-box-shadow: none; -moz-box-shadow: none; box-shadow: none; }
-.cms-content-toolbar button.ss-ui-button:focus { -webkit-box-shadow: none; -moz-box-shadow: none; box-shadow: none; outline: 0; }
+.cms-content-toolbar button.ss-ui-button:focus { -webkit-box-shadow: none; -moz-box-shadow: none; box-shadow: none; }
 .cms-content-toolbar button.ss-ui-button:before { font-size: 1.1em; }
 .cms-content-toolbar button.ss-ui-button.ui-state-focus { -webkit-box-shadow: none; -moz-box-shadow: none; box-shadow: none; }
 .cms-content-toolbar button.ss-ui-button.active { -webkit-box-shadow: 0 0 5px #b3b3b3 inset; -moz-box-shadow: 0 0 5px #b3b3b3 inset; box-shadow: 0 0 5px #b3b3b3 inset; }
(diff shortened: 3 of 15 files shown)


tractorcow:
I would really like to merge this but I think it'd be wise to leave it open for discussion for users to give feedback.

flashbackzoo:
@jonom @clarkepaul

tests/behat/features/insert-a-link.feature
```diff
@@ -52,9 +52,9 @@ So that I can link to a external website or a page on my site
     Given I select "awesome" in the "Content" HTML field
     When I press the "Insert Link" button
     When I select the "Download a file" radio button
-    And I fill in the "File" dropdown with "file1.jpg"
+    And I attach the file "testfile.jpg" to "file[Uploads][]" with HTML5
     And I press the "Insert link" button
-    Then the "Content" HTML field should contain "<a href="[file_link,id=1]" target="_blank">awesome</a>"
+    Then the "Content" HTML field should contain "<a href="[file_link,id=3]" target="_blank">awesome</a>"
```
tractorcow: It would be nice if there was a way of wildcarding that file ID, but for the sake of fixing this PR it's fine.

jonom:
This is broken now? Shouldn't be visible unless active or hovered over in this view, that's the whole point of the dots.

pasted_image_29_06_2015_4_19_pm

jonom:
But otherwise I think it's rad :) :+1:

flashbackzoo:
Yeah that's broken - I'll fix it.

SpiritLevel:
OK! Some feedback from a noob! Looks great as it is, but if you want some really nit-picky ideas, here goes...feel free to completely disregard if not helpful..

(1) the plus sign on the Add New button is pixelated. Same with check marks on Saved, Published.
(2) the "O" on the Home page icon seems misleading. Perhaps an "M" for modified ?
(3) it may seem "O"bsessive of me, but the "O" at the bottom left of the screen doesn't suggest what it does. Perhaps a more representative icon ?
(4) just wondering about "Page name", "URL segment", and "Navigation label" as section descriptions; and the URL given is the full one not just the segment... 
(5) Edit button: "Edit" is not vertically centered in the button. Also, button is too close to end of the URL.
(6) Place "Your Site Name" at top of main window or perhaps in breadcrumbs: "Your Site Name/Home/New Page" ?
(7) security icon? Perhaps a lock instead of three people ?
(8) bar graph instead of a pie chart icon for reports ?
(9) file icon looks like it is only for images...perhaps an icon that represents a more generic file type ?

What an awesome CMS!

flashbackzoo:
Fixed badges as pointed out by @jonom 

3-col-fix1

Hover:

3-col-fix2

jonom:
Thanks, that's better :) just FYI @flashbackzoo, you probably saw this and it shouldn't affect your work here but 3.2 now uses a hollow dot (ring) for new pages and a solid dot for modified, and has a much lighter blue for regular hover and active states for consistency. I assume that will be merged down to 3 at some point. https://github.com/silverstripe/silverstripe-framework/pull/4335 

Feels weird that I'm referencing those changes in the Framework project... shouldn't most of the stuff in the `framework/admin` folder be in the CMS project? I'm sure there's a good reason for it, but it does make it a bit hard to know where to go when you want to change stuff in the CMS interface.

flashbackzoo:
Hey @SpiritLevel thanks for your feedback.

This pull request is just dealing with splitting the layout into three columns and the styles related to that change. Your points valid - but are outside the scope of this pull request.

There's another pull request that's dealing with the button styles you mentioned https://github.com/silverstripe/silverstripe-framework/pull/4353

@clarkepaul is currently updating some of the icons and he'd be keen for input. Head over to http://silverstripe.uservoice.com/forums/251266-new-features/suggestions/8107791-use-vector-icons-for-all-icons-in-the-cms and leave your feedback :)

SpiritLevel:
Sorry about that @flashbackzoo ... Just got on a roll ;)

flashbackzoo:
Cheers @jonom yeah I saw that.

The reason for `framework/admin` is allowing `controllers` etc to have a permissions model i.e. only members access `MyAPIController`. If you're providing a permissions model, you need a way to administer members, so there's ModelAdmin.

It's possible this could be split out into a separate module that's required by CMS and Framework in 4.0, but I'm not sure if anyone's given it too much thought.

It is a pain having to make two pull requests for these type of changes, a re-factor of the CSS is definitely going to happen for 4.0, so we'll see what can be moved across to CMS then.

flashbackzoo:
@SpiritLevel no worries - always appreciate another set of eyes :)

flashbackzoo:
Well - at least the Behat tests are passing on 3 now! I'll see what's going on with the unit tests...

flashbackzoo:
Updated CMSMainTest.php to fix the failing build.

The top level (Pages) breadcrumb now appears in the SiteTree panel - that was causing the issue.

3-col-testfail

chillu:
This makes the CMS so much more polished, wow! Some niggles:

 * New breadcrumbs break on long page titles: http://take.ms/CSm0V. Please also test that this still works with multiple rows of tabs. Yes, we should avoid that, but tab rows are a reality we have to deal with at the moment :) And check that long "parent" breadcrumb link titles also work.
 * Tree icon isn't enabled when in page edit mode: http://take.ms/piKTB. I can see why (it's switching to full-page tree mode on click), but still confused me. Any ideas?
 * Keyboard focus is hidden when tabbing through to the new UI elements (e.g. tree and list icons). We'll want to keep the CMS at least as accessible as it is at the moment, right?
 * Tree and list icons don't have alternative titles. Same reason as above.

flashbackzoo:
Nice spotting. I'll fix those issues tomorrow.

chillu:
* When filtering the tree in page edit mode, clicking the tree or list icons doesn't persist the filter.
 * Hiding the search form after submitting makes it much less obvious than in the old implementation that a search filter is currently applied. We have the "Search results. clear" label on there, but its very easy to miss. I'd suggest that we leave the panel expanded, and close it when hitting "clear filter". Also, if you manually close it with filters still applied by clicking on the loupe icon, that icon should be highlighted somehow (blue glow?).
 * The green search button right above a green "add page" button is a bit confusing. Maybe the green is too strong in this context? I suggest we make it a grey background and remove the button styling on "clear filter" in favour of a more subdued link styling (like "more options" in the edit page screen)

flashbackzoo:
Thanks - I'll fix the button mode persist issue.

@clarkepaul - Thoughts on the search filters staying expanded?

I'll wait and see what happens with https://github.com/silverstripe/silverstripe-framework/pull/4353 before trying to solve the button issue. But agree - it could be improved.

tractorcow:
Passing behat tests on cms, in my travis?

flashbackzoo:
@chillu I've fixed some of those issues. Also updated https://github.com/silverstripe/silverstripe-framework/pull/4355

Long breadcrumbs:

3-col-breadcrumbs

Button active and focus states:

3-col-focus

I couldn't reproduce the mode persisting issue. Only tree view mode is available in the content editing view, so clicking the list view button will take you back to the top level. That state (for me) persists on subsequent navigations to page content then back to the top level.

tests/behat/features/manage-files.feature
```diff
@@ -63,6 +63,7 @@ Feature: Manage files
 
   Scenario: I can filter the files list view using name
     Given I expand the content filters
+    And I wait for 2 seconds
```
chillu: I see why that's required (animation). But it makes the test less readable, and makes the test even slower to run. Should be solved by having the previous step ("expand the content filters") waiting for animation to finish. Roughly `$this->waitFor('jQuery(':animated').length == 0')`. Do you want to have a crack at it?

flashbackzoo: Yeah good call - I'll update it.

clarkepaul:
@chillu if the search stayed expanded we would need to make that a consistent pattern even when there are 20 searchable fields, meaning you would need to scroll to see the results, the user wouldn't even really see any change to the interface. If the search notification isn't obvious enough then I think we need to deal with that rather than showing the search fields. 

Here's an option to consider instead (done pretty fast and crudely so just use as wireframe).

clarkepaul:
silverstripe ui_search

tractorcow:
I've restarted the broken builds; Composer timeouts it seems.

chillu:
@clarkepaul "... the user wouldn't even really see any change to the interface". Only if we don't make the search panel individually scrollable, and limit its height. Has the disadvantage of pushing the search button out of view, so only feasible if we can fix the search button to the bottom of that panel - which in turn is difficult to achieve with the current layout engine customisations. Even with the new wireframe, its far less obvious what is searched for (or how to find that out) than in 3.1. "Search results for ..." with nice labels would be ideal. Or a "details" link which expands the search panel?

@flashbackzoo Thanks, fixed looking good! The filter not persisting happens in this view when I press the list button: 
image

clarkepaul:
@chillu yup I agree with your statement around what has been searched for (myself and @flashbackzoo already discussed this previously and thought it would be a nice addition) but as always the end of the sprint is looming. If we added the link "Details" for now would that be a suitable improvement (nice idea). I don't really like the idea of a fixed height search panel still.

clarkepaul:
silverstripe ui_search

chillu:
Sweet, yep both of these approaches work for me. Thumbs up for merging as-is for a first step (once the Behat "wait 2 seconds" thing is fixed). Do you think it is feasible to add to your sprint backlog? How do we ensure it doesn't get forgotten about?

flashbackzoo:
@chillu fix. see https://github.com/silverstripe/silverstripe-framework/pull/4355 for the context update.

jonom:
I think option 2 with the coloured background for the whole panel is a super nice way of indicating a filtered tree

flashbackzoo:
ahh yeah - the Behat build is going to fail until the framework PR has been merged...

flashbackzoo:
@chillu We've created a story in our backlog to improve the search filters notification :smile:

tractorcow:
The tests just went green before it was squashed; So I merged.

PR #1245, opened 2015-07-14
Update CONTRIBUTING.md

torleif:
[Fix] dead links for 3.1

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ See our [high level overview](http://silverstripe.org/contributing-to-silverstri
 
 Or, for more detailed guidance, read one of the following pages:
 
- * [Sharing your opinion and raising issues](http://doc.silverstripe.org/framework/en/trunk/misc/contributing/issues)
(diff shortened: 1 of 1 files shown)


chillu:
@cam-findlay @wilr Do you know why those links have been broken? Do we need a redirect from trunk/master to 4.0? :(

camfindlay:
@ingo there should be a rewrite for trunk however I'm finding the rewrites from a while back didn't include a /?$ at the end of the incoming link so in some cases non-trailing slash links haven't been matching. There is some work going on over in docs repo on the rewrites at the moment by a community member with some mentoring from me so hopefully we can clean many of these up.

jonom:
Looks like /contributing used to be nested under /misc but isn't anymore?

Might be a silly question but shouldn't the contents of this file be in the README.md file? Much better chance of people reading it there I would have thought.

camfindlay:
@jonom the CONTRIBUTING file is a GitHub standard.

jonom:
Ohhh. Cool. :+1:

PR #1246, opened 2015-07-18
Behat tests hotlinking to non-existant image

chillu:
Caused by silverstripe.com redesign. Ideally we'd have a permanent asset hosted there which is clearly marked as a test dependency. @tractorcow If you can be bothered, I'd suggest redeploying ss.org with such an image (e.g. `behat-test-image-do-not-delete.png`).

Or create a hostname placeholder for Behat feature files, so images can be served localhost within the Behat tests, in which case we wouldn't require a network connection to execute Behat.

### DIFF
--- a/tests/behat/features/insert-an-image.feature
+++ b/tests/behat/features/insert-an-image.feature
@@ -17,12 +17,12 @@ Feature: Insert an image into a page
     Then I should see "Choose files to upload..."
 
     When I press the "From the web" button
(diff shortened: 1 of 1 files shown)


tractorcow:
All of travis is sad at the moment, I'm afraid.

We need to merge up travis changes from https://github.com/silverstripe/silverstripe-framework/commit/2036f7e6c21a024d51ad0e0b2ae19aba25ea1e61 into other modules.

PR #1248, opened 2015-07-23
FIX: ContentController::getViewer() not returning all found templates (fixes #1244)

### LINKED ISSUE #1244: ContentConroller::getViewer() fails to correctly resolve templates for actions
This code fails to execute the correct template in 3.2:

```php

class Page_Controller extends ContentController {

    private static $allowed_actions = array(
       'success',
    );

}
```

This executes `$this->renderWith(array('Page_success', 'Controller'));` instead of `$this->renderWith(array('Page_success', 'Page'));`

kinglozzer:
`$arr + $arr` will overwrite values with matching numeric keys whereas `array_merge()` will append them.

The new assertion is a little messy, but I didn’t want to modify the test templates because then I’d have to modify all the other tests!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,13 @@ env:
 
 matrix:
   allow_failures:
-    - php: hhvm-nightly
+    - php: hhvm
+    - php: 7.0
+    - php: nightly
+
   include:
+    - php: 5.4
+      env: DB=MYSQL PDO=1
     - php: 5.5
       env: DB=MYSQL
     - php: 5.6
@@ -37,7 +42,11 @@ matrix:
       env: DB=MYSQL BEHAT_TEST=1
     - php: 5.3
       env: DB=MYSQL
-    - php: hhvm-nightly
+    - php: 7.0
+      env: DB=MYSQL
+    - php: nightly
+      env: DB=MYSQL
+    - php: hhvm
       env: DB=MYSQL
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,11 @@
-language: php 
+language: php
+
+sudo: false
+
+addons:
+  apt:
+    packages:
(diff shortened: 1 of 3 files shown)


kinglozzer:
I have no idea what’s going on with Behat. The scenarios that are failing on Travis all pass locally, and I can’t see how my change could affect these tests...

The PHP 5.3 failure is for using `$this` in a closure. I’ve fixed that and force pushed, but Travis won’t acknowledge that fact.

confuse

kinglozzer:
Right, removing the composer cache stuff from `.travis.yml` fixed the issue of it not picking up my latest commit... is it caching stuff it shouldn’t be? Still no idea what’s going on with Behat

dhensby:
@tractorcow interesting that the cache appears to cause problems... You think it's related?

SpiritLevel:
@kinglozzer Love the clip of Tom Hanks :) It perfectly captures technology induced psychosis! Is that your favorite movie, by the way :P

tractorcow:
Yikes, I wonder if the composer cache is actually a breaking change to the way travis-support works. Maybe we should remove it.

tractorcow:
Bug logged at https://github.com/silverstripe-labs/silverstripe-travis-support/issues/18. I suggest we disable the cache altogether for now.

tractorcow:
@dhensby I've also noticed this issue in my own modules where I've introduced the cache; It's definitely confirmed.

kinglozzer:
Phew, the behat failures aren’t my fault: 3.2 is [already failing](https://travis-ci.org/silverstripe/silverstripe-cms/jobs/71780241).

dhensby:
@tractorcow I always knew the cache would be trouble. Glad I didn't add it to all our repos now

dhensby:
I'll test this locally and merge later

Issue #1237, opened 2015-07-01
Issues with new "archived" state


sweddell:
Giving the new 3.2 beta a test drive and have found issues with how the "archived" state is treated. A common usecase for us is the need to restore an entire section when a user archives a page, either knowingly or unsuspectingly, and there were subpages in the tree that needed to be moved before archiving, or the whole lot needed to remain in the case of an unscheduled archive. 
Archiving a page with children,
![screenshot-2015-07-01 13-25-14](https://cloud.githubusercontent.com/assets/75909/8447319/f28dd018-1ff9-11e5-837a-c77edfcf939e.png)
Results in the children pages being archived, as expected.
![screenshot-2015-07-01 13-25-55](https://cloud.githubusercontent.com/assets/75909/8447328/236be468-1ffa-11e5-8e3e-605a30a7b2d7.png)
There needs to be a "restore" action in the Multi-selection dropdown, especially if a very large section was archived and needs to be returned to its former glory.
![screenshot-2015-07-01 13-28-38](https://cloud.githubusercontent.com/assets/75909/8447431/0af3938e-1ffc-11e5-8fe0-0373805c2dd4.png)
Restoring the subpages individually only gives an option to restore to the top level, which is expected if the parent is archived, but not if the page is moved to an active page first.
![screenshot-2015-07-01 14-42-54](https://cloud.githubusercontent.com/assets/75909/8447675/8d761856-1fff-11e5-8224-e13cb58bd196.png)
However, if the parent page that was archived is restored, then the expected "restore draft" option is available, as a child of the same restored parent or another parent in the tree if it is moved.

![screenshot-2015-07-01 14-48-58](https://cloud.githubusercontent.com/assets/75909/8447725/6ce5c676-2000-11e5-91cb-fcd65786a0ea.png)
![screenshot-2015-07-01 14-51-11](https://cloud.githubusercontent.com/assets/75909/8447731/b66fab7c-2000-11e5-8741-c9ec17873f89.png)
On restoring, or the tree refreshing, the page returns to the previous parent. It is probably a good idea to restrict the ability to move the page unless restored first. I understand what is going on here and the message about the move not taking effect until the page is saved relates to the reason the page moves back, but a CMS user won't understand that and will think that it should stay in the place they moved it to the same as if the moved a published or draft page.

jonom:
[...]

PR #1255, opened 2015-08-03
API batch restore action

tractorcow:
**MUST** be merged with https://github.com/silverstripe/silverstripe-framework/pull/4471

Partially resolves https://github.com/silverstripe/silverstripe-cms/issues/1237

This gives a prompt prior to restore, describing that if you are restoring child pages (without parents) they'll be restored to the root.

### DIFF
--- a/code/batchactions/CMSBatchActions.php
+++ b/code/batchactions/CMSBatchActions.php
@@ -68,6 +68,48 @@ class CMSBatchAction_Archive extends CMSBatchAction {
 }
 
 /**
+ * Batch restore of pages
+ */
+class CMSBatchAction_Restore extends CMSBatchAction {
+	
+	public function getActionTitle() {
+		return _t('CMSBatchActions.RESTORE', 'Restore');
+	}
+
+	public function run(SS_List $pages) {
+		// Sort pages by depth
+		$pageArray = $pages->toArray();
+		// because of https://bugs.php.net/bug.php?id=50688
+		@usort($pageArray, function($a, $b) {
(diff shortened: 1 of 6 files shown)


tractorcow:
image

sminnee:
The code looks good to me; we'll see how the test-run goes.

The message is a little confusing but improvements can be left for post-3.2. In particular, it would be a lot better if the batch action detected which pages are going to be moved to the root level and listed their names: users shouldn't have to execute the logic in their head. ;-)

I'm not sure if this requires a new API for providing a dynamic error message.

tractorcow:
I thought that, so I've updated the message.

image

tractorcow:
I've added https://github.com/silverstripe/silverstripe-cms/pull/1256

both travis tests can race, because I'm sick of waiting for them to fail to update with a fix.

code/batchactions/CMSBatchActions.php
```diff
@@ -68,6 +68,48 @@ public function applicablePages($ids) {
...
+	
+	public function getActionTitle() {
+		return _t('CMSBatchActions.RESTORE', 'Restore');
+	}
+
+	public function run(SS_List $pages) {
+		// Sort pages by depth
+		$pageArray = $pages->toArray();
+		// because of https://bugs.php.net/bug.php?id=50688
+		@usort($pageArray, function($a, $b) {
```
tractorcow: This solution uses error suppression (ugh)

sminnee:
It looks like 1256 has worked, and since that doesn't involve error suppression, it's my preferred approach.

PR #1256, opened 2015-08-03
API batch restore action (ALTERNATE SOLUTION)

tractorcow:
Alternative solution to https://github.com/silverstripe/silverstripe-cms/pull/1255

### DIFF
--- a/code/batchactions/CMSBatchActions.php
+++ b/code/batchactions/CMSBatchActions.php
@@ -68,6 +68,51 @@ class CMSBatchAction_Archive extends CMSBatchAction {
 }
 
 /**
+ * Batch restore of pages
+ */
+class CMSBatchAction_Restore extends CMSBatchAction {
+	
+	public function getActionTitle() {
+		return _t('CMSBatchActions.RESTORE', 'Restore');
+	}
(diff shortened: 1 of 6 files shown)


code/batchactions/CMSBatchActions.php
```diff
@@ -68,6 +68,51 @@ public function applicablePages($ids) {
...
+	public function getActionTitle() {
+		return _t('CMSBatchActions.RESTORE', 'Restore');
+	}
+
+	public function run(SS_List $pages) {
+		// Sort pages by depth
+		$pageArray = $pages->toArray();
+		// because of https://bugs.php.net/bug.php?id=50688
+		foreach($pageArray as $page) {
+			$page->getPageLevel();
```
tractorcow: This solution attempts to pre-load the page level calculation so that usort doesn't hit the `PHP Warning:  usort(): Array was modified by the user comparison function` error

sminnee:
It looks like 1256 has worked, and since that doesn't involve error suppression, it's my preferred approach.

tractorcow:
I agree. Best solution is the best.

PR #1259, opened 2015-08-14
UPDATE: make all response code use getResponse() method

stevie-mayhew:
Needs https://github.com/silverstripe/silverstripe-framework/pull/4500

### DIFF
--- a/code/controllers/ModelAsController.php
+++ b/code/controllers/ModelAsController.php
@@ -8,7 +8,7 @@
  */
 class ModelAsController extends Controller implements NestedController {
 	private static $extensions = array('OldPageRedirector');
-	
+
 	/**
 	 * Get the appropriate {@link ContentController} for handling a {@link SiteTree} object, link it to the object and
 	 * return it.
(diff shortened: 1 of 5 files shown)


code/controllers/ModelAsController.php
```diff
@@ -49,82 +49,79 @@ public function init() {
 	public function handleRequest(SS_HTTPRequest $request, DataModel $model) {
 		$this->setRequest($request);
 		$this->setDataModel($model);
-		
-		$this->pushCurrent();
 
-		// Create a response just in case init() decides to redirect
-		$this->response = new SS_HTTPResponse();
```
kinglozzer: It looks like the response was intentionally created *before* `init()` is called here. Have you checked there are no regressions if `init()` is relying on `$this->response`?

stevie-mayhew: If init is relying on `$this->response` then it would have to be calling `$this->getResponse()` which is an API change I didn't account for :\

Should I add in a local `$response` variable to this code or submit this further up the chain @kinglozzer ?

kinglozzer: Sorry for the delay, I think storing a local `$response = $this->getResponse()` variable before `init()` is called should cover this, right? Because then `$this->response` will be set

dhensby: This is a regression in the framework PR too...

Storing $this->getResponse() as a local var doesn't solve any problem directly. it's the calling of `getResponse()` that will fix the regression

stevie-mayhew: I've added the call without a return parameter to initialise the response on the request and have added a fix for the regression of the framework at https://github.com/silverstripe/silverstripe-framework/pull/4629

dhensby:
thanks

PR #1263, opened 2015-08-20
MINOR Change how OldPageRedirector gets URLs to fix nested redirects

ss23:
Previously, we would be limited by the way allParams will return 3 parameters at most. This way, we get the full URL.
Keep in mind, this code still needs a clean up, but at least it's not buggy now.

This should probably be merged up as required.

### DIFF
--- a/code/controllers/OldPageRedirector.php
+++ b/code/controllers/OldPageRedirector.php
@@ -10,8 +10,8 @@ class OldPageRedirector extends Extension {
 	 * @throws SS_HTTPResponse_Exception
 	 */
 	public function onBeforeHTTPError404($request) {
-		// Build up the request parameters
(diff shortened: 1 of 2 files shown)


sminnee:
Test looks like it covers it, and it's a bug so good for 3.1.

ss23:
I tried bouncing travis but it's not working. The old travis run had a failure on behat that looked unrelated (couldn't select ErrorPage). Is it overloaded or something, or can we just merge or what? :O

sminnee:
Bounced the behat build, it works.

sminnee:
But no, I wouldn't merge until I'd seen a green tick. Either bouncing the build will fix it or it's something that would risk introducing a failure to the main branch, which would suck.

PR #1274, opened 2015-09-03
Ensure we have a ParentID before trying to limit children by page type (fixes #1273)

### LINKED ISSUE #1273: CMS users can view/select pages they don’t have permission to create
Spotted this on 3.1.14-rc1.

If a user doesn’t have permission to create a page, it appears in the list of available page types (#1272) but is disabled. If you then choose “Under another page” then all of the page types are enabled and can be selected. This doesn’t seem to happen if a parent page is pre-selected by clicking “Add page” when already editing a page.

The user cannot add the page, the validation error occurs as expected.

Steps to reproduce:
- Create a new page type with `canCreate() { return false; }`
- Visit `admin/pages/add/`, note that your page type is there but disabled
- Click “Under another page”, the page type is now available to be selected

Screenshots:

screen shot 2015-09-03 at 14 29 51
screen shot 2015-09-03 at 14 30 06

kinglozzer:
Pretty simple fix, it was trying to get SiteTree hints for “SiteTree” instead of “Root” when no parent page is selected (so it _should_ be trying to get the hints for “Root”).

### DIFF
--- a/javascript/CMSMain.AddForm.js
+++ b/javascript/CMSMain.AddForm.js
@@ -52,7 +52,7 @@
 						? (parentTree.getValue() || this.getParentID())
 						: null,
 					newClassName = metadata ? metadata.ClassName : null,
-					hintKey = (newClassName && parentMode === 'child')
+					hintKey = (newClassName && parentMode === 'child' && id)
 						? newClassName
 						: 'Root',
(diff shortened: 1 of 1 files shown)


dhensby:
can you rebase rather than merge, please?

PR #1275, opened 2015-09-07
Check result before foreach

mikenz:
MINOR Don't assume all classes in 'allowed_children' actually exist.

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -2714,7 +2714,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 					$allowedChildren[] = substr($candidate,1);
 				} else {
 					$subclasses = ClassInfo::subclassesFor($candidate);
-					foreach($subclasses as $subclass) {
+					if ($subclasses) foreach($subclasses as $subclass) {
 						if($subclass != "SiteTree_root") $allowedChildren[] = $subclass;
 					}
 				}


tractorcow:
Except that subclassesFor does always have a result. :) The initial item in the list is the class that you pass in.

mikenz:
Correct, and if I pass in a class that doesn't exist (I'm migrating lots of templates from post-2.4) everything breaks. So this eases my migration so I don't have to comment out children then re-enable them once I've migrated them.

kinglozzer:
@mikenz `ClassInfo::subclassesFor()` returns the class itself if it doesn’t exist:

```php
var_dump(ClassInfo::subclassesFor('FooBarBaz'));
// array(1) { ["FooBarBaz"]=> string(9) "FooBarBaz" }
```

So I don’t see how this will change anything - it will always pass that `if()` statement

mikenz:
@kinglozzer here's the first line of `ClassInfo::subclassesFor()` 
```php
public static function subclassesFor($class) {
		if(is_string($class) && !class_exists($class)) return null;
```

It returns null if the class doesn't exist. This was introduced recently: https://github.com/silverstripe/silverstripe-framework/@/1f0602d42fd9e1c0a4268f3a51aa7f483100a935

kinglozzer:
Ah, I was looking at 3.1: this was fixed in 3.1 in https://github.com/silverstripe/silverstripe-framework/pull/4482.

@sminnee https://github.com/silverstripe/silverstripe-framework/pull/4497 introduced `return null` in a few methods that are marked as `@return array` or `@return string`. Do you think it would be safer to revert that? Otherwise, those return statements will need to be removed for the fix in https://github.com/silverstripe/silverstripe-framework/pull/4482 to work when we merge 3.1 up.

tractorcow:
Lol, I was doing the same thing. :) My bad.

sminnee:
Of course it's not safe to revert a bugfix—the bug would be reintroduced!

However, it's probably better to return an `array()` rather than `null` for the array-returning methods.

I'm a little more reticent about returning and empty string for `ClassInfo::baseDataClass()`. It seems to give the wrong impression. Perhaps we change the type hint to `string|null` for that one?

tractorcow:
> However, it's probably better to return an array() rather than null for the array-returning methods.
 
 What I would do. ;)

kinglozzer:
> Of course it's not safe to revert a bugfix—the bug would be reintroduced!

Temporarily re-introduced until the fix in 3.1 is merged up, but I take your point - probably easier to just tidy it up in a PR instead of reverting and merging up :)

My thinking was that it might be easier to approach it like that than try to re-instate the old behaviour for each of the methods individually (returning an empty array doesn’t guarantee BC in all cases, like the `ClassInfo::subclassesFor()` example above).

dhensby:
Is this fixed now or does it need merging?

tractorcow:
It was fixed by returning an array in some other PR.

PR #1283, opened 2015-09-28
fixing phpdoc blocks to refering to Objects to built-in types of PHP

spekulatius:


dhensby:
Looks good to me @assertchris, you're my fallback for phpdoc expertise ;) - thoughts?

Issue #736, opened 2013-05-13
[2010-09-23] ModelAdmin Delete Button invokes Form Validation


silverstripe-issues:
created by: dswnz
created at: 2010-09-23
original ticket: http://open.silverstripe.org/ticket/6032

---

When viewing a record via the ModelAdmin (ajax url: http://my.silverstripe.site/admin/mymodeladmin/MyDataObject/6/edit) and deciding to delete it (ajax url: http://my.silverstripe.site/admin/mymodeladmin/MyDataObject/6/EditForm?action_doDelete=Delete) ModelAdmin decides to validate the data before deleting it.

The "MyDataObject" has been set up with a '''getCMSValidator()''' function which returns '''new RequiredFields('Name')'''
and the Form contains an '''EmailField()'''

Screenshot: Result of clicking the Delete Button.
[[Image(http://grab.by/grabs/12b68fdf791be27625fc846a26638ed6.png)]]

patricknelson:
For anyone interested, please see #586 for more information on a potential fix (in case this is directly related or a duplicate) and PR #1287 for a solution that address it directly in the `3.1` branch of `silverstripe-cms`.

wilr:
Fixed

Issue #586, opened 2013-04-03
[2012-11-08] Cannot delete page if required CMS field is not filled in


silverstripe-issues:
created by: Friksel
created at: 2012-11-08
original ticket: http://open.silverstripe.org/ticket/8009

---

Cannot delete a page if a required CMS field (with getCMSValidator) is not filled in. It's impossible to delete the page, because the CMS field that's not filled in has changed to read only....

Error message that appears on the topright after clicking 'delete' says 'Validation error'.

Restoring the page doesn't work either and throws the same error message.

Silverstripe 3.0.3-rc1

robby-nzta:
Is there any progress with this?

kinglozzer:
@robby-nzta I think https://github.com/silverstripe/silverstripe-cms/pull/863 was supposed to resolve this (amongst other issues) but it looks like I may have missed the `unpublish` action. Is this issue occurring when attempting to unpublish?

robby-nzta:
Thanks, Yes, this is working on Page type. 
After reading through my case, the validation is used  RequiredFields. 
So finally, I should write custom validation for my case. 

I have two action doUnpublish and doDelete should bypass validation in my requirement.

Cheers

patricknelson:
If it helps anyone else:

It appears that this is resolved for me by taking into account @kinglozzer's fix above for #863 with one tiny little change: Adding `delete` action. 

**My particular situation** (which may vary from others): This occurs when, for example, I create a new page type that does have validation of any sort, but immediately decide that I want to delete that accidental creation. Adding the `delete` action in there solves this for me. Thankfully this is safely (and well) exposed via the `->setValidationExemptActions()` and `->getValidationExemptActions()` duo on `CMSForm`.

patricknelson:
By the way, I should share my actual solution (duh): I'm easily able to work around this without modifying any core code via an `Extension` of `CMSForm` using the `->updateEditForm()` method in my extension. For example:


```php
class CMSMainExtension extends DataExtension {

	/**
	 * Update actions to disable validation so that we don't get validation errors when
	 * deleting things that have validation since, of course, they don't need it.
	 *
	 * @param Form $form
	 */
	public function updateEditForm(Form $form) {
		if ($form instanceof CMSForm) {
			$actions = array_merge($form->getValidationExemptActions(), ['delete']);
			$form->setValidationExemptActions($actions);
		}
	}

}
```

Logically you'd need to register the extension as well, like in your `config.yml`:

[...]

PR #1287, opened 2015-10-01
FIX for #586: Don't perform validation upon deletion, since it isn't necessary.

patricknelson:
Fix for #586 and possible fix for #736 and relates to https://github.com/kinglozzer/silverstripe-cms/commit/66bfff4d150ce5a6544ff5a21c2644c1469a1761. I believe this addresses bugs relating to removing a newly created draft page which also happens to have validation on it. May also apply to any other validated objects being handled via `CMSForm`.

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -644,7 +644,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 			// if($form->Fields()->hasTabset()) $form->Fields()->findOrMakeTab('Root')->setTemplate('CMSTabSet');
 			$form->setAttribute('data-pjax-fragment', 'CurrentForm');
 			// Set validation exemptions for specific actions
-			$form->setValidationExemptActions(array('restore', 'revert', 'deletefromlive', 'rollback'));
+			$form->setValidationExemptActions(array('restore', 'revert', 'deletefromlive', 'delete', 'rollback'));
 
 			// Announce the capability so the frontend can decide whether to allow preview or not.
 			if(in_array('CMSPreviewable', class_implements($record))) {


kinglozzer:
@patricknelson Do we need the `unpublish` action added too? Given that `publish` invokes validation (and _should_ prevent users from ever publishing an “invalid” page in the first place) it shouldn’t ever happen, but in my mind it makes sense to make all destructive actions exempt from this sort of validation.   What do you think?

patricknelson:
Probably, yeah.

For me a very valid use case was `delete` simply because you can get to it by proxy of simply creating (in which case validation may have *never* occurred). In the case of `unpublish` you're likely to have already validated anyway (99% of the time) since, as you said, you will have had to validate first to get it into that published state. **However:** That of course doesn't take into account the possibility that you've edited the form prior to `unpublish` but, in that case, the `unpublish` would have to perform a validation/write (does it?).

kinglozzer:
I’m not sure whether unpublishing will invoke a write first, but even if it does - would removing validation be an issue? You can already create a draft page with “invalid” data (as demonstrated by the bug being fixed!), so I don’t think it’s an issue if you can unpublish “into” an invalid state.

Pinging @silverstripe/core-team for a second opinion

patricknelson:
Right now since it's 5:30AM for me and I'm not on my work computer, I'm too lazy to really verify if it does a write. However, no matter what, if it does write I think correct functionality should be to always validate regardless. So, notwithstanding intermediate form modifications (between current publish state and user clicking unpublish) I think that not including `unpublish` action would generally be the best route to take, even if it did validate -- precisely for that reason (due to a possible write).

This is important I think for database consistency, since validation is likely very important to those who write it :) So I've flopped from "probably" to "probably not"!

svandragt:
Regarding unpublish, it's possible that the author wants to unpublish content whose datamodel has changed since publishing. For example new validation was added to make the datamodel more conformant based on real world usage.

So I could have a blog where the title was optional because I was publishing tweets but because of a design change titles are now mandated (eg summary pages). I should be able to unpublish the tweets that I don't want to add a title to, imho.

Or I have added completely new functionality which existing items do not have yet, where said new functionality has validation requirements.

dhensby:
Validation in the CMS is all over the place; really you should be able to save invalid pages IMO.

But that's a side point - really this needs a test to make sure this does what it's meant to!

dhensby:
I think the approach of "is this valid or not" is too limiting for complex models like `SiteTree` objects.

Maybe there's something that is literally invalid and can't be saved (unique index) but then on the other hand, something that's required for publishing (adding a blog summary). With the latter, a content author would obviously like to be able to draft content without publishing it, so saying it's invalid is unhelpful.

Personally, I just don't do CMS validation, there are too many edge-cases - it's for the CMS user to make sure the page is broken when they publish it.

tractorcow:
Sometimes pages can be made invalid due to changes in other components (e.g. parent page). This is especially the case with complicated custom validators that deal with relationships.

I'm in favour of adding unpublish to the exclusion list as well. Do you think you are able to add a test to prove CMSMain allows delete action? If you find it awkward we can probably just merge, but it's definitely preferable to have a test. :) Even if it's just `$main = new CMSMain(); $form = $main->getEditForm(); $this->assertContains('delete' ,$form->getValidationExemptActions());`

patricknelson:
@tractorcow In light of your comment on an unrelated ticket here, should I proceed with the extra minor one line update (incorporation of `unpublish`) and the new tests on the `3.2` or `3` branch instead of here, or, keep it where it's at and not open another PR?

tractorcow:
I think we should probably move this to 3, if that's ok with you. If it wasn't for 3.2 just about to go stable I'd say there, sorry.

You're doing some good work lately; Keep it up!

patricknelson:
@tractorcow  No problem and thank you! 

**Quick question:** Will things going into `3` (right now before stable) eventually make their way into `3.2`? Like, in a patch version like `v3.2.1` for example. I'm asking because I'm making lots of fixes that I need and have some features that I'm using that are directly committed to my repository that I have to manually workaround as I apply security fixes (etc) due to a lag in the release cycle. Like the `TabSet` issue I posted literally minutes before you merged these :) https://github.com/silverstripe/silverstripe-framework/issues/4663

patricknelson:
@tractorcow  If so, I can take an initiative to go through all of the PR's (and issue) that I've submitted and rebase them on `3` as soon as possible to clean things up ASAP. Since I've got some PR's that go back to July or earlier and most (if not all) of them are based probably on `3.1`.

tractorcow:
Changes in 3 won't be merged back into 3.2... it goes the other way only sorry.
What we should have is a 3.2.0 branch for the current RC so that changes for 3.2.1 can be merged into the 3.2 branch (but we don't at the moment).

I would say come back to this after the 3.2.0 stable is out next week and we can try to fix it then. Sorry for being awkward. :D

patricknelson:
@tractorcow  Are you saying that right now there's no particular branch that I can commit to in order to ensure that it will make it into `3.2.1` (for example)?  This seems to be a bug fix. As such it *should* presumably make it into `3.2.x` eventually. 

What I think I might do instead, then, is go ahead and address this (and any other bug fixes) now or as soon as I can directly on `3.2` branch and resubmit my PR's and queue/wait for the RC to come out so that then, in all likelihood, they'll make it into `3.2.1`. Does that make sense?

tractorcow:
you can pull against 3.2. We just won't release the 3.2.0 from this branch, but 3.2.1 will.

patricknelson:
Ok, great. I'm glad I waited for your feedback first before doing this on `3`!

I'm fine with that on anything bug related too. This is important for me since I've got a desire to migrate a very large and important site from `3.1.x` to `3.2.x` and want as few of my custom modifications in there as possible :smile:

patricknelson:
@tractorcow @dhensby @kinglozzer 

Setup new PR based on `3.1` here complete with a few tests: https://github.com/silverstripe/silverstripe-cms/pull/1297

tractorcow:
Fixed in https://github.com/silverstripe/silverstripe-cms/pull/1297

It'll be in 3.2.1

PR #1291, opened 2015-10-05
API CMS Implementation of RFC-1 Asset Abstraction (DataObject Refactoring) 

tractorcow:
Support new asset abstraction and remove filesystem sync

Necessary pull request companion to https://github.com/silverstripe/silverstripe-framework/pull/4652


code/controllers/AssetAdmin.php
```diff
@@ -62,13 +58,11 @@ public function currentPageID() {
...
 
 	/**
-	 * Set up the controller, in particular, re-sync the File database with the assets folder./
+	 * Set up the controller
 	 */
 	public function init() {
 		parent::init();
 
-		// Create base folder if it doesnt exist already
-		if(!file_exists(ASSETS_PATH)) Filesystem::makeFolder(ASSETS_PATH);
```
chillu: Confirmed that flysystem creates `ASSETS_PATH` automatically: https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Local.php#L73

tractorcow: If it's using that adapter, it does. :)

tractorcow:
Since file link tracking was broken in the prior framework merge, this pull request will rely on https://github.com/silverstripe/silverstripe-framework/pull/4679 before it can pass tests.

File::updateLinks has been moved to be invoked after write, and link rewriting now uses the same mechanism as HtmlEditorField, which is functionality refactored out of that class into `Image:: regenerate_html_links`.

tractorcow:
@sminnee would you like to review my implementation of ErrorPage logging? Hopefully I've built on top of your work in the most appropriate way. :)

chillu:
I've merged the underlying framework PR and gave the tests a kick

tractorcow:
Hope composer picks up the change straight away.

tractorcow:
Damn it, something broke again. =(

tractorcow:
fulltextsearch components (obviously not those tested in framework tests) needed fixing. See https://github.com/silverstripe/silverstripe-framework/pull/4682 for remaining blocker for this pull request.

sminnee:
The MySQL build passed; re-running the others now. Fingers crossed!

sminnee:
Looks like SQLite, PostgreSQL, and the Behat extension still have issues.

tractorcow:
Yikes, it looks like behat extension needs updating for master.

tractorcow:
I'm getting really fed up with these postgres errors. =(

tractorcow:
Ok, so here's the problem:

SiteTreeBacklinksTest has hard coded file ids, which postgres doesn't like (since it doesn't reset the increment id in PostgreSQLDatabase::clearTable).

I tried changing it to use truncate, and it fixed that problem, but now there's a primary key violation error in the tests.

tractorcow:
HMMM postgres is passing finally, but a minor behat failure is going on still. =(

tractorcow:
I'm running this on `develop` branch temporarily so that I can extract the travis screenshots and debug this issue.

tractorcow:
Tests are reporting success, but if you inspect the behat log you'll see that one of them failed.

tractorcow:
The test passes running behat locally.

I think we should just probably merge this.

tractorcow:
I've merged develop into master of behat-extension and removed the reference in this pull request.

tractorcow:
That broken behat test is fixed. ;)

Issue #1294, opened 2015-10-08
Fatal error when saving ErrorPage: Referencing undefined var and calls method on non-object.


patricknelson:
I noticed that `ErrorPage` fails when performing a `->doPublish()`. There are two main issues (the second being fatal) and affects **all branches**:

1. Undefined `$errorPage`. I think this is supposed to be `$filePath`
2. An method `->httpError()` is being called on an undefined attribute of `$this->response` (or method `->getResponse()`) which probably shouldn't be happening outside of a controller anyway, since this is essentially a database level operation (`->doPublish()` inside a `DataObject`) and there may not necessarily be a request associated with that particular execution path, but I digress.

**Affects branches:**

- `3` https://github.com/silverstripe/silverstripe-cms/blob/3/code/model/ErrorPage.php#L260 (not sure why this branch alone references `->getResponse()`)
- `3.1` https://github.com/silverstripe/silverstripe-cms/blob/3.1/code/model/ErrorPage.php#L260
- `3.2` https://github.com/silverstripe/silverstripe-cms/blob/3.2/code/model/ErrorPage.php#L260
- `master` https://github.com/silverstripe/silverstripe-cms/blob/master/code/model/ErrorPage.php#L260

To reproduce easily, just change this code (example from `3.1`):

```php
		if($fh = fopen($filePath, "w")) {
```

... to this (with `false` there)...

```php
		if(false && $fh = fopen($filePath, "w")) {
```

patricknelson:
Closing. Resolved via merge of #1295 and #1296.

PR #1295, opened 2015-10-08
FIX for #1294 to workaround ErrorPage fatal errors (and undefined var) when publishing.

patricknelson:
Addresses #1294. Also ensuring the method follows the parent method definition! Should return `bool`.

**Note:** I'm assuming that `user_error()` is the best route here. I'm not sure how else to register that sort of error.

### DIFF
--- a/code/model/ErrorPage.php
+++ b/code/model/ErrorPage.php
@@ -222,11 +222,10 @@ class ErrorPage extends Page {
 	 * content, so the page can be shown even when SilverStripe is not
 	 * functioning correctly before publishing this page normally.
 	 *
-	 * @return void
+	 * @return bool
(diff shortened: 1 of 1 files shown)


patricknelson:
Going to resubmit this against `3.1` since I realize now it's similar enough! Sorry.

tractorcow:
I actually ran into this issue myself recently. Thanks for the fix.

patricknelson:
No problem @tractorcow -- there's a related PR #1296 for the same issue for the `3.1` branch, should we close that since `3.2` is now released? Or is this the sort of bug fix you'd bother to back port to the older branch? PR is already there, so I figured "maybe" :)

PR #1297, opened 2015-10-09
FIX for #586 and possible fix for #736 and relates to #2449: Don't perform validation upon deletion, since it isn't necessary.

patricknelson:
**Note:** This is a duplicate of (and also supersedes) the previous PR #1287 but rebased upon the `3.2` branch instead of `3.1`. 

---

Fix for #586 and possible fix for #736 and relates to https://github.com/kinglozzer/silverstripe-cms/commit/66bfff4d150ce5a6544ff5a21c2644c1469a1761. I believe this addresses bugs relating to removing a newly created draft page which also happens to have validation on it. May also apply to any other validated objects being handled via `CMSForm`.

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -593,7 +593,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 	/**
 	 * @param int $id
 	 * @param FieldList $fields
-	 * @return Form
+	 * @return CMSForm
 	 */
 	public function getEditForm($id = null, $fields = null) {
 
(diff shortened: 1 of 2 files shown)


patricknelson:
Apologies in advance for the white space & formatting. I didn't notice the IDE did that until after the PR. however, I don't disagree with the changes so let me know if you want me to revert that :smile:

youpixxl:
Hi,
I am using uncle-cheese better button for deletion,and the method here is 'doDelete'. Is there a way to add custom exempt action to form ?

patricknelson:
@youpixxl Try first calling `->getValidationExemptActions()`, adding your action to the list, and then passing that updated array to `->setValidationExemptActions()`. Only issue then I guess is when/where you call it, since if this code runs after yours, that change would be lost. Also I would t know precisely where you'd call that off hand. @tractorcow?

youpixxl:
Hi @patricknelson and thanks a lot for your answer.
However I did not found a way to call that method as I don't have the form available : is there a way to retrieve the form in the getCMSFields function for a dataobject ? Also, I tried to add manually the action to the list in CMSMain.php (just to check) and it didn't solve my issue. Any Idea?

tractorcow:
You can also use `$formAction->setValidationExempt(true);` on individual actions.

patricknelson:
@tractorcow How could you get the form instance in his situation, however? I believe @youpixxl  is attempting to do this from some sort of `DataObject` context, which seems difficult. 

Another approach may be to simply override `->getCMSValidator()` and then return a benign validator which will not trigger any errors.

youpixxl:
@patricknelson, @tractorcow, I finally found the right place to call `->getValidationExemptActions()` and `->setValidationExemptActions()` (in the modelAdmin extension, getEditForm) and your solution works like a charm.
Thanks a lot for your support, and by the way I love this frameworks :-) !

PR #1300, opened 2015-10-14
Generated files API

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/4680 for the core pull request

CMS-specific errorpage handling has been refactored out of framework module into a separate FriendlyErrorFormatter, which replaces the core framework FriendlyErrorFormatter when cms module is installed.

The ErrorPage static caching has been refactored into the new generated file handling API. However, for the sake of apache error handling (which relies on a physical file at `/assets/error-XXX.html`) an option to continue to write to this specific location (`ErrorPage.enable_static_file`) is included (on by default).

### DIFF
--- a/code/model/ErrorPage.php
+++ b/code/model/ErrorPage.php
@@ -1,4 +1,6 @@
 <?php
+
+use SilverStripe\Filesystem\Storage\GeneratedAssetHandler;
 /**
  * ErrorPage holds the content for the page of an error response.
  * Renders the page on each publish action into a static HTML file
@@ -8,7 +10,8 @@
  * ErrorPages
  * 
  * @see Debug::friendlyError()
- * 
+ *
+ * @property int $ErrorCode HTTP Error code
  * @package cms
  */
 class ErrorPage extends Page {
@@ -25,11 +28,30 @@ class ErrorPage extends Page {
 	private static $allowed_children = array();
 	
 	private static $description = 'Custom content for different error cases (e.g. "Page not found")';
+
+	/**
+	 * Allows control over writing directly to the static_filepath directory. Only required if relying on
+	 * apache ErrorDocument, and can be turned off otherwise.
+	 *
+	 * @config
+	 * @var bool
+	 */
+	private static $enable_static_file = true;
 	
 	/**
 	 * @config
+	 * @var string
 	 */
 	private static $static_filepath = ASSETS_PATH;
+
+	/**
+	 * Prefix for storing error files in the {@see GeneratedAssetHandler} store.
+	 * Defaults to empty (top level directory)
+	 *
+	 * @config
+	 * @var string
+	 */
+	private static $store_filepath = null;
 	
 	/**
 	 * @param $member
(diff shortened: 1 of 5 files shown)


tractorcow:
@sminnee would you like to review my implementation of ErrorPage logging? Hopefully I've built on top of your work in the most appropriate way. :)

code/model/ErrorPage.php
```diff
@@ -274,47 +301,65 @@ public function fieldLabels($includerelations = true) {
...
-	 * 
-	 * @param int $statusCode A HTTP Statuscode, mostly 404 or 500
-	 * @param String $locale A locale, e.g. 'de_DE' (Optional)
+	 * Get error content for the given status code. Only returns pre-cached content
+	 * stored either in the asset store or in the static path (if enabled).
 	 *
+	 * @param int $statusCode A HTTP Statuscode, typically 404 or 500
 	 * @return string
 	 */
-	public static function get_filepath_for_errorcode($statusCode, $locale = null) {
```
chillu: Wanna note this removal in the 4.0.0 changelog?

code/model/ErrorPage.php
```diff
@@ -274,47 +301,65 @@ public function fieldLabels($includerelations = true) {
...
-		if (singleton('ErrorPage')->hasMethod('alternateFilepathForErrorcode')) {
```
chillu: Same here, removal of that `alternateFilepathForErrorcode` hook should be documented in changelog.

code/model/ErrorPage.php
```diff
@@ -274,47 +301,65 @@ public function fieldLabels($includerelations = true) {
...
-	static public function set_static_filepath($path) {
-		Deprecation::notice('4.0', 'Use "ErrorPage.static_file_path" instead');
-		self::config()->static_filepath = $path;
+	protected static function get_error_filename($statusCode, $instance = null) {
+		if(!$instance) {
+			$instance = ErrorPage::singleton();
+		}
+		// Allow modules to extend this filename (e.g. for multi-domain, translatable)
+		$name = "error-{$statusCode}.html";
+		$instance->extend('updateErrorFilename', $name, $statusCode);
```
chillu: We should add a github issue to subsites once this is merged to adapt the new API, could easily be missed otherwise (https://github.com/search?q=alternateFilepathForErrorcode&type=Code&utf8=%E2%9C%93)

code/model/ErrorPage.php
```diff
@@ -274,47 +301,65 @@ public function fieldLabels($includerelations = true) {
...
-	 * @param String $locale A locale, e.g. 'de_DE' (Optional)
+	 * Get error content for the given status code. Only returns pre-cached content
+	 * stored either in the asset store or in the static path (if enabled).
 	 *
+	 * @param int $statusCode A HTTP Statuscode, typically 404 or 500
 	 * @return string
 	 */
-	public static function get_filepath_for_errorcode($statusCode, $locale = null) {
-		if (singleton('ErrorPage')->hasMethod('alternateFilepathForErrorcode')) {
-			return singleton('ErrorPage')-> alternateFilepathForErrorcode($statusCode, $locale);
```
chillu: Did you remove `$locale` support on purpose? Subsites uses it for (optional) translatable support: https://github.com/silverstripe/silverstripe-subsites/blob/9c9e0bfa94a568e54b37b431529293da09bf8eb9/code/extensions/ErrorPageSubsite.php#L31

I guess it's only one of many variants which could influence the error page content and presentation, so in general I agree we shouldn't pass it through here as a parameter. The extension hooks can get all the info they need from global state anyway (current subsite, current locale). Just making sure it's been actively considered.

sminnee: If we remove locale we'll forget to add support for it back in later with a yet-to-be-planned 'better solution'. My preference would be to retain locale for now and look at refactoring later. I think Hamish calls the proper solution the "state set API"

code/model/ErrorPage.php
```diff
@@ -47,36 +69,32 @@ public function canAddChildren($member = null) {
...
-			);
+			return ModelAsController::controller_for($errorPage)
+				->handleRequest(
+					new SS_HTTPRequest('GET', ''),
+					DataModel::inst()
+				);
 		}
 		
 		// then fall back on a cached version
-		$cachedPath = self::get_filepath_for_errorcode(
```
sminnee: Yes, this is coupling, but we didn't put it in for jollies: we've tolerated the coupling in here because it implements an important feature. Refactor it out in a future PR, but don't just delete it in this one!

I don't think that making the state set API a blocker for 4.0-alpha1, which would be the alternative solution, is a good idea.

Please re-add the coupling to Translatable.

code/model/ErrorPage.php
```diff
@@ -47,36 +69,32 @@ public function canAddChildren($member = null) {
...
-			$statusCode, 
-			class_exists('Translatable') ? Translatable::get_current_locale() : null
```
sminnee: Will `Translatable::updateErrorFilename()` completely replace the need for this coupling? If so, could we add a GitHub issue for it in translatable, as well as subsites?

code/model/ErrorPage.php
tractorcow: It's removed, because it's not necessary for the core to respect specific implementations.

My intention was to provide `updateErrorFilename` as sam suggested, as an alternative to this. All that was done here was to remove the locale parameter, but that can be retrieved by `Translatable::get_current_locale()` in the extension (rather than in the ErrorPage.php itself) in any case.

code/model/ErrorPage.php
tractorcow: Yes. :D I can add an issue.

tractorcow: Actually, it would be best to put this into the upgrading notes.

tractorcow:
https://github.com/silverstripe/silverstripe-framework/pull/4699 needs merging to restore missing functionality.

PR #1308, opened 2015-10-27
Revert use of GeneratedAssetHandler in ErrorPage

tractorcow:
Reverts much of https://github.com/silverstripe/silverstripe-cms/pull/1300

After implementing this change earlier I realised there was an architectural flaw in how ErrorPage was using the GeneratedAssetHandler service.

This service is a caching mechanism for generated content, and CANNOT be used to pre-generate content which may not be re-generated on-demand, as is possible with combined files.

The underlying caching mechanism, SS_Cache, also is only a caching mechanism, not a data store. Using it as such would be disastrous on clustered systems, with multiple nodes each having their own locale cache.

In place of this service, I have left the original functionality in place with the ability to customise the `static_filepath` via config.

See https://github.com/silverstripe/silverstripe-framework/pull/4709

### DIFF
--- a/code/model/ErrorPage.php
+++ b/code/model/ErrorPage.php
@@ -28,30 +28,12 @@ class ErrorPage extends Page {
 	private static $allowed_children = array();
 	
 	private static $description = 'Custom content for different error cases (e.g. "Page not found")';
-
-	/**
-	 * Allows control over writing directly to the static_filepath directory. Only required if relying on
-	 * apache ErrorDocument, and can be turned off otherwise.
-	 *
-	 * @config
-	 * @var bool
-	 */
-	private static $enable_static_file = true;
 	
 	/**
 	 * @config
 	 * @var string
 	 */
 	private static $static_filepath = ASSETS_PATH;
-
-	/**
-	 * Prefix for storing error files in the {@see GeneratedAssetHandler} store.
-	 * Defaults to empty (top level directory)
-	 *
-	 * @config
-	 * @var string
-	 */
-	private static $store_filepath = null;
 	
 	/**
 	 * @param $member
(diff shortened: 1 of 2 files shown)


chillu:
Technically, this just moves the problem of shared state from `SS_Cache` to the filesystem - but since any multi-node setup will already have to deal with this (through NFS drives or csync), it's a non-issue.

But with the new assets system, the need for asset sync hopefully would go away because you can centralise this state, e.g. in AWS S3. Given `Requirements::combine_files()` now supports asset backends, would the error page be the only remaining use case requiring any form of state syncing between nodes? That's a lot of infrastructure complexity for comparatively little gain (static error pages).

tractorcow:
Another solution is to have an API for un-versioned assets. I.e. assets with only 'filename' and no variant / hash. That way there is nothing that needs to be stored?

jonom:
I notice [RFC-2](https://github.com/silverstripe/silverstripe-framework/issues/4149) which proposes to remove ErrorPage has an accepted label now... does this change anything for you @tractorcow? As noted in the RFC the static 404 is useless out of the box and the 500 static page is hazardous because it is regenerated so rarely and goes stale quickly. (Sorry I haven't made any progress on implementing that RFC yet, I hadn't realised it had been accepted.)

tractorcow:
@jonom accepted label is an accident. :D It was added as "rfc" and the label was renamed to "rfc/accepted" after the fact, it seems.

tractorcow:
Had a quick chat to @hafriedlander and solution seems hard.

We will pick this up in a few weeks again when we can come up with a solution.

I feel it will either be "apl no longer needs a hash" or "we don't use apl for static error pages".

jonom:
> accepted label is an accident. :D It was added as "rfc" and the label was renamed to "rfc/accepted" after the fact, it seems.

Oh lol, thought it seemed a bit odd.

tractorcow:
Ok, I've had a chat with @sminnee and I think we've come up with an alternative solution.

https://github.com/silverstripe/silverstripe-framework/blob/master/filesystem/storage/CacheGeneratedAssetHandler.php

Rather than the default asset handler using SS_Cache to store the tuple (filename, hash, etc) we only identify assets by their filename.

Rather than use AssetStore (which relies on a hash) we inject a flysystem instance directly as a dependency.

tractorcow:
This fix has been superceded by https://github.com/silverstripe/silverstripe-cms/pull/1337

PR #1314, opened 2015-11-04
FIX: Setting hide_ancestor=true causes a random page type to be hidden

kinglozzer:
Yes, I know that `hide_ancestor` is meant to be a string, not boolean! However as the config system can’t warn us about that, this prevents the first page type (usually `Page`) from incorrectly being hidden if it’s set to true (the name _does_ suggest it’s a boolean after all). Plus stricter code = better code, right?

tldr:
```php
array_search(true, array('somestring')) === 0 // meaning that the first class is removed
array_search(true, array('somestring'), true) === false // so we can correctly do nothing
```

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -392,7 +392,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	 *
 	 * @return array
 	 */
-	static public function page_type_classes() {
+	public static function page_type_classes() {
 		$classes = ClassInfo::getValidSubClasses();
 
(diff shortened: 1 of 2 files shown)


tractorcow:
Wouldn't it be better to throw an \InvalidArgumentException if an invalid config value is detected?

dhensby:
It would be a stricter API but I don't think it would be better just because of that. 

I think flexibility is good as long as the maintenance is acceptable. 

Can we get some tests, please?

tractorcow:
Agreed; Tests and I'll merge. :)

kinglozzer:
Test added!

PR #1170, opened 2015-02-24
BUG Fix CMSMainTest attempting to render page on Security permission error

tractorcow:
Normally this method is posted via ajax. Leaving out the ajax parameter during mocking of these functions can occasionally cause the `Security::permissionFailure` method to attempt to render a page using the default `Page` class, occasionally breaking due to user-code (especially with custom Requirements).

Basically, this just makes the tests behave a bit more like real life and breaks less unit tests with certain usercode in place.

PR #1315, opened 2015-11-05
BUG Ensure CMSMainTest uses correct siteconfig

tractorcow:
Sometimes tests would end up with the default SiteConfig appearing alongside the one specified in the test fixtures.

Additionally, ->post requests which failed were causing Security::LoginForm() to be invoked, causing needless work being done (and potentially breaking if it encountered errors during page rendering).

Using ajax => 1 means that the tests will raise security permission failures appropriately without having to generate rendered error pages.

Some indentation was cleaned up, and a few tests strengthened.

### DIFF
--- a/tests/controller/CMSMainTest.php
+++ b/tests/controller/CMSMainTest.php
@@ -9,6 +9,18 @@ class CMSMainTest extends FunctionalTest {
 
 	static protected $orig = array();
 
+	public function setUp() {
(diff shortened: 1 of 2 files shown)


tractorcow:
This is a continuation of the work I started at https://github.com/silverstripe/silverstripe-cms/pull/1170

PR #1317, opened 2015-11-07
FIX Remove unnecessary config updates which clears memory cache.

micmania1:
There's probably a lot more of these, but every time Config::inst()->update() is called the Config_MemCache gets cleared meaning config that has already been merged has to be merged again.


micmania1:
Pending outcome of: https://github.com/silverstripe/silverstripe-framework/issues/4752

PR #1320, opened 2015-11-10
Restore deprecated reports

tractorcow:
Merge with https://github.com/silverstripe-labs/silverstripe-reports/pull/12

Fixes https://github.com/silverstripe/silverstripe-framework/issues/4703


tractorcow:
I've had to fix tests that were, up until now, broken, but not active (because the tested classes didn't exist).

tractorcow:
Thanks. :D

PR #1322, opened 2015-11-12
API Add page/admin redirect to admin section for that page

tractorcow:
Because I am so lazy and hate having to find the page again after logging into the CMS.

mypage/admin will send me to the right place to immediately edit that page.

### DIFF
--- a/code/controllers/ContentController.php
+++ b/code/controllers/ContentController.php
@@ -25,6 +25,7 @@ class ContentController extends Controller {
 	private static $extensions = array('OldPageRedirector');
 
 	private static $allowed_actions = array(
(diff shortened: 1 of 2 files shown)


micmania1:
Doesn't that risk confusing/frustrating users? If they're already on /blog/archive they'll try and add admin to the end and it won't work.

(archive being an action)

tractorcow:
Yeah that is a risk; `admin` can't be applied on top of existing actions, but you can always remove `archive` and still get redirected.

tractorcow:
Doh, I pushed to origin not to my fork, sorry guys. =(

PR #1323, opened 2015-11-12
API Add page/admin redirect to admin section for that page

tractorcow:
Replacement for https://github.com/silverstripe/silverstripe-cms/pull/1322, but not pushed to origin. :)


dhensby:
So this is an action to allow people to quickly edit a particular page in the CMS?

I think this is a nice to have feature that is non-essential for the use of the CMS and, thus, adds code complexity and maintenance...

I think you should be installing a module like https://github.com/jonom/silverstripe-betternavigator if you want this kind of feature.

tractorcow:
I agree. This is nice and easy, so it'll be easy to modularise. Closing

PR #1325, opened 2015-11-15
NEW allow for setting Current, Section and Link values via Config system

muskie9:
The goal of this is to allow for setting the `current`, `section` and `link` return values for easier use with various front-end frameworks such as bootstrap. It uses the Config system to set the default value and allow the developer to override that value depending on their project need.

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -235,6 +235,30 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	private static $meta_generator = 'SilverStripe - http://silverstripe.org';
 
 	protected $_cache_statusFlags = null;
+
+	/**
+	 * CSS Class to use for links matching the currently viewed page.
+	 *
+	 * @config
+	 * @var string
+	 */
+	private static $linking_mode_current = 'current';
+
+	/**
+	 * CSS Class to use for links matching the currently viewed page.
+	 *
+	 * @config
+	 * @var string
+	 */
+	private static $linking_mode_section = 'section';
+
+	/**
+	 * CSS Class to use for links matching the currently viewed page.
+	 *
+	 * @config
+	 * @var string
+	 */
+	private static $linking_mode_link = 'link';
 	
 	/**
 	 * Determines if the system should avoid orphaned pages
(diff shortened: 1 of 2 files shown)


muskie9:
Apologies for the formatting, I have my local environment set to use spaces rather than tabs. I know there is an active discussion regarding code style conventions. If tabs are preferred at this point let me know and I can update the PR.

muskie9:
I'll update the formatting to tabs and look into the docs as mentioned by @jonom and see what I can come up with.

muskie9:
see silverstripe/silverstripe-framework#4785

dhensby:
@jonom makes a very good point.

Modules that rely on what `LinkingMode` returns will break if anyone customises them. They are effectively "constants", I don't think they should be customised in this way.

Devs who customise the linkingmodes could override ALL the templates that make assumptions about this but it seems a bit nasty to me.

We have to make a compromise between being able to have control of these classes vs maintaining compatibility.

my solution for master would probably be to have `isSection`, `isLinkOrSection`, `isCurrent` to return boolean values and then `Section`, `LinkOrSection`, etc return the classes.

tractorcow:
After reading @dhensby's comment I tend to agree. It would probably be more risky to make this no longer act as a constant.

If you need to have a custom css class name, you could implement it using a method separate from `LinkingMode`.

In fact, I would go as far as to define some `const` to that effect to enforce this in core, rather than the string literals.

dhensby:
I think I'm ok with letting it be customised, but devs need to accept the real risk this won't work well with modules. 

Not allowing this to be customised does seem overly restrictive for little benefit.

muskie9:
I guess after all the discussion I could go either way. I didn't think of modules' templates/layouts as I'm usually working on projects where we do our own templates/layouts. I'm all for using the `is` methods as it essentially adds the flexibility without needing to change code. Would it be worth deprecating the methods returning the string values, customizable or not, in 4 and update docs to reflect the change?

jonom:
@muskie9 considering you're happy either way I'd probably lean towards scrapping this PR and instead updating the docs to use the `is` methods. No point adding complexity if we don't need to :)

I'm not sure about deprecation... `LinkOrCurrent` and `LinkOrSection` don't seem super useful to me but others might disagree. `LinkingMode` seems more useful - I don't think it's an ideal function, because 'section' and 'current' aren't mutually exclusive. Neither is 'link' for that matter - in my menus every item including the current page gets a hyperlink. I suspect that function gets used a lot though so shouldn't be removed.

For me the ideal menu item would probably look like this
```html
<a class="
		<% if $isSection %>section<% end_if %>
		<% if $isCurrent %>current<% end_if %>
	" href="foo">Foo</a>
```

dhensby:
From a purist point of view, we should scrap the functions that output unchangeable class names but it probably is purity for purity's sake because they actually have a use.

really, the framework should never be defining classnames but it is quite hard to avoid.

muskie9:
Thanks for working through this, I'll update my other PR in framework to reflect the `is` functions in the docs. At least I know of a simpler solution for next time ;)

PR #1327, opened 2015-11-16
FIX: call getParent() in orphaned record checks

stevie-mayhew:
In the situation where you have a record with a `Parent()` method or `ParentID` has_one this check can fail.

This was found during an upgrade from 3.1 -> 3.2 and is on the blog module - still looking where it occurred to see if its user code or blog code :)

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -608,7 +608,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 		if(empty($this->ParentID)) return false;
 		
 		// Parent must exist and not be an orphan itself
-		$parent = $this->Parent();
+		$parent = $this->getParent();
 		return !$parent || !$parent->exists() || $parent->isOrphaned();
 	}
 	


stevie-mayhew:
Appears this root cause of this issue has already been fixed at https://github.com/silverstripe/silverstripe-lumberjack/commit/bf0db23c1d606ca62fef87ee038223e35f751151

tractorcow:
Great, I assume I can close this then. :D

stevie-mayhew:
@tractorcow the root cause doesn't mean that this isn't an issue for other instances where there is a `Parent` or `ParentID` defined in other modules. I think being explicit about the function call rather than relying on the magic method is a good idea.

tractorcow:
If Parent or ParentID is defined in other modules, then that'd be the cause of a problem there too.

In any case, I guess the fix you're proposing doesn't hurt.

tractorcow:
There are 6 other matches for `$this->Parent()` in SiteTree.php... won't all of those other calls have the same problem?

Maybe a better fix is to deprecate getParent() in SiteTree, and then create a new Parent() method in site tree, that overrides the has_one declared in Hierarchy. Otherwise we end up with two methods; getParent() and Parent(), which both attempt to do the same thing, with only slight nuances.

We won't be able to do this until 3.4, but shall we address it there?

stevie-mayhew:
Sounds good, will look at writing the deprecations for a 3.3 PR this week.

PR #1330, opened 2015-11-17
FIX: Ensure VirtualPage forwards request/response data to virtual controllers (fixes #1329)

### LINKED ISSUE #1329: VirtualPage pointing to a RedirectorPage doesn’t work
If you point a `VirtualPage` at a `RedirectorPage`, it errors because `RedirectorPage::getResponse()` returns null. This is because `VirtualPage_Controller::__call()` magic will instantiate a controller so it can behave like the target page, but doesn’t set the request/response on that controller.

Two routes to fix this:

1. Make `RedirectorPage` check the result `$this->getResponse()` before attempting to call `isFinished()` on the response;
2. Make sure that the current request/response data is available on the “virtual” controller that’s being called.

In my opinion, option 2 is the better fix. Currently if you were to create a controller that uses request data and point a `VirtualPage` to it, the request data wouldn’t be available.

Proposed fix: https://github.com/kinglozzer/silverstripe-cms/commit/4b2ca41b136920cb1571924c71c388720c98458c

kinglozzer:


code/model/VirtualPage.php
```diff
@@ -413,7 +413,7 @@ public function CMSTreeClasses($numChildrenMethod="numChildren") {
 	public function __get($field) {
 		if(parent::hasMethod($funcName = "get$field")) {
 			return $this->$funcName();
-		} else if(parent::hasField($field)) {
+		} else if(parent::hasField($field) || ($field === 'ID' && !$this->getField('ID'))) {
```
kinglozzer: Without this change, the test fails: if you specify a target page _before_ writing, `DataObject::write()` thinks the VirtualPage already exists because it finds the ID of the target page.

I thought about always avoiding fetching the target page ID (or any other fixed fields like `Created` or `Updated`), but I figured doing it like this has the lowest BC risk.

kinglozzer:
I’m not really sure about the test here, it somehow feels wrong to be including references to `RedirectorPage` in `VirtualPageTest`... feedback appreciated

code/model/VirtualPage.php
dhensby: `!$this->exists()` is a nicer way of checking the ID is non-zero and being more self documenting

dhensby:
It is a bit nasty, I suppose, but you need to do this to test your bug.

Do a quick `class_exists` at the start of the test (if it doesnt exist then do `$this->markTestSkipped()`.

Thats the loosest way to couple it at the moment

code/model/VirtualPage.php
kinglozzer: Oh, awesome. I’d avoided using `inInDB()` because that triggers the magic lookup I’m trying to avoid, but `exists()` uses `$this->record['ID']` directly so shouldn’t suffer from the same issue

kinglozzer:
Okay, updated commits have been pushed, builds are passing (except behat because... well, behat - other PRs are failing in the same place)

PR #1331, opened 2015-11-17
FIX: RedirectorPage toggles not working (fixes #1328)

### LINKED ISSUE #1328: RedirectorPage internal/external link toggle doesn’t work
It currently just shows both options, the toggles do nothing:

screen shot 2015-11-16 at 09 48 14

kinglozzer:


dhensby:
Let's not use IDs if using IDs got us into the mess in the first place :P

kinglozzer:
Agreed it’s not perfect, but it’s the same pattern as the settings fields (e.g https://github.com/silverstripe/silverstripe-cms/pull/1281). Perhaps we should be using the `SelectionGroup` field (assuming it works? I’ve never even tried it) for this sort of thing.

Will relying on the existing IDs be okay for a patch release @dhensby? Or should I introduce classes to hook into or something?

dhensby:
I suppose it is a patch release...

PR #1342, opened 2015-12-08
Added avoiding unhandled exception from Translatable extension in ContentController

mejt:
Hi, I've found a bug with unhandled exception from Translatable extension in ContentController.
When you go to page with wrong locale parameter then you'll see 'User Error' on dev environment, but on production you'll see blank page (instead 404 or normal page). 
Try for example: http://youraddress/?locale=en_234


tractorcow:
I'll forgive unit tests, since you'd need to add translatable to even test this. ;)

PR #1352, opened 2015-12-28
Decouple ErrorPage

jonom:
This moves baked in ErrorPage functionality to an extension, so it can be split out in to an optional module in a subsequent PR. For motivation see https://github.com/silverstripe/silverstripe-framework/issues/4149

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -1,6 +1,9 @@
 LeftAndMain:
   extensions:
     - LeftAndMainPageIconsExtension
-Security:
+Controller:
+  extensions:
+    - ErrorPageControllerExtension
+Form:
   extensions:
     - ErrorPageControllerExtension
--- a/code/controllers/ContentController.php
+++ b/code/controllers/ContentController.php
@@ -211,16 +211,6 @@ class ContentController extends Controller {
 	}
 
 	/**
-	 * @uses ErrorPage::response_for()
-	 */
-	public function httpError($code, $message = null) {
-		// Don't use the HTML response for media requests
-		$response = $this->getRequest()->isMedia() ? null : ErrorPage::response_for($code);
-		// Failover to $message if the HTML response is unavailable / inappropriate
-		parent::httpError($code, $response ? $response : $message);
-	}
-
-	/**
 	 * Get the project name
 	 *
 	 * @return string
@@ -423,12 +413,6 @@ HTML;
 		// Return 410 Gone if this site is not actually a fresh installation
 		if (!file_exists(BASE_PATH . '/install.php')) {
 			$this->httpError(410);
-		}
-		// The manifest should be built by now, so it's safe to publish the 404 page
-		$fourohfour = Versioned::get_one_by_stage('ErrorPage', 'Stage', '"ErrorPage"."ErrorCode" = 404');
-		if($fourohfour) {
-			$fourohfour->write();
-			$fourohfour->publish("Stage", "Live");
 		}
 
 		// TODO Allow this to work when allow_url_fopen=0
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -439,7 +439,6 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 		if (
 			   !($page = DataObject::get_by_id('SiteTree', $arguments['id']))         // Get the current page by ID.
 			&& !($page = Versioned::get_latest_version('SiteTree', $arguments['id'])) // Attempt link to old version.
-			&& !($page = DataObject::get_one('ErrorPage', array('"ErrorPage"."ErrorCode"' => 404)))   // Link to 404 page directly.
 		) {
 			 return; // There were no suitable matches at all.
 		}
--- a/code/controllers/ModelAsController.php
+++ b/code/controllers/ModelAsController.php
@@ -111,8 +111,7 @@ class ModelAsController extends Controller implements NestedController {
 		if(class_exists('Translatable')) Translatable::enable_locale_filter();
 
 		if(!$sitetree) {
-			$response = ErrorPage::response_for(404);
-			$this->httpError(404, $response ? $response : 'The requested page could not be found.');
+			$this->httpError(404, 'The requested page could not be found.');
 		}
 
 		// Enforce current locale setting to the loaded SiteTree object
--- a/tests/model/SiteTreeTest.php
+++ b/tests/model/SiteTreeTest.php
@@ -613,7 +613,6 @@ class SiteTreeTest extends SapphireTest {
 
 	public function testLinkShortcodeHandler() {
 		$aboutPage = $this->objFromFixture('Page', 'about');
-		$errorPage = $this->objFromFixture('ErrorPage', '404');
 		$redirectPage = $this->objFromFixture('RedirectorPage', 'external');
 
 		$parser = new ShortcodeParser();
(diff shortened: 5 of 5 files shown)


code/model/SiteTree.php
```diff
@@ -439,7 +439,6 @@ static public function link_shortcode_handler($arguments, $content = null, $pars
 		if (
 			   !($page = DataObject::get_by_id('SiteTree', $arguments['id']))         // Get the current page by ID.
 			&& !($page = Versioned::get_latest_version('SiteTree', $arguments['id'])) // Attempt link to old version.
-			&& !($page = DataObject::get_one('ErrorPage', array('"ErrorPage"."ErrorCode"' => 404)))   // Link to 404 page directly.
```
jonom: I guess I could leave this in (and related tests) by testing for existence of ErrorPage class, but since this covers an edge case which shouldn't occur and is not a great fallback anyway (who wants to click a link to a 404 page) I think it would be okay if we don't render anything in the case of an ID not pointing to any valid page ID.

code/controllers/ContentController.php
```diff
@@ -424,12 +414,6 @@ public function successfullyinstalled() {
 		if (!file_exists(BASE_PATH . '/install.php')) {
 			$this->httpError(410);
 		}
-		// The manifest should be built by now, so it's safe to publish the 404 page
```
tractorcow: Why is this removed?

jonom: 404 and 500 pages are created and published in populateDefaultRecords, so this is redundant as far as I can tell. This bit of code seems to assume a draft 404 page exists and needs publishing but I couldn't find any reference to that anywhere. It only handles the 404 case too, not 500 - so I assume this is a piece of legacy code that got forgotten about?

_config/config.yml
```diff
@@ -1,6 +1,12 @@
 LeftAndMain:
   extensions:
     - LeftAndMainPageIconsExtension
+ContentController:
```
tractorcow: I wonder if we should just go ahead and promote `ErrorPageControllerExtension` onto `Controller`?

That would cover Form::httpError, etc.

tractorcow:
I think someone else should have a look at this pull request too. @dhensby any opinion?

code/controllers/ContentController.php
tractorcow: Ah fair enough. :)

_config/config.yml
dhensby: Yes, probably should go on `Controller`

code/controllers/ErrorPageControllerExtension.php
```diff
@@ -9,7 +9,8 @@
 class ErrorPageControllerExtension extends Extension {
 
 	public function onBeforeHTTPError($statusCode, $request) {
-		$response = ErrorPage::response_for($statusCode);
+		// Don't use the HTML response for media requests
+		$response = $request->isMedia() ? null : ErrorPage::response_for($statusCode);
```
dhensby: What is a media request? A request for *.jpg ? Why not use it for those requests?

dhensby:
Seems fine to me, doesn't appear to do anything too crazy except simplify things, which I like.

Just a couple of comments

_config/config.yml
jonom: Okay sure. I was focused on changing as little as possible but no harm in sneaking in an improvement or two I suppose! So applying it to Controller means I don't need to specify ContentController, Security or ModelAsController as they all extend Controller. So it will just be on Controller and Form then, yes?

code/controllers/ErrorPageControllerExtension.php
jonom: This is merged from the ContentController::httpResponse() method. I assume it's to prevent page rendering overhead, like if you have a page with a bunch of broken links to static resources and you're not caching your 404 response effectively it could cause some stress on the server because each page load would result in 40 error page rendering cycles, which are redundant because the user will not see them.

I assumed I was following the current functionality here but actually I've just tested and static resources are receiving a full error page response prior to this change. Following this change you would get only a plain text error message for static resources. It looks like we have some intended functionality here that isn't currently working - my guess is that checking whether a request is media or not in ContentController::httpError() is too late in the chain as an error would be thrown before we get to this point? In that case I'm not sure if it's best to fix this once-intended behaviour, or remove it.

dhensby: I think full 404s are needed. aren't they cached in assets to prevent the overhead?

jonom: Ah well you would think so but see https://github.com/silverstripe/silverstripe-framework/issues/4149 :)

jonom: Yeah I agree full 404 response is better and consistent with current behaviour so I'll remove that check.

jonom: FYI: Example of missing image with full 404 response. This is using 3 branch without this PR in place. Note response time compared to found resources.

home_ _your_site_name

But even if you do a plain text response it still takes 150ms or so for me so it's not a massive saving.

tractorcow:
The isMedia() check is necessary because you can't return HTML content for a .jpg file.

Otherwise you end up with an  potentially getting populated with HTML.

jonom:
> The isMedia() check is necessary because you can't return HTML content for a .jpg file.

That would make sense for a 200 response but I don't think it applies to a 404 does it? Or any other error response. By that logic a plain text response wouldn't be any good either as that is just as unsuitable for the content of an image file. This check seems to be being skipped currently anyway - if you try adding `` to a template and look for it under the Network tab of dev tools and check the 404 response, you should find it's html - same as if you browse directly to http://site.dev/dontexist/blank.gif.

But yeah like you say, it won't matter once your pull request is merged in anyway :)

dhensby:
Indeed, that's not quite right that a 404 needs the same data type as the requested media (or the context you're requesting it in, which is more what @tractorcow is implying)

If someone navigates directly to a URL with a .jpg at the end, we should still serve the same 404 to them. 

We wouldn't server a "404 PDF" to a user, so why create a special image. I'd agree if the site was an image host, maybe.

dhensby:
@jonom I reckon we are getting the html 404 because Apache has a directive set to return the file in assets if a 404 is returned?! 

Maybe that's how this saves overhead?

jonom:
> We wouldn't server a "404 PDF" to a user, so why create a special image. I'd agree if the site was an image host, maybe.

I agree with the first sentence but not so much the second (not that you had a strong opinion about it). Admittedly I don't know much about HTTP but surely there is not an expectation that simply requesting something will guarantee a result? If a client says "hey, can I have this thing?" the server can say "Yes. Here you go." or "No. It doesn't exist." If the client gets a No it shouldn't try to use the response. And wouldn't MIME types come in to it somewhere?

> I reckon we are getting the html 404 because Apache has a directive set to return the file in assets if a 404 is returned?!

Nope - try manually editing the content of your error-404.html and see if you can ever get that file to be served up. One of the key points from https://github.com/silverstripe/silverstripe-framework/issues/4149 is that the static 404 page is never used, as far as I can tell. I believe that Apache directive would only kick in if Apache handled the whole request, but it's handed off to PHP so it doesn't take effect. Everyone has the impression that we serve static 404 pages, and we certainly imply this, but It's not true as far as I can tell.

I think that as the framework and CMS have evolved, the error handling process hasn't kept up, and we have legacy code in place that doesn't have any effect anymore. So I want to strip it back to a simple base that can be extended upon.

I was aiming to introduce minimal changes in this PR as it's not meant to implement what's proposed in https://github.com/silverstripe/silverstripe-framework/issues/4149, I just want to lay the groundwork so that ErrorPage could be separated out as a module later if that RFC is ever accepted. It's not a very sexy issue apparently so it's been a bit slow going :D if you have time @dhensby it would be awesome if you could review that RFC and add any thoughts you have. It's fairly short and recommends quite minimal changes (making ErrorPage optional so alternative error page handling approaches can be implemented). @tractorcow not sure if you've reviewed it but I've simplified it a bit lately so might be worth another look.

I don't think I should re-introduce the isMedia() check in this PR because that was previously only used on the ContentController and that would mean it's now getting used on all Controller sub-classes, and actually has an effect now. As far as I can tell it didn't get a chance to kick in before so may as well have not existed. Also the isMedia() check is based on file extension alone (not MIME type) and only covers a handful of extensions so it's not very robust. @tractorcow maybe you can provide an example of where that isMedia() check would kick in normally upon a httpError() and be useful - is there some specific use I haven't thought of? If so I guess we should try to add a test for it.

tractorcow:
Well, in the past you wouldn't normally have contentcontroller handling requests for image content, so it's not like it really had any benefit in the past. :) It's a feature that wasn't really hit.

I think given that, we can probably just merge this as is, and ASSUME that image 404s will be handled by the asset system. I'll move the solution for media file 404s into the other pull request (which I will update in a few days).

@dhensby sound ok?

dhensby:
Ok

jonom:
:clap: thanks! One step closer to ousting ErrorPage... mwah hah hah

tractorcow:
Progress is made my taking risks. :)

Issue #1334, opened 2015-11-23
Changing Page class prevented if SiteConfig.CanCreateTopLevel permission is missing 


sb-relaxt-at:
If a user does not have the permission to create pages at the top level, he cannot change the class of an existing page. This is due to SiteTree#getClassDropdown using canCreate to determine which page types should be shown in the Settings-Tab. As canCreate is called without parameters, it falls back to checking whether or not the user is allowed to create the page on the top level. I suggest to [add a context parameter](https://github.com/silverstripe/silverstripe-cms/blob/3/code/model/SiteTree.php#L2679): 

    !$instance->canCreate(null, array('Parent' => $this)

As this seems to work, and I do not believe that there is any side effect, I will try to provide a pull request asap.

tractorcow:
Should it be canCreate() || canEdit() ?

sb-relaxt-at:
Normally creating a Page seems to be only limited by canCreate. It is possible for a user to create a Page, although it cannot edit this Page after creating. Therefore it would be only consistent to just check canCreate.

tractorcow:
See https://github.com/silverstripe/silverstripe-cms/issues/1333#issuecomment-160462954 to link these two discussions. :)

tractorcow:
Ok, I think these two issues are similar but still separate.

I would suggest setting the context parameter to the parent of the current object, rather than the object itself.

```php
!$instance->canCreate(null, array('Parent' => $this->Parent())
```

sb-relaxt-at:
Yes of course, my mistake.

PR #1353, opened 2016-01-02
Adding context parameter to canCreate-check in getClassDropdown

sb-relaxt-at:
Fix as suggested in #1334

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -2639,7 +2639,8 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 
 			// if the current page type is this the same as the class type always show the page type in the list
 			if ($this->ClassName != $instance->ClassName) {
-				if((($instance instanceof HiddenClass) || !$instance->canCreate())) continue;
+				if($instance instanceof HiddenClass) continue;
+				if(!$instance->canCreate(null, array('Parent' => $this->ParentID ? $this->Parent() : null))) continue;
 			}
 			
 			if($perms = $instance->stat('need_permission')) {


tractorcow:
It looks as though the test cases aren't working; Might be that the tests themselves were wrong?

sb-relaxt-at:
Sorry, my bad. The testcases are fine. The Parent() method returns an unsaved SiteTree object if there is no parent specified. The canCreate method doesn't care whether or not the supplied parent object is saved and performs a check of allowedChildren (https://github.com/silverstripe/silverstripe-cms/blob/3/code/model/SiteTree.php#L1003). Therefore the test case was failing. I have added a differentiation based on the ParentID in the getClassDropdown method.

Besides it might be a good idea to add a testcase for the fixed behaviour, but I am uncertain where to add this, as it is a crosscuting change in siteconfig as well as in cms.

tractorcow:
For your test case you want a page that has allowed_children = a specific list, and then make sure that getClassnameDropdown() only includes items that match that list.

Before your change, it would ignore the parent allowed_children config. After, it should respect it.

Sound good? :P


```php

class CustomParent extends SiteTree {
	private static $allowed_children = array('FirstClass', 'SecondClass');
}

class Test {
	function testThing() {
		$child = $this->objFromFixture('FirstClass', 'obj1');
		$allowed = $child-> getClassDropdown();
		$this->assertArrayContains('FirstClass', $allowed)
		$this->assertArrayContains('SecondClass', $allowed)
		$this->assertArrayNotContains('ThirdClass', $allowed);
	}
}
```

Super pseduo-coded for succinctness; Please fill out as necessary.

PR #1354, opened 2016-01-05
MINOR: Remove trailing whitespace

sminnee:
Ran this command on all files:

    find . -not \( -name .svn -prune -o -name .git -prune \) -type f | grep -vE .\(gif\|png\|jpg\) | xargs perl -pi -e 's/ +$//'

The main benefit of this is so that authors who make use of
.editorconfig don't end up with whitespace changes in their PRs.

See also: https://github.com/silverstripe/silverstripe-framework/pull/4888


tractorcow:
You need to fix ContentControllerTest to expect the trailing newline (as it was added to the template).

You can also trim the newline from the result as an alternative, to make the test newline agnostic. :)

sminnee:
OK I've broken out the non-whitespace changes into a separate commit, for clarity, and I've opted for ignoring the whitespace rather than allowing for it, as it's not material to the code being tested.

tractorcow:
There's a behat test failure; Can you test it locally to see if it's real?

tractorcow:
image

tractorcow:
Possibly related to the 3 branch changes: https://github.com/silverstripe/silverstripe-framework/issues/4895

sminnee:
It looks like upstream changes have fixed the failure; should be ready for merge now

PR #1355, opened 2016-01-06
Use setFailover() in ContentController (closes silverstripe/silverstripe-framework#4821)

kinglozzer:
Depends on https://github.com/silverstripe/silverstripe-framework/pull/4870.

### DIFF
--- a/code/controllers/ContentController.php
+++ b/code/controllers/ContentController.php
@@ -43,8 +43,10 @@ class ContentController extends Controller {
 		}
 		
 		$this->dataRecord = $dataRecord;
-		$this->failover = $this->dataRecord;
+
 		parent::__construct();
+
+		$this->setFailover($this->dataRecord);
 	}
 	
 	/**


tractorcow:
I've merged the other PR and restarted the tests. Please get back to me when this is ready to merge. :)

kinglozzer:
Ping @tractorcow :)

tractorcow:
Thanks :D

dhensby:
I've just installed 3.x-dev and I'm getting:

```
ERROR [User Error]: Uncaught Exception: Object->__call(): the method 'setfailover' does not exist on 'ContentController'
IN POST /admin/pages/edit/EditForm
Line 740 in /vagrant/www/framework/core/Object.php
```

I think `setFailover` is only defined on `ViewableData` and not `Controller` so this will break - not sure why it's not causing tests to fail though...

This is just plain 3.x-dev install

tractorcow:
I just did a merge up from 3.2 -> 3.3 -> 3... must be related?

tractorcow:
> I think setFailover is only defined on ViewableData and not Controller so this will break - not sure why it's not causing tests to fail though...

But ViewableData is a parent class of Controller, so that can't be the cause. :P

dhensby:
I probably typoed in the composer requirements and took in something bad. it's all ok now!

PR #1360, opened 2016-01-11
FIX: preg_quote() anchors in SiteTreeLinkTracking (fixes #1359)

### LINKED ISSUE #1359: Pasting a full URL in "anchor" causes an internal server error
A client of ours pasted a full URL into the “anchor” field mistakenly, and then tried to save the page. This causes a server error in `SiteTreeLinkTracking::process()` as it tries to parse out the anchor and use it as part of a regular expression. This then left the user unable to do anything with the page in question due to server errors. The link looked like `href="[sitetree_link,id=2]#http://www.mysite.com/page-2/"` (where `page-2` is the URL segment of page ID 2), and the error message was `preg_match(): Unknown modifier 'w'`.

Not sure whether this should be fixed in JavaScript, but either way we should probably code more defensively in `SiteTreeLinkTracking` just in case.

kinglozzer:


dhensby:
test?

kinglozzer:
Sure, I’ll just add an assert that would fail to one of the existing tests in `SiteTreeLinkTrackingTest`

tractorcow:
Nice work @kinglozzer

PR #1361, opened 2016-01-11
Add behat coverage to check that file field is reset when adding links

kinglozzer:
This should fail until https://github.com/silverstripe/silverstripe-framework/pull/4908 is merged


tractorcow:
Merged https://github.com/silverstripe/silverstripe-framework/pull/4908, restarting these tests now.

PR #1367, opened 2016-01-18
Implement RFC-7 JavaScript module loader

flashbackzoo:
Required for https://github.com/silverstripe/silverstripe-framework/pull/4937


chillu:
Checked build failure locally, I think it's a false positive - possibly due to not having the depedant framework PR merged.

Issue #1358, opened 2016-01-08
3.3: Page breadcrumbs inconsistently display "Pages" crumb


purplespider:
Click on a page in the sitetree in 3.3 it will display the title at the top (plus breadcrumbs if a child page):
![](http://spdr.me/P64w+)

However, if you perform an action such as click **Save & Publish**, a "Pages" breadcrumb appears from nowhere:
![](http://spdr.me/Uoye+)

tractorcow:
Looks like a minor bug in the UX revamp. pinging @flashbackzoo  / @scott1702

jonom:
Related: https://github.com/silverstripe/silverstripe-cms/issues/1351

tractorcow:
Fix at https://github.com/silverstripe/silverstripe-cms/pull/1369

kinglozzer:
Closed via dbb8b5852fae0366ddbaa071ab845df88d7c0bdd

PR #1369, opened 2016-01-20
BUG prevent "Home page" being selected when no selection was made

### LINKED ISSUE #1351: /admin/pages breadcrumb displays home page information
When no page is selected (i.e. you're at /admin/pages/) a breadcrumb for the home page is displayed indicating it is selected when it is not. It should just say 'Pages'.

silverstripe_-_pages

tractorcow:
Fixes #1351 and #1358

This looks like old intentional behaviour we don't really want anymore; Selecting "pages" no longer auto-selects a page.

I've also refactored an old update to breadcrumbs which wasn't working with the partial template loading; It should be ok now. ;)

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -869,15 +869,6 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
 		$this->extend('updateCurrentPageID', $id);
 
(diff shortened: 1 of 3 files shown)


kinglozzer:
Checked out and tested, both fixes look good :+1:

tractorcow:
Yay, thanks @kinglozzer

PR #1373, opened 2016-01-25
API VirtualPage permissions now can be set independently of the mirrored pages

### LINKED ISSUE #1318: Virtual Page Permissions
I have recently encountered an issue regarding virtual page permissions (which I consider to be unexpected behaviour), and have confirmed this is still present in **3.2**..

So basically, when you set the **can view** (and most likely the **can edit**) for a virtual page, hitting **save and publish** only writes the permissions for the draft version. I might want to set my virtual page to only be viewable by administrators (maybe from a restricted multisite instance), and although it seems as though this is the case from the CMS, the page will not actually be restricted (the permissions will be determined based on the original page). Perhaps the virtual page is pointing to a page that already has some permissions? The same problem.

I would have expected permissions to be determined at an individual page level, regardless of content, however it does not help that the interface isn't correctly reflecting what has been published. I have found that the following resolved my issues, and I would consider these changes to be fundamental..

```yaml
VirtualPage:
  non_virtual_fields:
    - 'CanViewType'
    - 'CanEditType'
```

tractorcow:
Fixes #1318

Apologies for the late fix, but I couldn't find a way to resolve this that doesn't introduce new api / behaviour, so it has to be targetted for 3.4.

### DIFF
--- a/code/model/VirtualPage.php
+++ b/code/model/VirtualPage.php
@@ -27,6 +27,8 @@ class VirtualPage extends Page {
 		'Version',
 		"Embargo",
 		"Expiry",
+		"CanViewType",
+		"CanEditType",
 	);
 	
 	/**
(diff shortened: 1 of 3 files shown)


dhensby:
I'm not sure we can do this in 3 release. It seems risky to change this behaviour. 

It could end up exposing hidden pages :/

tractorcow:
Hm, that is true. I'll rebase on master.

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-cms/pull/1377

PR #1378, opened 2016-01-27
BUG Use correct formaction for doRollback exemption

tractorcow:
"rollback" was added as an exempt action, but isn't actually an action anywhere. The correct name is "doRollback".


mandrew:
This fix is pretty awesome if I do say myself heh. Tested it with a dropdown field and rather then complaining that there was nothing in the field, it reverted to correct version.

kinglozzer:
@tractorcow Just looking at the list of actions there, do we need to add `archive` too?

dhensby:
I'm going to merge this as it fixes a pretty big bug - @kinglozzer can you either open an issue or PR or we can leave @tractorcow to reply here in his tomorrow morning?

PR #1379, opened 2016-01-27
BUG Check csrf timeout error and preserve unsaved form data

d-rivera-c:
Added a check to getEditForm() to ensure data doesn't get lost if a CSRF error is triggered on Form submission.

When editing any page in the CMS, the SecurityID token becomes invalid after some time. On those ocassions, the unaware user tries to save the data and gets a "Your session has expired. Please re-submit the form." error and the user loses all the modified data.

The [framework saves that data on session](https://github.com/silverstripe/silverstripe-framework/commit/058219c0ee62a7f6313ae5be109956a0e498bed3) when that happens, but CMS doesn't load it back.

Added a fix to check if there's form data present in session, and if only that CSRF error was triggered, and then load the session data into the form so the user can resubmit it.

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -688,6 +688,16 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 				$form->setFields($readonlyFields);
 			}
 
+			// retain stored session information if a CSRF error timeout was triggered
+			$data = Session::get("FormInfo.{$form->FormName()}.data");
+			$message = $form->Message();
+			// check against error, and whether data exists in the session
+			// and override the form data if so.
+			if($message === _t("Form.CSRF_EXPIRED_MESSAGE", "Your session has expired. Please re-submit the form.")
+				&& $data) {
+				$form->loadDataFrom($data);
+			}
+			
 			$this->extend('updateEditForm', $form);
 			return $form;
 		} else if($id) {


tractorcow:
I think the problem needs addressing, but wouldn't the issue remain on other forms? E.g. frontend or other areas outside of the CMS?

The reason that a seemingly "validation error" isn't correctly resetting the form on error, is because the CSRF field doesn't actually do proper form validation. It simply hacks in an exception in Form::httpSubmission() on error.

I would suggest updating SecurityToken::updateFieldSet to instead inject a SecurityTokenField which implements a proper validate() mechanism to catch CSRF errors. That way the issue is resolved for all forms, and allows these errors to be reported by form validators.

dhensby:
@d-rivera-c will you take @tractorcow's feedback and implement it?

d-rivera-c:
@dhensby, I have it in my list of to-do, but so far I haven't had time to implement it. I'll be happy if someone else can do it before I get to it.

tractorcow:
Sorry my suggestions are always hard work. :)

maxime-rainville:
I think this issue has been at least partially address in SS4 with the new "Log back in" dialog.

image

There's also an RFC for an autosave feature ... https://github.com/silverstripe/silverstripe-framework/issues/7877

Unfortunately, [SilverStripe 3 has entered limited support in June 2018](https://www.silverstripe.org/blog/update-on-silverstripe-5-x/). This means we'll only be fixing critical bugs and security issues for SilverStripe 3 going forward.

You can read the [SilverStripe Roadmap](https://www.silverstripe.org/software/roadmap/) for more information on our support commitments.

tractorcow:
That feature also is meant to update all CSRF tokens on all active forms, by the way. :)

```
/**
     * Callback activated by CMSSecurity_success.ss
     */
    reauthenticate(data) {
      // Replace all SecurityID fields with the given value
      if (typeof(data.SecurityID) !== 'undefined') {
        $(':input[name=SecurityID]').val(data.SecurityID);
      }
      // Update TempID for current user
      if (typeof(data.TempID) !== 'undefined') {
        $('body').data('member-tempid', data.TempID);
      }
      this.close();
    },
```

PR #1381, opened 2016-01-29
API Support versioned File management

tractorcow:
API Decouple FIle and ErrorPage
API Refactor versioned API methods out of SiteTree and into Versioned

Requires https://github.com/silverstripe/silverstripe-framework/pull/5001 to be merged first.


code/controllers/ErrorPageFileExtension.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+class ErrorPageFileExtension extends DataExtension {
+
+	/**
+	 * Used by {@see File::getErrorRecordFor}
+	 *
+	 * @param DataObject $record
+	 * @param int $statusCode
+	 */
+	public function updateErrorRecordFor(&$record, $statusCode) {
```
chillu: Could we make this a return value instead of a reference? Looks a bit cleaner.

code/model/SiteTree.php
```diff
@@ -1078,39 +1078,13 @@ public function canEdit($member = null) {
...
-		$extended = $this->extendedCan('canPublish', $member);
-		if($extended !== null) return $extended;
-
-		// Normal case - fail over to canEdit()
-		return $this->canEdit($member);
-	}
-	
 	public function canDeleteFromLive($member = null) {
-		// Standard mechanism for accepting permission changes from extensions
-		$extended = $this->extendedCan('canDeleteFromLive', $member);
```
chillu: Do we need to throw a deprecation for any extensions which have defined `canDeleteFromLive`?

code/model/SiteTree.php
```diff
@@ -1078,39 +1078,13 @@ public function canEdit($member = null) {
...
-		if($extended !==null) return $extended;
-
-		return $this->canPublish($member);
+		Deprecation::notice('4.0', 'Use canUnpublish');
```
chillu: `VirtualPage->canDeleteFromLive()` still exists, and there's a couple of tests invoking it

tractorcow: I'll update. :)

code/model/SiteTree.php
tractorcow: I think we only need to show the deprecation notice if it's actually called. In which case, it'll get shown regardless of whether or not extensions define them.

If we never invoke this method, we can't show the notice otherwise.

code/model/SiteTree.php
tractorcow: Ah, also CMSMain still has many references still. I'll update. :)

tractorcow:
Yikes, I have massively broken link tracking in this rewrite. Don't touch anything until I can fix it. ;)

tractorcow:
Ok, this pull request is ready to review. I've had to do a lot of work on link tracking for the mean time; Instead of trying to fix it fully, I've tentatively disabled live-mode link tracking (it'll still occur, but only on draft).

code/model/SiteTree.php
```diff
@@ -129,21 +128,10 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	);
 
 	private static $many_many = array(
-		"LinkTracking" => "SiteTree",
-		"ImageTracking" => "File",
 		"ViewerGroups" => "Group",
 		"EditorGroups" => "Group",
 	);
 
-	private static $belongs_many_many = array(
```
tractorcow: All of these relationships are in SiteTreeLinkTracking, so I've simply de-duped them.

code/model/SiteTree.php
```diff
@@ -1735,43 +1710,42 @@ public function getLiveURLSegment() {
...
+	 * Rewrites any linked images on this page without creating a new version record.
 	 * Non-image files should be linked via shortcodes
 	 * Triggers the onRenameLinkedAsset action on extensions.
-	 * TODO: This doesn't work for HTMLText fields on other tables.
+	 *
+	 * @todo Implement image shortcodes and remove this feature
 	 */
 	public function rewriteFileLinks() {
+		// Skip live stage
+		if(\Versioned::current_stage() === \Versioned::get_live_stage()) {
```
tractorcow: When we shortcode images we won't need this method at all. ;D

code/model/SiteTreeFileExtension.php
```diff
@@ -88,31 +95,35 @@ public function BackLinkTrackingCount() {
 	}
 	
 	/**
-	 * Updates link tracking.
+	 * Updates link tracking in the current stage.
 	 */
 	public function onAfterDelete() {
+		// Skip live stage
```
tractorcow: Because BackLinkTracking() isn't versioned, we only process back links on stage.

code/model/SiteTreeFileExtension.php
```diff
@@ -121,6 +132,11 @@ public function onAfterDelete() {
 	 * @uses SiteTree->rewriteFileID()
 	 */
 	public function updateLinks() {
+		// Skip live stage
```
tractorcow: Same as above; Don't try to update back links on live.

code/model/SiteTree.php
tractorcow: Fixed.

code/model/SiteTree.php
tractorcow: Fixed in new commit.

code/controllers/ErrorPageFileExtension.php
tractorcow: Yep, it's now a getter.

code/model/SiteTree.php
```diff
@@ -1735,43 +1710,42 @@ public function getLiveURLSegment() {
...
 
-				// Tell static caching to update itself
-				if($table == 'SiteTree_Live') {
-					$publishedClass = $origPublished['ClassName'];
-					$origPublishedObj = new $publishedClass($origPublished);
-					$this->invokeWithExtensions('onRenameLinkedAsset', $origPublishedObj);
-				}
-			}
+			// Update linked assets
+			$this->invokeWithExtensions('onRenameLinkedAsset');
```
chillu: This reverses the semantics of the `onRenameLinkedAsset` though, right? Previously it was *only* called on live, now it's *only* called on stage. And why is the `$origPublishedObj` no longer needed? Might be acceptable given we're not expecting to have this method in 4.0 final at all, just checking

chillu:
Happy with this change. Can we get at least one other +1 from @silverstripe/core-team please?

code/model/SiteTree.php
tractorcow: > $origPublishedObj

Not so much not needed as no longer available. I checked the static publisher module and it doesn't use that parameter anyway.

This method will get nuked in a near update in any case; once all images are shortcoded we won't need to do this anymore.

I expect that static publisher will need to hook into another extension point to publish pages; onAfterPublish etc.

tractorcow:
+1 ? :P Or do I not count? :P haha

kinglozzer:
Had a look yesterday, forgot to comment on this one as well as the framework one: +1 from me

tractorcow:
Rebased to keep it all current. :)

Tests of course won't pass until the framework PR is merged.

PR #1382, opened 2016-01-30
SiteTree can* methods flow updated.

manja:
First call extensions, then default to admin rights.


tractorcow:
I think this change is ok, but it will break sites that rely on admins having full access (regardless of extended permissions) so it'll need to go into master.

Can you please rebase, and in addition add some upgrading notes to the 4.0.0.md changelog about the change? It'll need a separate PR to framework for the notes, but we can merge them together.

dhensby:
Did this ever get opened on master?

I think this will break a lot of existing modules that expect that `ADMIN`s will not get to their handler... This will also have a performance impact :/

dhensby:
OK - I tried to merge this by hand and it looks like this fix has already made it's way into core with #1509

PR #1385, opened 2016-02-02
Fix typo

AppChecker:
Varaible $toVersion is checked twice, but $fromVersion is not checked.

This possible defect found by AppChecker


tractorcow:
Nice fix, thanks @AppChecker . you can check my apps any time!

PR #1392, opened 2016-02-09
BUG Prevent folders deleted on the filesystem from breaking asset interface

tractorcow:
If a folder exists in the DB but not the filesystem, then sometimes editing that in the admin interface will try to show fields from both child-listview and edit-fields in the same interface, although it will not work properly.

Before:

image

After:

image


tractorcow:
Any @silverstripe/core-team want to help me get this over the line? This will go into 3.3.0-rc2

tractorcow:
Thanks @chillu

PR #1400, opened 2016-02-24
BUG Fix behat tests for asset abstraction

tractorcow:
Requires https://github.com/silverstripe-labs/silverstripe-behat-extension/pull/111

Also corrects reference from "Delete" to "Archive" in asset admin, since assets are now versioned.


tractorcow:
Ping @chillu

tractorcow:
Thanks. :D

PR #1401, opened 2016-02-25
API Mark image tracking as owned

tractorcow:
CMS Component of ownership api.

See https://github.com/silverstripe/silverstripe-framework/pull/5095


tractorcow:
Additional regression fixed with https://github.com/silverstripe/silverstripe-framework/issues/5143 :D

PR #1404, opened 2016-02-29
"npm run lock" command

chillu:
Uses https://github.com/uber/npm-shrinkwrap instead of the built-in "npm shrinkwrap" since it works more reliably.
Specifically, "npm install" doesn't fail depending on node_modules/ being installed in the local cache or not.
It also makes npm-shrinkwrap.json easier to diff by more consistently ordering its output between runs.

If you need any convincing that this is a problem, look at the over 400 issues related to "shrinkwrap"
in https://github.com/npm/npm/search?q=shrinkwrap&type=Issues&utf8=%E2%9C%93

Requires https://github.com/silverstripe/silverstripe-framework/pull/5106

/cc @flashbackzoo

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,7488 @@
+{
+  "name": "silverstripe-cms",
+  "version": "4.0.0",
+  "npm-shrinkwrap-version": "5.4.1",
+  "node-version": "v4.2.3",
+  "dependencies": {
+    "babel-core": {
+      "version": "6.5.2",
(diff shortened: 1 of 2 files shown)


flashbackzoo:
Nice. I see you remembered to shrinkwrap npm-shrinkwrap too :smile:

chillu:
Builds will pass once https://github.com/silverstripe/silverstripe-framework/pull/5087 is resolved

dhensby:
Still not passing @chillu

chillu:
Committed through https://github.com/silverstripe/silverstripe-cms/commit/ee3158183f66115467b00977f1bc99c8587f8547

PR #1411, opened 2016-03-03
Removed dependency for translatable module

micmania1:
To go with https://github.com/silverstripe/silverstripe-translatable/pull/231

### DIFF
--- a/code/controllers/CMSPageAddController.php
+++ b/code/controllers/CMSPageAddController.php
@@ -91,7 +91,7 @@ class CMSPageAddController extends CMSPageEditController {
 		// TODO Re-enable search once it allows for HTML title display,
 		// see http://open.silverstripe.org/ticket/7455
 		// $parentField->setShowSearch(true);
-		
+
 		$parentModeField->addExtraClass('parent-mode');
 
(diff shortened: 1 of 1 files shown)


code/controllers/CMSPageAddController.php
```diff
@@ -148,9 +148,7 @@ public function doAdd($data, $form) {
 		}
 
 		$record = $this->getNewItem("new-$className-$parentID".$suffix, false);
-		if(class_exists('Translatable') && $record->hasExtension('Translatable') && isset($data['Locale'])) {
-			$record->Locale = $data['Locale'];
-		}
+		$this->extend('updateDoAdd', $record, $data);
```
dhensby: passing the form (and not data) would be better

dhensby:
happy for merge when green ticks

dhensby:
Why is behat failing :(

micmania1:
Can you restart the tests? They don't seem related and I have has random test failures in the past.

dhensby:
restarted

tractorcow:
Restarted again... cms tests have been weird lately. Will investigate soonish.

PR #1415, opened 2016-03-08
API Update uses of versioned

tractorcow:
To be merged after https://github.com/silverstripe/silverstripe-framework/pull/5157

The CMS component of this change actually includes a lot of refactoring; Virtual pages were mostly refactored due to the fundamental changes in the way that related objects must be published under owned versioning.

I have no idea why so many trimmed whitespace ended up in the code... we flushed these ages ago, but I guess they made their way back into the code after merging up from 3 -> master?

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -132,6 +132,14 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 		"EditorGroups" => "Group",
 	);
 
+	private static $has_many = array(
+		"VirtualPages" => "VirtualPage.CopyContentFrom"
+	);
+
+	private static $owned_by = array(
+		"VirtualPages"
+	);
+
 	private static $casting = array(
 		"Breadcrumbs" => "HTMLText",
 		"LastEdited" => "SS_Datetime",
@@ -181,7 +189,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 
 	private static $extensions = array(
 		"Hierarchy",
-		"Versioned('Stage', 'Live')",
+		"Versioned",
 		"SiteTreeLinkTracking"
 	);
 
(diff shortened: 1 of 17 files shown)


kinglozzer:
Reviewed this too, looks great - more features with less code = win win! Same question about deprecation notices as the framework PR (`VirtualPage::copyFrom()` for example), but other than that, +1.

tractorcow:
Rebased on the whitespace cleanup. ;) much better!

code/model/SiteTree.php
```diff
@@ -1814,26 +1802,14 @@ public function DependentPages($includeVirtuals = true) {
...
-		}
-
-		// Check existing virtualpages
-		if(class_exists('VirtualPage')) {
-			return VirtualPage::get()->where(array(
-				'"VirtualPage"."CopyContentFromID"' => $this->ID
-			));
+		// Disable subsite filter for these pages
+		if($pages instanceof DataList) {
+			return $pages->setDataQueryParam('Subsite.filter', false);
```
tractorcow: Todo: A better way for extensions to update RelationList components?

code/model/SiteTreeLinkTracking.php
```diff
@@ -28,12 +28,39 @@
...
-		'parser' => '%$SiteTreeLinkTracking_Parser'
+		'Parser' => '%$SiteTreeLinkTracking_Parser'
 	);
 
+	/**
+	 * Parser for link tracking
+	 *
+	 * @return SiteTreeLinkTracking_Parser
+	 */
+	public function getParser() {
```
tractorcow: This was a work-around for virtual pages trying to virtualise the parser; Having an explicit getter / setter prevented this.

tractorcow:
It's hard to prove that the tests pass without having framework merged, so here's the output from my local testing.

image

code/model/SiteTreeFileExtension.php
```diff
@@ -109,7 +109,7 @@ public function BackLinkTrackingCount() {
 	 */
 	public function onAfterDelete() {
 		// Skip live stage
-		if(\Versioned::current_stage() === \Versioned::get_live_stage()) {
+		if(\Versioned::get_stage() === Versioned::LIVE_STAGE) {
```
mikenz: Shouldn't that be `\Versioned::LIVE_STAGE` on the right?

tractorcow: Turns out it doesn't matter, but yes it SHOULD be consistent.

I changed IDEs since startinging this project haha. :D

hafriedlander:
Behat tests fail

hafriedlander:
It's an issue with behat extension not supporting new Version syntax, but will need fixing, since all behat tests will be broken from now on.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/5074

PR #1416, opened 2016-03-08
FIX: Incorrect title attribute on CMS tabs

kinglozzer:


tractorcow:
Sorry, it turns out that weird "title" attribute was necessary for tabs to work.

Reverted with https://github.com/silverstripe/silverstripe-cms/pull/1425.

PR #1417, opened 2016-03-08
Cleanup trailing whitespace

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/5160

I've only touched php files, and not any thirdparty files.


tractorcow:
Closed with https://github.com/silverstripe/silverstripe-cms/commit/849cd898a4ecd2fb535a3655f91f195386b53f73

PR #1418, opened 2016-03-09
FIX 'Settings' fields being overwritten by 'Content' fields

mikenz:
CMSMain::getEditForm wasn't passing the $fields parameter to the parent. The the "Settings" fields we being replaced with the "Content" settings when requesting /admin/pages/settings/show/123345


tractorcow:
Thank you very much! I'm sorry I didn't notice this PR earlier.

PR #1422, opened 2016-03-10
FIX Use `Controller::join_links()` in Reports

### LINKED ISSUE #1420: Recently Edited Report contains broken links when used with Translatable module
When the Translatable module is installed, the Recently Edited Report generates invalid links to pages it lists. The format it uses is /admin/pages/edit/show?locale=en_US/1, rather than /admin/pages/edit/show/1?locale=en_US

This sounds like a very similar issue to #1183

dhensby:
Fixes #1420 maybe?


dhensby:
@tractorcow will this need to be applied to the reports module?

tractorcow:
Yes; Can you raise there separately? Otherwise I have to merge up from cms -> reports module.

dhensby:
Ah, this report is actually now in CMS :P

PR #1425, opened 2016-03-13
Revert "FIX: Incorrect title attribute on CMS tabs"

### LINKED ISSUE #1421: CMS Tabs are off on `3.1.x-dev`
I updated composer while on `3.1.x-dev`, and this happened to me when clicking on random pages. Now I rolled back to `3.1.18`, and it disappeared.

screenshot 2016-03-09 19 05 52

tractorcow:
Reverts #1416

Fixes #1421

### DIFF
--- a/templates/Includes/CMSMain_Content.ss
+++ b/templates/Includes/CMSMain_Content.ss
@@ -8,18 +8,18 @@
 		<div class="cms-content-header-tabs">
 			<ul class="cms-tabset-nav-primary">
 				<li class="content-treeview<% if class == 'CMSPageEditController' %> ui-tabs-active<% end_if %>">
-					<a href="$LinkPageEdit" class="cms-panel-link" title="<%t CMSMain.TabContent 'Content' %>" data-href="$LinkPageEdit">
-						<%t CMSMain.TabContent 'Content' %>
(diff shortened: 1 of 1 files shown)


dhensby:
Erm... lol? Is the CSS targeting title attributes? Can't we fix the CSS rather than having bad markup?

tractorcow:
I did a search and I have no idea which script was reading those title properties, so I just reverted it.

kinglozzer:
> I did a search and I have no idea which script was reading those title properties, so I just reverted it.

I did exactly the same before merging and couldn’t find it, but I’ve since tracked it down to here: https://github.com/silverstripe/silverstripe-framework/blob/3.1/thirdparty/jquery-ui/jquery-ui.js#L14240-L14253.

That entire block is marked as deprecated, and is [only enabled if `$.uiBackCompat` isn’t false](https://github.com/silverstripe/silverstripe-framework/blob/3.1/thirdparty/jquery-ui/jquery-ui.js#L13984). We never actually explicitly set it to true, but `$.uiBackCompat == undefined` so it passes that check anyway. Easiest, and probably lowest-risk, fix would be to just delete that “title” tabs block. I don’t think we’d want to risk disabling jQuery UI bc-mode (even if we never intentionally enabled it :wink:)

dhensby:
oh gawd - or just leave it for now ;)

PR #1426, opened 2016-03-14
Added readonly version of SiteTreeURLSegmentField

SilbinaryWolf:
Added readonly version of SiteTreeURLSegmentField, small quality of life thing when a user is looking at readonly data and wants to easily view the frontend of that page.


dhensby:
Looks like a good addition to me.

dhensby:
Thanks

PR #1429, opened 2016-03-15
Remove redundant CSS

jonom:
StripThumbnail now provides icon for all file types. To be merged after https://github.com/silverstripe/silverstripe-framework/pull/5186, if accepted


jonom:
@dhensby could you review/merge this too? There's a failed test here but don't see how it could be due to this - all that has been changed is that a background image has been removed.

The background images are no longer necessary in asset admin gridfield because StripThumbnail is providing an image or icon for all files now.

jonom:
I'll re-run the tests now that related changes have been merged to Framework in case it makes a difference

jonom:
Green. That's better.

PR #1436, opened 2016-04-04
Add license file

camfindlay:
While this is in the README it's common to have a separate license file to make it clear this is FOSS.


tractorcow:
The license was in the readme file; If you are updating it, the text between readme and license.md should match. Can you please fix that?

PR #1445, opened 2016-04-04
Fix IE support by using babelify transforms

chillu:
See https://github.com/silverstripe/silverstripe-asset-admin/pull/103

We're not using a whole lot of ES6 in the CMS yet, but this is syncing up build tooling with `framework` and `cms`

Note that spacing in the gulpfile is a bit all over the show, this will be addressed at a larger scale by introducing ESLint into the codebase.


tractorcow:
I've attempted to rebase and force push this up... still waiting for the changes to appear on Github.

tractorcow:
I'll just restart the build for now.

Issue #1447, opened 2016-04-07
ErrorPage not properly writing static files 


patricknelson:
Currently `ErrorPage->writeStaticPage()` ([here](https://github.com/silverstripe/silverstripe-cms/blob/3/code/model/ErrorPage.php#L253)) is not properly checking the return value of `file_put_contents()`. It is performing loose type checking instead of ensuring the result is actually `false` ([see documentation](http://php.net/manual/en/function.file-put-contents.php#refsect1-function.file-put-contents-returnvalues)).  When an error page, for some reason or another, doesn't contain a body and is zero bytes, `file_put_contents()` will dutifully write zero bytes and return the integer `0`, as it should.

So:
```php
// ...
$filePath = self::get_filepath_for_errorcode($this->ErrorCode, $this->Locale);
if (!file_put_contents($filePath, $errorContent)) {
    // ...
}
```

Should instead be:

```php
// ...
$filePath = self::get_filepath_for_errorcode($this->ErrorCode, $this->Locale);
if (file_put_contents($filePath, $errorContent) === false) {
    // ...
}
```

I think the fact that an error page has returned a `$response->getBody() == ''` result is a different issue and, as such, the error checking that this method *may* need to be in a different discussion. However the core point here is that it is causing a crash with an irrelevant error message in my particular edge case when, instead, it should simply be continuing as normal or at least producing an accurate error message (if we decide a zero length body is indeed an error in its own right).

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-cms/pull/1448

PR #1448, opened 2016-04-07
FIX for #1447 to ensure proper type-safe checking of file_put_contents() in ErrorPage->writeStaticPage()

patricknelson:
Addresses #1447 by simply performing a type-safe check instead of using loose type handling in `if (...)`. Ensuring error message is relevant to the error that actually occurs.


tractorcow:
Thanks @patricknelson !

patricknelson:
:+1:

PR #1456, opened 2016-04-19
Adjust file paths for new static file structure in framework

chillu:
Depends on https://github.com/silverstripe/silverstripe-framework/pull/5346


chillu:
Merging on behalf of, and with the approval of @tractorcow - see https://github.com/silverstripe/silverstripe-framework/pull/5346

PR #1457, opened 2016-04-21
BUG Correct framework/module dependencies for cms

tractorcow:
Fixes https://github.com/silverstripe-labs/silverstripe-fulltextsearch/pull/117

Discovered when testing an install of 3.3 was bringing in 3 branch (alias 3.4) of cms. That cms branch relied on features not present in the 3.3 branch of framework, and busted everything.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,9 +18,9 @@
 	"require": {
 		"php": ">=5.3.3",
 		"composer/installers": "*",
-		"silverstripe/framework": "~3.3",
-		"silverstripe/reports": "~3.3",
-		"silverstripe/siteconfig": "~3.3"
+		"silverstripe/framework": "~3.4",
+		"silverstripe/reports": "~3.4",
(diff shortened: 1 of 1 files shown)


sminnee:
Why are we not using self.version for this?

sminnee:
Is it necessary/appropriate to bump the requirements for reports and siteconfig?

tractorcow:
We had issues with self.version during releases; It doesn't let you use dev branches with the equivalent stable tag.

tractorcow:
> Is it necessary/appropriate to bump the requirements for reports and siteconfig?

If we added new features for siteconfig in 3.4, installing cms 3.4 that relied on those features could install 3.3 and have the same issues we have now with fulltextsearch.

tractorcow:
Test failures fixed by https://github.com/silverstripe/silverstripe-framework/pull/5370

PR #1458, opened 2016-04-21
API: reports no longer mandatory.

sminnee:
The reports and siteconfig modules should be included by our default
installer they aren't a requirement of the CMS. They were included as
requirements in 3.x to retain backwards-compatability. In master, they
can be removed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,8 +19,7 @@
 		"php": ">=5.5.0",
 		"composer/installers": "*",
 		"silverstripe/framework": "^4.0",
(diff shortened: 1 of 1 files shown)


sminnee:
Heh, oops, turns out siteconfig *is* a requirement.

tractorcow:
Nice idea. +1

sminnee:
Lol, okay, reports is a requirement too.

NEVER MIND.... 8-)

PR #1459, opened 2016-04-21
Add 'Add to campaign' button to pages

hafriedlander:
Dependant on https://github.com/silverstripe/silverstripe-framework/pull/5368

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -1191,6 +1191,11 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 		return $this->getResponseNegotiator()->respond($this->getRequest());
 	}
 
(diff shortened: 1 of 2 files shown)


tractorcow:
Can you please phpdoc the new method?

code/controllers/CMSMain.php
```diff
@@ -1191,6 +1191,11 @@ public function doRollback($data, $form) {
 		return $this->getResponseNegotiator()->respond($this->getRequest());
 	}
 
+	public function addtocampaign($data, $form) {
+		$handler = new AddToCampaignHandler($form, $data);
```
tractorcow: Shall we use ::create here as we do below?

tractorcow: `use Injectable` in AddToCampaignHandler.php :D

PR #1460, opened 2016-04-21
Fix font-icon regression on tree view

flashbackzoo:
Before:
icons-cms-before

After:
fix-icons-cms

Caused by https://github.com/silverstripe/silverstripe-framework/commit/35e062d9f32df475eb41c78bd709059b417b8b01 updating a CSS class name.


flashbackzoo:
Related to https://github.com/silverstripe/silverstripe-framework/issues/5367

PR #1461, opened 2016-04-21
Adapted JS lib references to renamed paths

chillu:
Dependant on https://github.com/silverstripe/silverstripe-framework/pull/5372


chillu:
Looked this over with @tractorcow and @flashbackzoo - they've both done a partial peer review. Merging now, since it's starting to inhibit our workflow around other pull requests.

PR #1475, opened 2016-05-05
MINOR Correcting the default HTML so TinyMCE doesn't auto-clean it up…

### LINKED ISSUE #1473: TinyMCE HTML cleanup modifes default content
When creating a new blank install, the default records for the "About Us" and "Contact Us" pages contain HTML that TinyMCE cleans up immediately. This clean-up means that when a user navigates to the page, they get prompted to save the modified page. A page the user is not aware that they modified. #

candidasa:
…on first viewing of a new database (fixes #1473)


chillu:
Thanks Julian! Translations normally need to be modified on transifex directly (https://docs.silverstripe.org/en/3.3/contributing/translations/), otherwise they'll get overwritten on the next `tx pull` call. While pushing translations is possible in theory, it involves a few steps which we don't have time for now (`tx push` and `tx pull` in a certain order with some sanity checks). So technically, this PR only fixes the issue in the `en` default language.

Issue #1480, opened 2016-05-06
Add page markup is different for steps


clarkepaul:
Because the markup is different the CSS looks different for each step which breaks the layout in some browsers.
![pasted_image_6_05_16__1_51_pm](https://cloud.githubusercontent.com/assets/555033/15062172/a46ae5e2-1391-11e6-95a9-3d2de23f4929.png)

![pasted_image_6_05_16__1_49_pm](https://cloud.githubusercontent.com/assets/555033/15062159/75cb71de-1391-11e6-8e07-066d81d98146.png)

PR #1484, opened 2016-05-06
Make page steps use the same styles, consistent markup

clarkepaul:
This will require style changes within framework which I will link to shortly.

Before:
pasted_image_6_05_16__1_25_pm

After:
pasted_image_6_05_16__3_21_pm


clarkepaul:
Partial fix for #1480

PR #1488, opened 2016-05-09
FIX add_i18n_javascript calls not being updated after JS move

hafriedlander:
This fixes alert / confirm boxes that were popping up without text (for
example silverstripe-cms/issues/1476), although ideally we wouldn't
show empty dialog boxes on this sort of error - we'd have some default,
or a way to detect the issue.


hafriedlander:
See also https://github.com/silverstripe/silverstripe-framework/pull/5475

PR #1489, opened 2016-05-09
FIX Not being able to save when viewing page settings

hafriedlander:


chillu:
I'm not entirely sure why this was broken in the first place, my best guess is https://github.com/silverstripe/silverstripe-cms/commit/70062ebc9cef301f2742ee08d2fade1bb7a41866. It complies to the method arguments (which mention `int`), and fixes the issue at hand, so merging.

PR #1490, opened 2016-05-09
Display the name of a deleted members in the page history

mikenz:
Requires https://github.com/silverstripe/silverstripe-framework/pull/5482

When a page has been edited/published by a Member that has since been deleted currently we display "Unknown" as the editor or publisher making it difficult to work out what changes were previously made.

This change relies on a change in framework that creates a DeletedMember record when a Member is deleted. If the Author or Publisher ID for a version doesn't match and existing member it checks for a deleted member and displays their name. DeletedMember automatically prefixes the name with "(deleted) ".

### DIFF
--- a/code/controllers/CMSPageHistoryController.php
+++ b/code/controllers/CMSPageHistoryController.php
@@ -209,6 +209,13 @@ class CMSPageHistoryController extends CMSMain {
 				foreach($versions as $k => $version) {
 					$active = false;
 
+                    if ($version->AuthorID && !$version->Author) {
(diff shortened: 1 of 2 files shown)


sminnee:
Once https://github.com/silverstripe/silverstripe-framework/issues/5603 is addressed we should revisit this.

tractorcow:
I think given that https://github.com/silverstripe/silverstripe-framework/pull/5482 has been closed, this is going to need to be re-addressed once we agree / implement a solution. Let's likewise close this until that time.

PR #1492, opened 2016-05-10
Linting and CodeClimate support

chillu:
Not doing a whole lot at the moment since all JS and SCSS is in ignored legacy/ folders.
But it's just as important to ensure we write new code properly once the pull requests coming in.


chillu:
Set up on CodeClimate - obviously won't build much until we merge the `.codeclimate.yml` here: https://codeclimate.com/github/silverstripe/silverstripe-cms/

PR #1498, opened 2016-05-17
DBField namespace in Search for php7 support

WindSaber:
I think this class was forgotten when DBField namespaces were added to code


tractorcow:
Thanks @WindSaber :D

PR #1500, opened 2016-05-19
Removed branch alias which should only apply to latest release branch

chillu:
See https://github.com/silverstripe/silverstripe-framework/commit/070ae2555b4b6beddec71cd2b9e54057a03e872d
for similar commit when going from 3.1 to 3.2.

See https://github.com/silverstripe/silverstripe-lumberjack/pull/40 for context.

Merge with https://github.com/silverstripe/silverstripe-framework/pull/5561

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,11 +22,6 @@
 		"silverstripe/reports": "~3.3",
 		"silverstripe/siteconfig": "~3.3"
 	},
-	"extra": {
-		"branch-alias": {
-			"3.x-dev": "3.3.x-dev"
-		}
-	},
 	"autoload": {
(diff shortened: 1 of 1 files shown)


tractorcow:
These aliases are actually ignored if not on the branch they declare aliases for... it doesn't hurt to remove them though, but I'm not sure it will fix anything.

dhensby:
@tractorcow wouldn't it cause confusion about whether 3.3.x-dev was branch 3 or 3.3 ?

chillu:
@tractorcow The reason I've removed them is because they *weren't* ignored in composer's dependency resolution, see the lumberjack build failure referenced in the ticket.

PR #1501, opened 2016-05-19
SiteTree configurable css classes

axyr:
Same as https://github.com/silverstripe/silverstripe-cms/pull/1325
Red the thread.

- Yes, maybe a framework should not take care of this.
- But, silverstripe sitetree is a kind of its own and that is very good.
- So, we use css frameworks that have other active states for css classes.
- And, we don't want if/else in our template classes, nor override these methods in Page.php just for .active
- Now, we are going from 3 to 4. this might be a good moment to fix this...

Let me know if you want me to check documentation references for checks like <% if $LinkingMode = 'section' %>

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -105,6 +105,27 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	 * @var string
 	 */
 	private static $hide_ancestor = null;
+	
+	/**
+	 * @see \SiteTree::LinkingMode, \SiteTree::LinkOrSection, \SiteTree::LinkOrCurrent
+	 * @config
+	 * @var string
+	 */
+	private static $link_css_class = 'link';
+
+	/**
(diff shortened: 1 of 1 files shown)


dhensby:
> Let me know if you want me to check documentation references for checks like <% if $LinkingMode = 'section' %>

Please do.

I think this change is acceptable, but I know it's been raised before and had less enthusiasm...

tractorcow:
I've seen this PR before, and the issue I've had was with non-configurable files (such as physical CSS files) referencing 'link', 'current', etc... and do NOT support a configurable class approach.

My response to this is the same as it was then; I disagree with configurable elements, where every config option other than the default will result in an inconsistent or broken CMS user experience.

If you wished, however, to have a configurable EXTRA class added (in addition to the core classes), and have these default to empty, then that would be fine.

dhensby:
Source repo is dead, so this will require manual intervention and @tractorcow aint having any of this

PR #1507, opened 2016-05-23
API Allow extensions to influence can* methods

tractorcow:
This is necessary in cases where extensions apply specific business logic. There is currently no way to make this logic apply to admins.

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -174,7 +174,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	 * @var string
 	 */
 	private static $icon = null;
-	
+
 	/**
 	 * @config
 	 * @var string Description of the class functionality, typically shown to a user
@@ -187,7 +187,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 		"Versioned('Stage', 'Live')",
 		"SiteTreeLinkTracking"
 	);
-	
+
 	private static $searchable_fields = array(
 		'Title',
 		'Content',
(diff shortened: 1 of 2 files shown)


dhensby:
Looks good. What are the risks?

code/model/SiteTree.php
```diff
@@ -1082,38 +1107,45 @@ public function canEdit($member = null) {
 	 * @return bool True if the current user can publish this page.
 	 */
 	public function canPublish($member = null) {
-		if(!$member || !(is_a($member, 'Member')) || is_numeric($member)) $member = Member::currentUser();
-		
-		if($member && Permission::checkMember($member, "ADMIN")) return true;
+		if(!$member || !(is_a($member, 'Member')) || is_numeric($member)) {
+			$member = Member::currentUser();
```
stevie-mayhew: Indentation.

stevie-mayhew:
Looks good apart from one minor indentation issue. Great change.

sminnee:
Any chance we could do this in 3.4.0-rc2?

chillu:
Seems like a good change, but it changes assumptions which project code might make. Can we add the following to the upgrading guide?

`SiteTree` permission methods check extensions before enforcing their own permissions. If any extension returns a non-null value (`true` or `false`), this overrules the built in `SiteTree` checks. This affects `canCreate`, `canEdit`, `canDelete`, `canPublish`, `canAddChildren` and `canView`.

+1 on Sam's suggestion for a 3.4.0-rc2 - it seems the internal release workflow constraints (audit) here are resolveable?

dhensby:
RCs should be pretty final in terms of APIs and this strikes me as being a bit outside what RC process should be...

However, I do think this is an important fix that should be released sooner rather than later and we can either be silly and just kill the 3.4 line now and release a 3.5 and be pedantic or we could slip it into the RC... So let's get it into the RC but just try to be a bit more careful going forward.

tractorcow:
After discussion with @hafriedlander , there is a risk of regression if merging this, so it will need to go into the master branch.

code/model/SiteTree.php
tractorcow: There's nothing wrong with this indentation, is there?

tractorcow:
> RCs should be pretty final in terms of APIs and this strikes me as being a bit outside what RC process should be...

Sorry, I have to agree with @dhensby on this. ;D

code/model/SiteTree.php
stevie-mayhew: https://github.com/silverstripe/silverstripe-cms/pull/1507/files

Check it in that view - weird.. Will download everything from now on :D

tractorcow:
Moved to https://github.com/silverstripe/silverstripe-cms/pull/1509

dhensby:
see #1382

PR #1508, opened 2016-05-23
BUG Fix rollback page crash

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/5532


dhensby:
Behat failed... Restarting just in case.

tractorcow:
Behat passes. ;D

PR #1520, opened 2016-06-03
API Respect default_cast

tractorcow:
API Revert hack in VirtualPage
API Respect default_cast

To go with https://github.com/silverstripe/silverstripe-framework/pull/5653


dhensby:
Right, there are behat test failures, but I'm merging this because the framework one has been and I think that this will mean there are LESS errors if I merge.

PR #1521, opened 2016-06-09
FIX if a language is set in subsites make sure this is used in

ohararyan:
See issue here https://github.com/silverstripe/silverstripe-subsites/issues/83

### DIFF
--- a/code/controllers/ContentController.php
+++ b/code/controllers/ContentController.php
@@ -12,7 +12,7 @@
  *
  * Subclasses of ContentController are generally instantiated by ModelAsController; this will create
  * a controller based on the URLSegment action variable, by looking in the SiteTree table.
- * 
+ *
  * @todo Can this be used for anything other than SiteTree controllers?
  *
  * @package cms
@@ -29,7 +29,7 @@ class ContentController extends Controller {
 		'deleteinstallfiles', // secured through custom code
 		'LoginForm'
 	);
-	
+
 	/**
 	 * The ContentController will take the URLSegment parameter from the URL and use that to look
 	 * up a SiteTree record.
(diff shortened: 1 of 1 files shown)


wilr:
Ideally this would be refactored so that translatable, subsites and module X could all hook in. Also test failing so will need to investigate that

code/controllers/ContentController.php
```diff
@@ -363,21 +363,25 @@ public function SiteConfig() {
...
 	 */
 	public function ContentLocale() {
 		if($this->dataRecord && $this->dataRecord->hasExtension('Translatable')) {
 			$locale = $this->dataRecord->Locale;
 		} elseif(class_exists('Translatable') && SiteTree::has_extension('Translatable')) {
 			$locale = Translatable::get_current_locale();
+		} elseif(class_exists('Subsite') && $subsite = Subsite::currentSubsite()) {
+			$locale = i18n::get_locale_from_lang($subsite->Language);
+			if($locale) {
+				i18n::set_locale($locale);
```
tractorcow: Not very good style; A getter should not by nature modify the state.

tractorcow:
If you want to allow extensions (such as subsites) set the locale, then the correct way to do it is via an extension.


```
public function ContentLocale(){
	$locale = ''; // some logic
	$this->extend('updateContentLocale', $locale);
	return i18n::convert_rfc1766($locale);
}
```

This means that we have less module-specific code, and other modules can also hook into this behaviour.

This would need to go into the 3 branch please. :D

code/controllers/ContentController.php
```diff
@@ -363,21 +363,25 @@ public function SiteConfig() {
...
-	 * 
+	 *
 	 * Suitable for insertion into lang= and xml:lang=
 	 * attributes in HTML or XHTML output.
-	 * 
+	 *
 	 * @return string
 	 */
 	public function ContentLocale() {
 		if($this->dataRecord && $this->dataRecord->hasExtension('Translatable')) {
```
tractorcow: Changing this to `$this->dataRecord->hasField('Locale')` would mean being able to remove the hard-coded translatable dependency too.

code/controllers/ContentController.php
```diff
@@ -363,21 +363,25 @@ public function SiteConfig() {
...
 			$locale = $this->dataRecord->Locale;
 		} elseif(class_exists('Translatable') && SiteTree::has_extension('Translatable')) {
```
tractorcow: To delete; Made redundant by the above check. :P

PR #1522, opened 2016-06-09
FIX reading mode can change during sitetree generation

micmania1:
This causes a bug where the site tree incorrectly displays the structure (pages seem to be missing).

To reproduce the bug:
1. Create 3 pages (Parent page > Child page > Grandchild Page) and publish all 3
2. Move grandchild page to be under Parent page (same level as child page)
3. Click on child page
4. Unpublish child page
5. Delete child page
6. Refresh the page (still on the deleted child page)
(Notice that it doesn't show other children of Parent page)
7. Click onto another page (ie. Parent page) and refresh
(Notice that child pages display correctly)

This doesn't actually delete the page, but the site tree renders incorrectly.

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -4,29 +4,29 @@
  *
  * This class creates a 2-frame layout - left-tree and right-form - to sit beneath the main
  * admin menu.
- * 
+ *
  * @package cms
  * @subpackage controller
  * @todo Create some base classes to contain the generic functionality that will be replicated.
  */
 class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionProvider {
-	
+
 	private static $url_segment = 'pages';
-	
+
 	private static $url_rule = '/$Action/$ID/$OtherID';
-	
+
 	// Maintain a lower priority than other administration sections
 	// so that Director does not think they are actions of CMSMain
 	private static $url_priority = 39;
-	
+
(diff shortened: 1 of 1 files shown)


code/controllers/CMSMain.php
```diff
@@ -545,19 +546,22 @@ public function getRecord($id, $versionID = null) {
...
 			 *	This bit breaks the all FileIFrameField functions if the field is used in CMS
-			 *  and its relevent ajax calles, like loading the tree dropdown for TreeSelectorField. 
+			 *  and its relevent ajax calles, like loading the tree dropdown for TreeSelectorField.
 			 */
 			/* if($record && SiteTree::has_extension('Translatable') && $record->Locale && $record->Locale != Translatable::get_current_locale()) {
 				$record = null;
 			}*/
 
+			// Set the reading mode back to what it was.
+			Versioned::reading_stage($currentStage);
```
micmania1: FYI: See line 531/533

micmania1:
Test failures don't look related (they're currently failing on 3.1 before this commit).

tractorcow:
Can you please use get/set_reading_mode instead of reading_stage? If stage is null (as is the case) reading_stage(null) will break.

tractorcow:
@micmania1 any update on this? :P

dhensby:
merged into 3.2 with 5f2d8cae0d3da9f4bef13d7240aed9e07cae1099

PR #1524, opened 2016-06-10
Fixed Behat admin/old-assets url

chillu:
To be moved to the asset-admin module, but we should keep the build
passing until we actually move it!


tractorcow:
I want to merge, but tests are failing.

sminnee:
@chillu can you please comment on the relevance of this?

chillu:
No longer relevant

Issue #1253, opened 2015-07-31
SiteTree duplicate doesn't match the DataObject duplicate, so onAfterDuplication is very tricky


pitchandtone:
SiteTree calls $this->invokeWithExtensions('onAfterDuplicate', $clone) whereas DataObject calls $clone->invokeWithExtensions('onAfterDuplicate', $this)

I personally think the latter is better.

sb-relaxt-at:
This issue still exists in 3.2.1. The only usage of the on*Duplicate extension points in the core - [Versioned](https://github.com/silverstripe/silverstripe-framework/blob/3/model/Versioned.php#L1508) - is indicating that the second option is the intended one.

tractorcow:
Yep, but I think it is possibly risky to do this in 3, in case users are relying on the current sitetree behaviour. Do you want to raise a PR to master branch to change this API?

tractorcow:
SiteTree::duplicate also calls parent::duplicate, so these extensions are in fact run TWICE, once for both clone and original, ugh.

sb-relaxt-at:
Yes the duplicated (no pun intended ;-)) call of on*Duplicate is also an issue... Yes I can try to prepare a PR. Concerning the extension methods, it might be the best to create an onBeforeDuplicate and onAfterDuplicate method in DataObject, similar to the on*Write methods? This would allow clean overwriting by SiteTree and prevent duplicate extension calls.

tractorcow:
> it might be the best to create an onBeforeDuplicate and onAfterDuplicate method in DataObject, similar to the onWrite methods?

I'm not sure if that would solve anything though.

sb-relaxt-at:
Well, if we overwrite the duplicate method and call the parent's implementation this would trigger the before/after extensions. Therefore I considered creating onBeforeDuplicate and onAfterDuplicate. This would allow SiteTree to overwrite onAfterDuplicate instead of duplicate.

tractorcow:
SiteTree can already do that; You don't need to add the methods to DataObject for that to work.

All you need is the onBeforeDuplicate in SiteTree, and add the Sort = 0, right?

sb-relaxt-at:
My bad, I have forgotten that invokeWithExtensions also calls the method if it exists on the object itself. So its even easier, thanks :-)

tractorcow:
Cool. :)

gregsmirnov:
[...]

PR #1525, opened 2016-06-13
API Cleanup inconsistent SiteTree::duplicate API

tractorcow:
Merge this before accepting https://github.com/silverstripe-australia/silverstripe-advancedworkflow/pull/261

At the moment onBeforeDuplicate is run on both the original and the duplicate.

This corrects the behaviour to run only on the duplicate, and have it pass in the original to the first parameter consistently.

dataobject implements as:

```php
public function duplicate($doWrite = true) {
		$className = $this->class;
		$clone = new $className( $this->toMap(), false, $this->model );
		$clone->ID = 0;

		$clone->invokeWithExtensions('onBeforeDuplicate', $this, $doWrite);
		if($doWrite) {
			$clone->write();
			$this->duplicateManyManyRelations($this, $clone);
		}
		$clone->invokeWithExtensions('onAfterDuplicate', $this, $doWrite);

		return $clone;
	}
```

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -609,26 +609,13 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	}
 
 	/**
-	 * Create a duplicate of this node. Doesn't affect joined data - create a custom overloading of this if you need
-	 * such behaviour.
+	 * Reset Sort on duped page
 	 *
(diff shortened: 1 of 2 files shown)


tractorcow:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1253

Issue #1526, opened 2016-06-13
Add new page layout is broken after form updates in Framework forms


clarkepaul:
Caused by https://github.com/silverstripe/silverstripe-framework/issues/5677 but it's not a bug, it needs new styling applied to accommodate the new Bootstrap markup.
![pasted_image_13_06_16__4_39_pm](https://cloud.githubusercontent.com/assets/555033/15997578/d299d5e8-3189-11e6-99e4-73891d3cb223.png)

clarkepaul:
Requires silverstripe/silverstripe-framework#5697 merged first.
PR silverstripe/silverstripe-framework#5728
PR https://github.com/silverstripe/silverstripe-cms/pull/1533

Result:
![pasted_image_22_06_16__7_12_pm](https://cloud.githubusercontent.com/assets/555033/16257757/7a59cc2c-38ae-11e6-8189-4ac44e126a5b.png)

clarkepaul:
Can't seem to figure it out so will need a bit of dev help here. On step 1 when you select something from the tree dropdown it should do a check to see if the available children are allowed (which it isn't doing).

clarkepaul:
New PR created:
https://github.com/silverstripe/silverstripe-cms/pull/1554
https://github.com/silverstripe/silverstripe-framework/pull/5832

PR #1533, opened 2016-06-22
Add page layout based on Bootstrap markup

clarkepaul:
I've simplified the steps styles but added additional responsiveness. It does look a bit different from what we had, and could probably be improved still but this should satisfy the issue of the layout being broken.

Requires https://github.com/silverstripe/silverstripe-framework/pull/5697 merged first.
Requires https://github.com/silverstripe/silverstripe-framework/pull/5728
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1526

Result:
pasted_image_22_06_16__7_12_pm


clarkepaul:
Added styling to breadcrumb.
Moved notice to top.
pasted_image_23_06_16__10_28_am

clarkepaul:
Found some issues with the Javascript applied to the page, disabling of item if not allowed etc. Don't merge yet.

clarkepaul:
Can't seem to figure it out so will need a bit of dev help here. On step 1 when you select something from the tree dropdown it should do a check to see if the available children are allowed (which it isn't doing).

tractorcow:
I'm peer reviewing and manually merging this.

clarkepaul:
I've cleaned up this work and replacing it with https://github.com/silverstripe/silverstripe-cms/pull/1554

PR #1537, opened 2016-06-30
When page is disabled, rename "Edit" label in contextmenu and …

flamerohr:
When a page has been disabled, the context menu could hide the options that aren't relevant:
* Change "Edit" label to "View"
* Hide duplicate options, since they return a forbidden response when you can't edit anyway

### DIFF
--- a/client/lang/en.js
+++ b/client/lang/en.js
@@ -41,6 +41,7 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "Tree.AddSubPage": "Add new page here",
     "Tree.Duplicate": "Duplicate",
     "Tree.EditPage": "Edit",
+    "Tree.ViewPage": "View",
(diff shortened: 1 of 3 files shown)


client/lang/en.js
```diff
@@ -41,6 +41,7 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "Tree.AddSubPage": "Add new page here",
     "Tree.Duplicate": "Duplicate",
     "Tree.EditPage": "Edit",
+    "Tree.ViewPage": "View",
```
tractorcow: Please add this to client/lang/src/en.js too.

dhensby:
ping @tractorcow

tractorcow:
Restarting behat tests; These were recently fixed, so I'd expect them to pass this time.

tractorcow:
Merged, thanks @flamerohr

PR #1551, opened 2016-07-19
FIX: Fixes #1054 By preventing errors in the CMS _only_.

### LINKED ISSUE #1054: Compare mode on History tab on draft-deleted object produces Internal Server Error

phptek:
MINOR Boyscouting: Addition of missing braces, default return and missing types in @return annotation.

### DIFF
--- a/code/controllers/CMSPageHistoryController.php
+++ b/code/controllers/CMSPageHistoryController.php
@@ -358,7 +358,7 @@ class CMSPageHistoryController extends CMSMain {
 	/**
 	 * @param int $versionID
 	 * @param int $otherVersionID
-	 * @return Form
+	 * @return mixed
 	 */
 	public function CompareVersionsForm($versionID, $otherVersionID) {
 		if($versionID > $otherVersionID) {
@@ -369,16 +369,20 @@ class CMSPageHistoryController extends CMSMain {
 			$fromVersion = $versionID;
 		}
 
-		if(!$toVersion || !$toVersion) return false;
+		if(!$toVersion || !$toVersion) {
+			return false;
+		}
 		
 		$id = $this->currentPageID();
 		$page = DataObject::get_by_id("SiteTree", $id);
 		
-		if($page && !$page->canView()) {
-			return Security::permissionFailure($this);
-		}
+ 		if($page && $page->exists()) {
+			if(!$page->canView()) {
+				return Security::permissionFailure($this);
+			}
 
-		$record = $page->compareVersions($fromVersion, $toVersion);
+			$record = $page->compareVersions($fromVersion, $toVersion);
+		}
 
 		$fromVersionRecord = Versioned::get_version('SiteTree', $id, $fromVersion);
 		$toVersionRecord = Versioned::get_version('SiteTree', $id, $toVersion);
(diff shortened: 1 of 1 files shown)


code/controllers/CMSPageHistoryController.php
```diff
@@ -358,7 +358,7 @@ public function ShowVersionForm($versionID = null) {
 	/**
 	 * @param int $versionID
 	 * @param int $otherVersionID
-	 * @return Form
+	 * @return mixed Form|boolean|SS_HTTPResponse
```
dhensby: this syntax is slightly wrong. it should either be `mixed` OR `Form|bool`

I don't think this ever returns a `SS_HTTPResponse` object because `Security::permissionFailure` throws an exception rather than returns anything?! IIRC...

kinglozzer: `Security::permissionFailure()` does return a response, it’s `RequestHandler::httpError()` that doesn’t (this gets me every time!)

dhensby: Ah! well - I don't think this should be returning a `SS_HTTPResponse`, should it??

phptek: @dhensby `Security::permissionFailure()` returns an instance of `SS_HTTPResponse` ergo, returning `Security::permissionFailure()` on line 381 will mean that `CompareVersionsForm()` can return one of:

* A boolean
* An instance of `SS_HTTPResponse`
* An instance of `Form`

..personally I think multiple / mixed return types is crazy, but let's not over think this. This is how the method has been for years :-)

dhensby: > let's not over think this. This is how the method has been for years

Aceepted, the syntax is still wrong, though - would you mind doing one of either `mixed` or the types it can return?

phptek: @dhensby Explicitly adding a pipe-delimited list of the available return types _is permitted_ in phpDoc,:
 
>If you want to explicitly show multiple possible return types, list them pipe-delimited without spaces (e.g. "@return int|string").

https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.return.pkg.html

phptek: Will do :-)

dhensby: Thanks, really appreciate it!

assertchris: In general I think it's better to switch to mixed when return types > 2...

phptek: @assertchris @dhensby Done. Force-pushed. Over to you guys.

dhensby:
Thanks

PR #1555, opened 2016-07-21
FIX do not show HiddenClass pages in allowed children

### LINKED ISSUE #1553: "Add new page here" shows HiddenClass records
**SS/CMS:** 3.4.0
**Blog:** 2.4.0

When I "Add New", I get a list of Page Types that is Page, Blog and BlogPost.

I've extended the VirtualPage, ErrorPage and RedirectorPage classes with DisabledVirtualPage etc which `implements HiddenClass`. They do **not** show up in the list when I click "Add New".

However, if I right click on an existing page in the site tree and hover over "Add new page here", those page types **are** shown.

I suggest that wherever these page types are loaded from should be the same logic as the main add new page.

robbieaverill:
* Resolves #1553

This change could probably be back merged into earlier 3.x versions too.

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -2694,7 +2694,10 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 				} else {
 					$subclasses = ClassInfo::subclassesFor($candidate);
 					foreach($subclasses as $subclass) {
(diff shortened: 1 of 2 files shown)


dhensby:
are there any tests for this method that could be updated?

robbieaverill:
Good question, checking...

dhensby:
Great stuff, thanks!

PR #1558, opened 2016-07-24
FIX allow custom query params when archiving a page (resolves #1557)

### LINKED ISSUE #1557: Archiving a Page with custom query string parameters throws error
**SS:** 3.4.0
**CMS:** 3.4.0
**Blog:** 2.4.0

See tractorcow/silverstripe-fluent#219.

Is caused by `SilverStripeNavigatorItem_ArchiveLink::getLink` not using `Controller::join_links`

robbieaverill:
The original code doesn't use `Controller::join_links()` so if the `PreviewLink()` already contains query string parameters, it may end up resulting in a link looking like this:

```
/path/to/preview?my=query&string=here?archiveDate=1234...
```

This resolves the issue by using the `join_links()` method as used elsewhere in the same file.


tractorcow:
I was just about to create this very same PR. :)

robbieaverill:
Yours looks better!

tractorcow:
See https://github.com/silverstripe/silverstripe-cms/pull/1560/files?w=1

I noticed there were a few places where links weren't being joined properly...

robbieaverill:
Cool - closing as covered in #1560

PR #1562, opened 2016-07-25
Adjust layout of page search form

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/5692#issuecomment-235109269

Closes https://github.com/silverstripe/silverstripe-framework/issues/5692


clarkepaul:
Tested on IE and looks good.

PR #1564, opened 2016-07-26
FIX Direct edit file by URL

jonom:
Trying to navigate directly to the edit view of a file by URL would fail if a folder other than the file’s parent was held in session data. Removed redundant storage of current ID in session and added folder detection for edit view. This also fixes breadcrumbs when directly navigating to file edit URL.

### DIFF
--- a/code/controllers/AssetAdmin.php
+++ b/code/controllers/AssetAdmin.php
@@ -54,8 +54,6 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider{
 			return $this->getRequest()->requestVar('ID');
 		} elseif (is_numeric($this->urlParams['ID'])) {
 			return $this->urlParams['ID'];
-		} elseif(Session::get("{$this->class}.currentPage")) {
-			return Session::get("{$this->class}.currentPage");
 		} else {
 			return 0;
 		}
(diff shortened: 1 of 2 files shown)


dhensby:
behat failure looks related

jonom:
Yeah, I should have looked outside of `cms` for reliance on that session data. I'll try a different solution.

dhensby:
I do think this should be kept outside of the session, but it'll be a fairly difficult change, I'd have thought. We can store it in the URL as a parameter, but how many places will we need to touch to make it work?

jonom:
I think the only other place that session data is touched is in LeftAndMain.php `setCurrentPageID()` and `currentPageID()` methods, but removing the session part could have far reaching consequences I imagine.

This kind of reliance on session data often leads to a broken UX in my experience, as the data can outlive it's intended use and be read outside of original context. It's especially problematic if you ever try to have two CMS windows open at once, for instance if you want to reference one set of data while editing another. Click 'Back' or 'Next' and suddenly you're editing the wrong set :S

I have tried a new approach but couldn't re-open this PR due to force pushing, sorry. New one at  https://github.com/silverstripe/silverstripe-cms/pull/1566

dhensby:
agreed

PR #1566, opened 2016-07-28
FIX Direct edit file by URL

jonom:
Trying to navigate directly to the edit view of a file by URL would fail if a folder other than the file’s parent was held in session data, as it would take precedence over falsey but valid values such as 0 or null. Additionally, breadcrumbs would be missing if navigating directly to a file edit URL without navigating through the gridfield first. Storage of current folder ID in session retained for backwards compatibility, but reliance on it has been removed.

Behat test will fail without companion fix at https://github.com/silverstripe/silverstripe-framework/pull/5848

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ addons:
 
 env:
   global:
-    - CORE_RELEASE=3
+    - CORE_RELEASE=3.4
     - "ARTIFACTS_AWS_REGION=us-east-1"
     - "ARTIFACTS_S3_BUCKET=silverstripe-travis-artifacts"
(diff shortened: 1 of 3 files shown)


jonom:
Hmm I assumed behat tests would use framework version 3.4.x-dev (which contains required patch https://github.com/silverstripe/silverstripe-framework/pull/5848) since this PR is aimed at 3.4 branch, but it's actually using 3.x-dev.

Should I update this PR so .travis.yml is modified to read

```yml
env:
  global:
    - CORE_RELEASE=3.4
```

? It's set to 3 currently.

dhensby:
I think that patch to travis.yml is a good idea!

jonom:
That did the trick

PR #1568, opened 2016-07-30
Added missing translations

faloude:
-Label on 'Batch actions' button
-'Restore' in dropdown
-'Unpublish' in dropdown


dhensby:
@tractorcow does transifex pick up new translations this way?

tractorcow:
No it doesn't. It would get lost on the next release.

@faloude can you please create an account and I'll authorise you as the dutch translator on https://www.transifex.com/silverstripe/silverstripe-cms/

You don't seem to have a profile.

image

I have added these strings manually.

PR #1569, opened 2016-08-01
BUG Fix issue with old asset-admin repeating "level up" button

tractorcow:
This fixes issues with dist files not being built properly, as per the same fix applied to CMS at https://github.com/silverstripe/silverstripe-framework/commit/6e74b57c36816f3ba763ea0f8fb862306a362506.

I've removed some hacky JS that tried to move the level up button around the dom because it wasn't behaving very well.

### DIFF
--- a/client/dist/js/AssetAdmin.js
+++ b/client/dist/js/AssetAdmin.js
@@ -42,7 +42,7 @@
 			}
 		});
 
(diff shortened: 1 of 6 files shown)


dhensby:
These selectors are absolutely insane :/

Issue #1572, opened 2016-08-04
Namespace CMS


tractorcow:
See epic link: https://github.com/silverstripe/silverstripe-framework/issues/4562

Dependency PRs (can be merged up front safely).
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/5881
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/5873
 - [x] silverstripe/silverstripe-cms: https://github.com/silverstripe/silverstripe-cms/pull/1571
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/5874
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/214
 - [x] https://github.com/silverstripe-labs/silverstripe-behat-extension/pull/122
 
 
Once these are merged the main PR are at:
 - [x] https://github.com/silverstripe/silverstripe-postgresql/pull/53
 - [x] https://github.com/silverstripe-labs/silverstripe-sqlite3/pull/24
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/5875
 - [x] silverstripe/silverstripe-cms: https://github.com/silverstripe/silverstripe-cms/pull/1574
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/213
 - [x] https://github.com/silverstripe/silverstripe-installer/pull/130
 - [x] https://github.com/silverstripe/silverstripe-siteconfig/pull/34
 - [x] https://github.com/silverstripe-labs/silverstripe-reports/pull/42

tractorcow:
Framework behat tests:

![image](https://cloud.githubusercontent.com/assets/936064/17471087/a383e752-5d95-11e6-9351-fd5fc7e81f70.png)

tractorcow:
Er, don't merge yet... there's a bug in `# cms/tests/behat/features/manage-page-permissions.feature:14` that needs fixing.

tractorcow:
Also `# cms/tests/behat/features/search-for-a-page.feature:20`

tractorcow:
Also `# cms/tests/behat/features/publish-a-page.feature:56`

tractorcow:
Ack, caused by https://github.com/silverstripe/silverstripe-asset-admin/pull/214

tractorcow:
![image](https://cloud.githubusercontent.com/assets/936064/17502553/0442fc52-5e3d-11e6-9c3a-27c43cf6a5ce.png)

[...]

PR #1574, opened 2016-08-08
API namespace CMS

tractorcow:
See https://github.com/silverstripe/silverstripe-cms/issues/1572


code/Controllers/CMSMain.php
```diff
@@ -403,16 +414,18 @@ public function SearchForm() {
...
 		foreach($actions->dataFields() as $action) {
+			/** @var FormAction $action */
 			$action->setUseButtonTag(true);
 		}
 
 		// Create the form
-		$form = Form::create($this, 'SilverStripe\\CMS\\Search\\SearchForm', $fields, $actions)
+		/** @skipUpgrade */
+		$form = Form::create($this, 'SearchForm', $fields, $actions)
 			->addExtraClass('cms-search-form')
```
hafriedlander: This wasn't required before. How is this change related to namespacing?

tractorcow: I think it came out of a post-namespace test; It may have been a pre-existing issue I fixed here.

hafriedlander:
Failures are in psql & sqlite. Merging to avoid even-more-broken master, but need fixes for those modules ASAP.

PR #1585, opened 2016-08-18
Use Injector to create the default pages

Firesphere:
This way, the Page and Models in MySite can be namespaced.


Firesphere:
On second thought. Nevermind this one. I'll collect the issues into one PR, I think that's the better way to go.

PR #1586, opened 2016-08-22
API: Add checker for Ajax calls so error page is not returned

flamerohr:
This would allow for error messages to passed easily when an ajax call is made instead of a full form submit.


dhensby:
Rather than analysing the *type* of request, we should look at the Accept header and send an appropriate response.

flamerohr:
Yes, although the `Director::is_ajax()` is mainly checking the `x-requested-with:XMLHttpRequest` it also handles a few legacy fallback checks for ajax as well.

I had left it open to whichever error handler to format their own response, perhaps could be another pull request to format and send the appropriate response?

PR #1590, opened 2016-08-31
API Remove AssetAdmin (moved to asset-admin module)

chillu:
The last blocker for this was the inability to edit folders, which will be resolved shortly with https://github.com/silverstripe/silverstripe-asset-admin/issues/156. We still haven't fixed the files search frontend (https://github.com/silverstripe/silverstripe-asset-admin/issues/121), but that's not a blocker for keeping the old AssetAdmin around in my opinion - it's currently marked to be resolved in beta1.

The `admin/old-assets` UI looks pretty broken anyway (presumably since Paul's CSS refactorings), so it's hardly the fallback UI you'd want to use in production instead of the new `admin/assets` - so no point fixing it, no point keeping it around.

image

Merge together with https://github.com/silverstripe/silverstripe-framework/pull/5934

I've separated out "migrate existing translations from cms": https://github.com/silverstripe/silverstripe-asset-admin/issues/239


dhensby:
I don't really want to merge this until we have a new asset-admin tag because the current installer for 4.x is broken at the moment...

chillu:
@dhensby Fixing the installer is being discussed in https://github.com/silverstripe/silverstripe-installer/pull/132 at the moment. I don't see how this affects this PR though - either the installer is working and you can get both `cms` and `asset-admin`, or its broken and you can't get either. This PR won't make this situation any better or worse for devs or users.

dhensby:
Hmm, good point - ok

PR #1599, opened 2016-09-05
BUG Ensure changes in class write to an instance of the new class

### LINKED ISSUE #1210: BUG Changing the page class causes errors
This is because the new records is written as ClassName = 'NewClassName', but the actual PHP class that has ->write invoked on it is the old class. This causes errors where, for instance, the old PHPClass invokes has_many relations in onBeforeWrite that don't exist on the new class.

Potential fix (I would need to test / document, this is just a proof of concept)

https://github.com/tractorcow/silverstripe-cms/commit/703d50ec406aa01169e41aa9f4c3a60a7cafd68f

https://github.com/tractorcow/silverstripe-framework/commit/fa35d1d0f201e8990139839d218f68297e2c3c27

Not sure why the code is the way that it was.... >_>

tractorcow:
Fixes #1210

See https://github.com/silverstripe/silverstripe-framework/pull/5950 for probably-not-necessary cleanup in core to validate at the DataObject level.

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -1006,12 +1006,8 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
 		// Update the class instance if necessary
 		if(isset($data['ClassName']) && $data['ClassName'] != $record->ClassName) {
-			$newClassName = $record->ClassName;
-			// The records originally saved attribute was overwritten by $form->saveInto($record) before.
-			// This is necessary for newClassInstance() to work as expected, and trigger change detection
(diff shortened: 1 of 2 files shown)


code/Controllers/CMSMain.php
```diff
@@ -1006,12 +1006,8 @@ public function save($data, $form) {
 
 		// Update the class instance if necessary
 		if(isset($data['ClassName']) && $data['ClassName'] != $record->ClassName) {
-			$newClassName = $record->ClassName;
```
tractorcow: This is the part that was incorrectly creating the new instance with the wrong physical class.

PR #1602, opened 2016-09-07
Upgrade code for framework namespaces

tractorcow:
Merge with https://github.com/silverstripe/silverstripe-framework/pull/5921

Remove `@package / @subpackage`
Fix inifinite loop in VirtualPage_Controller


code/Model/SiteTree.php
```diff
@@ -2036,12 +2033,6 @@ public function getCMSFields() {
 				. "</p>"), "Title");
 		}
 
-		// Backwards compat: Rewrite nested "Content" tabs to toplevel
```
chillu: We haven't thrown deprecation notices on this, so might catch some devs out - but we've mentioned it in the 3.0.0 changelog: https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/04_Changelogs/3.0.0.md#changed-tab-paths-in-sitetree-getcmsfields-tab-paths

code/Model/SiteTreeLinkTracking_Parser.php
```diff
@@ -91,7 +92,7 @@ public function process(SS_HTMLValue $htmlValue)
 					'Target' => $matches['id'],
 					'Anchor' => null,
 					'DOMReference' => $link,
-					'Broken' => !DataObject::get_by_id('File', $matches['id'])
+					'Broken' => !DataObject::get_by_id('SilverStripe\\Assets\\File', $matches['id'])
```
chillu: We could take this as an opportunity to rewrite this all to the less verbose `File::get()->byID()` ... but then again, that's annoyingly not cached in the same way as `get_by_id()`

code/Model/SiteTree.php
tractorcow: Yes we have actually. See FieldList::rewriteTabPath in 3.x

code/Model/SiteTreeLinkTracking_Parser.php
tractorcow: Yeah that's the only reason. =(

Issue #1386, opened 2016-02-03
VirtualPage controller doesn't fall back to methods on copied page controller.


tractorcow:
VirtualPage_controller uses the copied page as a failover, when it should use the controller of that page as the failover, which in turn will failover to that copied page.

See https://github.com/silverstripe/silverstripe-subsites/pull/179

Related: Maybe we should get URL actions working as well? (If easy) https://github.com/silverstripe/silverstripe-cms/issues/1101

dhensby:
see https://github.com/silverstripe/silverstripe-cms/pull/1102

was deemed an API change :/

dhensby:
and https://github.com/silverstripe/silverstripe-cms/issues/1099 and https://github.com/silverstripe/silverstripe-cms/pull/905

tractorcow:
Yeah, but #1102 only invokes init on the copied page, not the controller of that page. A lot more is necessary to address this issue.

tractorcow:
This will be an API change... best to do it in master.

tractorcow:
Also see https://github.com/silverstripe/silverstripe-subsites/pull/169

tractorcow:
And https://github.com/silverstripe/silverstripe-subsites/pull/113

tractorcow:
Possibly related: https://github.com/silverstripe/silverstripe-subsites/issues/235 (although to do with template, not methods).

tractorcow:
Possibly related: https://github.com/silverstripe/silverstripe-cms/issues/1336#issuecomment-193352164

kinglozzer:
@tractorcow I was playing around with this a little this afternoon (because of the issue you mentioned about URL actions not working) before I saw your suggested approach 😅. The approach I went for was to simply get rid of `VirtualPage_Controller` and instead use the target page’s controller: setting the VirtualPage instance to be the data record on that.

I.e:
```
MyCustomPage_Controller --> dataRecord = VirtualPage --> CopyContentFrom = MyCustomPage
```

Here’s where I got up to: https://github.com/kinglozzer/silverstripe-cms/commit/a8750065e4ee96e06e42ab92fcc41981b219597e. It would need tidying up, but all the existing VirtualPage unit tests pass and it solves the issue about forwarding `allowed_actions`, `init()` etc. It’d also mean patching subsites, as `SubsiteVirtualPage_Controller extends VirtualPage_Controller`

dhensby:
[...]

Issue #1336, opened 2015-11-24
VirtualPage doesn't find correct templates


henrypenny:
The getViewer function of VirtualPage does not pass in a dataRecord - this causes the default Page.ss to be returned in all cases instead of the template suitable to render the 'copied' page.

	public function getViewer($action) {
		$originalClass = get_class($this->CopyContentFrom());
		if ($originalClass == 'SiteTree') $name = 'Page_Controller';
		else $name = $originalClass."_Controller";
		$controller = new $name();
		return $controller->getViewer($action);
	}

It needs to look more like:

	public function getViewer($action) {
		$originalClass = get_class($this->CopyContentFrom());
		if ($originalClass == 'SiteTree') $name = 'Page_Controller';
		else $name = $originalClass."_Controller";
                // pass in dataRecord of correct type.
		$controller = new $name(new $originalClass());
		return $controller->getViewer($action);
	}

henrypenny:
I have patched this problem in my current project by adding this constructor to my Page_Controller:

class Page_Controller extends ContentController {

    /**
     * Page_Controller constructor.
     * @param null $dataRecord
     *
     * Constructor is required to patch an issue with Virtual pages not finding the correct template when no dataRecord is passed
     */
    public function __construct($dataRecord = null)
    {
        if($dataRecord == null) {
            $dataClass = get_class($this);
            $dataClass = str_replace('_Controller', '', $dataClass);
            $dataRecord = new $dataClass();
        }
        parent::__construct($dataRecord);
    }

tractorcow:
Wouldn't this result in rendering a page with all of it's fields nullified?

tractorcow:
Maybe VirtualPage_Controller::getViewer should look like the below:

```php
public function getViewer($action) {
	$object = $this->CopyContentFrom();
	$controller = ModelAsController::controller_for($object, $action);
	return $controller->getViewer($action);
}
```

henrypenny:
It looks like that Controller instance is only created to locate the template in this instance.

The main problem is this:

    class ContentController extends Controller {
        ...
	    public function __construct($dataRecord = null) {
		    if(!$dataRecord) {
			    $dataRecord = new Page();
                            ...

If you don't pass in a $dataRecord the controller assumes that its a Page instance. And you end up rendering the Page.ss template.

[...]

Issue #1101, opened 2014-09-16
VirtualPage controller doesn't forward actions to target controller


stnvh:
Actions specified on the target controller via ```$allowed_actions``` will not be called when viewed via a virtual page.

SS 3.1.6

tractorcow:
Related to https://github.com/silverstripe/silverstripe-subsites/pull/179.

tractorcow:
Fixed with #1605

PR #1605, opened 2016-09-08
Virtualpage refactor

kinglozzer:
Opening this early for review, probably needs more test coverage before being merged. I also need to write a patch for subsites, as that includes a class that extends `VirtualPage_Controller` (which I’ve deleted).

Fixes:

- https://github.com/silverstripe/silverstripe-cms/issues/1336
- https://github.com/silverstripe/silverstripe-cms/issues/1386
- https://github.com/silverstripe/silverstripe-cms/issues/1101


code/Controllers/ModelAsController.php
```diff
@@ -32,30 +32,14 @@ class ModelAsController extends Controller implements NestedController {
...
-				: "SilverStripe\\CMS\\Controllers\\ContentController";
-		}
-
-		if($action && class_exists($controller . '_' . ucfirst($action))) {
-			$controller = $controller . '_' . ucfirst($action);
-		}
-
-		return class_exists($controller) ? Injector::inst()->create($controller, $sitetree) : $sitetree;
+		Deprecation::notice('5.0', 'Use $sitetree->getAssociatedController() instead');
+		return $sitetree->getAssociatedController($action);
```
kinglozzer: I was unsure about this at first - is it better for the model to choose a controller, or for the controller to work it out itself? The more I think about it, the more I think this approach makes sense, especially given that:

1. If a controller lives in a different namespace to its “associated” page type, it’s hard to work out which controller we should use (we can’t just to `{$class}_Controller` like we do currently).
2. This approach would allow two page classes to share a controller without relying on the `{$class}_Controller` logic & inheritance

code/Model/SiteTree.php
```diff
@@ -2711,6 +2712,43 @@ public function getPageLevel() {
 	}
 
 	/**
+	 * @param string $action
+	 * @return Controller
+	 */
+	public function getAssociatedController($action = null) {
```
kinglozzer: Not sure about the naming for this... perhaps just `getController()` would be better?

code/Controllers/ModelAsController.php
dhensby: I like this a lot :)

How how do we set it for new page types? Does it need to be set every time or is in going to be inferred by default?

code/Model/SiteTree.php
dhensby: `getController` seems nice

If we could also `setController` that'd be even nicer to allow per instance overrides or something

code/Model/SiteTree.php
```diff
@@ -2711,6 +2712,43 @@ public function getPageLevel() {
...
+		$controllerName = $this->getControllerName($action);
+		return Injector::inst()->create($controllerName, $this);
+	}
+
+	/**
+	 * @param string $action
+	 * @return string
+	 */
+	public function getControllerName($action = null) {
+		$action = ucfirst($action); // Needed as Injector service names are case sensitive
```
dhensby: who does the action need to be `ucfirst`?

dhensby:
The moving away from `MoelAsController::controller_for()` needs some wide-spread approval from @silverstripe/core-team, I think - maybe this should be raised on the dev mailing list too?

code/Controllers/ModelAsController.php
kinglozzer: The default behaviour (inherited from `SiteTree::getControllerName()`) would be exactly what we have currently - `Page` would return `Page_Controller`, `MyPageType` would return `MyPageType_Controller`. You could then override it in your page class with something like:

```php
public function getControllerName($action = null)
{
   return 'Mysite\Controllers\PageController';
}
```

I still need to do some testing with namespaces. If you call `->getControllerName()` on `Mysite\Model\Page`, it should return `Mysite\Model\Page_Controller` to preserve BC. I’m working on the assumption that `ClassInfo::ancestry()` returns FQCNs, but I haven’t actually tested that yet.

Tbh, the `getControllerName()` method only exists for some convenience. It just saves you having to type out `return Injector::inst()->create('Some\Namespace\Class'...` in full.

dhensby: I think the controller name should be an optional config value, then? If that exists, return it, else try to figure it out?

code/Model/SiteTree.php
kinglozzer: Hmm, sounds interesting. Would we run the risk of ending up with circular dependencies, though? Page has a controller, which has a page, which has a controller... you get the idea.

dhensby: If you do stupid things, expect stupid results :P

---

I'd expect that `setController` sets the record on the controller as part of it.

code/Model/SiteTree.php
kinglozzer: It’s how `ModelAsController::controller_for()` works currently. I added the comment inline to describe why I *think* it’s like that...

I’ve just spotted a mistake I made when trying to tidy this logic up... `Page_Controller_Action` should take priority over `Page_Controller`. Will fix.

code/Controllers/ModelAsController.php
kinglozzer: Good idea, `private static $controller_class` is probably nicer than having to override a method each time

code/Model/SiteTree.php
dhensby: hmm, this is why injector needs to be case insensitive, class names are and this shouldn't be something we do - it feels risky... but if that's how it was done before... carry on

kinglozzer:
Okay, I’ve scaled this back a lot so that we can concentrate on fixing the issues with `VirtualPage` for now. The other stuff can be discussed and implemented later if needed.

I have had a look at the subsites module, but it’s horribly broken with SilverStripe 4 at the moment, so I can’t really patch it... The BC-break of this PR shouldn’t be hard to fix, it’s a fairly simple class that would be broken by this change - https://github.com/silverstripe/silverstripe-subsites/blob/7c97a8ae74fb2841f8f0ca11058f89f6a922ac81/code/SubsitesVirtualPage.php#L175-L193

code/Controllers/ModelAsController.php
```diff
@@ -37,25 +37,13 @@ class ModelAsController extends Controller implements NestedController {
...
-				}
-			}
-			$controller = ($class !== null)
-				? "{$class}_Controller"
-				: "SilverStripe\\CMS\\Controllers\\ContentController";
-		}
+		$controller = $sitetree->getControllerName();
 
-		if($action && class_exists($controller . '_' . ucfirst($action))) {
+		if ($action && class_exists($controller . '_' . ucfirst($action))) {
```
dhensby: It seems strange to me to check `class_exists` and then use `Injector`, shouldn't we check for a registered service with `Injector`?

code/Model/SiteTree.php
```diff
@@ -2713,6 +2713,26 @@ public function getPageLevel() {
...
+			$controller = ContentController::class;
+		} else {
+			$ancestry = ClassInfo::ancestry($this->class);
+			while ($class = array_pop($ancestry)) {
+				if (class_exists($class . "_Controller")) {
+					break;
+				}
+			}
+
+			$controller = ($class !== null) ? "{$class}_Controller" : ContentController::class;
```
dhensby: Why strict type check $class? Surely empty `$class` shouldn't be accepted as valid?

dhensby:
No longer `[Dont Merge]`?

code/Model/SiteTree.php
kinglozzer: No reason, I just copied and pasted this block of code 😝

code/Controllers/ModelAsController.php
kinglozzer: We could do both, yeah. Tbh I wondered if we should deprecate this `Page_Controller_Action` stuff, it seems a bit “magic” and I don’t think it’s documented

dhensby:
OK - lets merge as is because that way we avoid unintended regressions - perhaps add an issue to clean this up later?

kinglozzer:
Thanks! I’m going to add a line to the upgrade docs in framework about the class that has been removed, I’ll just commit that directly though

mikenz:
Just found the class that has been removed ;-)

PR #1616, opened 2016-09-20
FIX allowedChildren only filters direct implementors of HiddenPage

dhensby:
Only direct implementors of `HiddenPage` should be hidden.

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -2695,7 +2695,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 				} else {
 					$subclasses = ClassInfo::subclassesFor($candidate);
 					foreach($subclasses as $subclass) {
-						if ($subclass == 'SiteTree_root' || singleton($subclass) instanceof HiddenClass) {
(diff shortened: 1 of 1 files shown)


tractorcow:
See SiteTreeTest::testAllowedChildrenContainsCoreSubclassesButNotHiddenClass

tractorcow:
Tests seem to enforce the existing interpretation of what `HiddenClass` means, which is inheritance. Unfortunately this means we can't merge this into 3.x without breaking semver.

That's not practically useful, but it is consistent with the object oriented inheritance nature of PHP. I would argue that having an interface for hidden classes is not a very appropriate mechanism, so shall we just deprecate it instead, and move to some kind of config?

We already have `hide_ancestor`, but we don't have a way to specify hidden directly on an object itself.

We could deprecate and add the new behaviour in 3.5, and remove the interface entirely in 4.x.

dhensby:
Understood - I think this breaks developer expectation and the tests were probably written to accept the currently written function and not the developer expectation *but* if we do break tests, we probably break someones code - so I accept that.

PR #1618, opened 2016-09-21
API use new bootstrap button row template

tractorcow:
Merge with https://github.com/silverstripe/silverstripe-framework/pull/6036


flamerohr:
looks good to me, a rebase would probably fix the behat test failing

PR #1621, opened 2016-09-22
FIX: Break dist javascript and onto newlines.

chillu:
This doesn’t have much impact on resulting JS size, but it will
hopefully make merge conflicts much less frequent.

The CSS growth is a little higher (6.5% increase in size) but is not
material.

If this materially reduces the number of merge conflicts we have, by
letting the git merge tools resolve some dist file mergers, I think it
would be worth it.

See https://github.com/silverstripe/silverstripe-framework/pull/6051


sminnee:
Fine by me once tests pass.

PR #1629, opened 2016-09-22
Simplifying SiteTree::getControllerName

dhensby:
@kinglozzer a bit of a cleanup - just want to get an A rating on scrutinizer ;)


kinglozzer:
Sorry @dhensby, totes forgot to do this when I got home 😞

PR #1631, opened 2016-09-26
Code Coverage via CodeCov.io

chillu:
See https://github.com/silverstripe/silverstripe-framework/pull/6089

Note that the `phpunit.xml.dist` won't apply until we make builds run without `travis-support` (out of scope for this PR). Once that's resolved it'll automatically pick up the PHPUnit defaults set here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
     - php: 5.6
       env: NPM_TEST=1
     - php: 7.0
-      env: DB=MYSQL PDO=1 PHPUNIT_TEST=1
+      env: DB=MYSQL PDO=1 PHPUNIT_COVERAGE_TEST=1
 
 before_script:
  - composer self-update || true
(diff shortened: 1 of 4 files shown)


dhensby:
should the coverage test (and php7) be allowed failures to allow it to not impact completion time for PRs?

chillu:
I don't think so - if PHP7 builds break, a merge should be blocked - to avoid regressions. And we don't want to run separate PHP7 and PHP7+Coverage builds, because it saturates our "build queue" for this particular Github organisation (API key) - even if a specific PR might go green a minute earlier.

chillu:
Don't merge yet - it's only running coverage against `mysite` for some reason.

dhensby:
I suppose that's a fair point.

Can we turn off havin @codecov-io leave comments on PRs, it's going to increase the amount of noise we get in our notifications

chillu:
@dhensby Yeah we probably should implement a `threshold` instead, and fail the build on a relative decrease of test coverage: http://docs.codecov.io/docs/commit-status. We can even set different thresholds for different folders, e.g. don't allow any coverage drop in `ORM/`, with more relaxed settings in `Dev/`.

dhensby:
I'm all for failing builds on coverage decrease, I just want to avoid notification/email flooding with codecov comments on PRs.

sminnee:
> Can we turn off having @codecov-io leave comments on PRs, it's going to increase the amount of noise we get in our notifications

I think the messages are of value.

sminnee:
@chillu the simpler test run work may have fixed the problems you were having (travis-support tends to scare 3rd party tools) so you may want to re-base and re-try this.

PR #1632, opened 2016-09-27
Switched to SASSLint

chillu:
Doesn't actually run any tests because we've excluded the legacy files,
but that's no different to the old scss-lint run.

See https://github.com/silverstripe/silverstripe-framework/pull/6016


chillu:
Had `npm run test`, which exits with `0` when no tests are found (should’ve been `npm run build`) - just rebuilding

chillu:
@sminnee Tests are green, wanna press merge? :)

PR #1634, opened 2016-09-27
FIX: Remove reference to Object class.

sminnee:
The Object class is due to be removed and so this removes dependence on it.


sminnee:
Yeah, don't ignore behat failures - re-trigger the build and confirm that it corrects itself.

sminnee:
OK @kinglozzer I fixed the remaining issue.

kinglozzer:
So *that’s* how the “Update branch” button works. Will squash and merge on green :)

PR #1635, opened 2016-09-27
MINOR: Replace get_class($this) with static::class

sminnee:
Thanks @assertchris for the recommendation

### DIFF
--- a/code/Model/ErrorPage.php
+++ b/code/Model/ErrorPage.php
@@ -122,7 +122,7 @@ class ErrorPage extends Page {
 		parent::requireDefaultRecords();
 
 		// Only run on ErrorPage class directly, not subclasses
-		if (get_class($this) !== 'SilverStripe\\CMS\\Model\\ErrorPage' || !SiteTree::config()->create_default_pages) {
+		if (static::class !== self::class || !SiteTree::config()->create_default_pages) {
 			return;
 		}
 
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -330,12 +330,12 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 			$conditions[] = array('"SiteTree"."ParentID"' => 0);
 		}
 		/** @var SiteTree $sitetree */
-		$sitetree = DataObject::get_one(__CLASS__, $conditions, $cache);
+		$sitetree = DataObject::get_one(self::class, $conditions, $cache);
 
 		/// Fall back on a unique URLSegment for b/c.
 		if(	!$sitetree
 			&& self::config()->nested_urls
-			&& $sitetree = DataObject::get_one(__CLASS__, array(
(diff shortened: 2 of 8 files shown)


code/Model/ErrorPage.php
```diff
@@ -122,7 +122,7 @@ public function requireDefaultRecords()
 		parent::requireDefaultRecords();
 
 		// Only run on ErrorPage class directly, not subclasses
-		if (get_class($this) !== 'SilverStripe\\CMS\\Model\\ErrorPage' || !SiteTree::config()->create_default_pages) {
+		if (static::class !== 'SilverStripe\\CMS\\Model\\ErrorPage' || !SiteTree::config()->create_default_pages) {
```
assertchris:
```php
use SilverStripe\CMS\Model\ErrorPage;

// ...

if (static::class !== ErrorPage::class)  {
```

code/Model/SiteTree.php
```diff
@@ -2231,7 +2231,7 @@ public function getCMSActions() {
 		$live = Versioned::get_one_by_stage('SilverStripe\\CMS\\Model\\SiteTree', Versioned::LIVE, array(
 			'"SiteTree"."ID"' => $this->ID
 		));
-		$infoTemplate = SSViewer::get_templates_by_class(get_class($this), '_Information', __CLASS__);
+		$infoTemplate = SSViewer::get_templates_by_class(static::class, '_Information', __CLASS__);
```
assertchris: Or `SSViewer::get_templates_by_class(static::class, '_Information', self::class);`

code/Model/SiteTree.php
```diff
@@ -1513,7 +1513,7 @@ public function requireDefaultRecords() {
 		parent::requireDefaultRecords();
 
 		// default pages
-		if(get_class($this) == __CLASS__ && $this->config()->create_default_pages) {
+		if(static::class == __CLASS__ && $this->config()->create_default_pages) {
```
assertchris: `static::class == __CLASS__` or `static::class == self::class`

tractorcow: 👍

sminnee:
Can anyone think of a place where `__CLASS__` could not be replaced with `self::class`? I only code by global search & replace.

assertchris:
According to docs, `::class` is FQCN. `__CLASS__` is FQCN where defined (so semantically the same as `self`, which is not changed by inheritance context. I can't think of a reason not to use `static::class` and `self::class` instead of `static::class` (to replace `get_class($this)`) and `__CLASS__`.

code/Model/ErrorPage.php
sminnee: Actually `static::class !== self::class` can be used here

sminnee:
Force pushed an amended commit.

assertchris:
For reference:
screen shot 2016-09-28 at 11 56 39 am

kinglozzer:
Good stuff 👍

PR #1636, opened 2016-10-03
FIX Sort order for duplicated child pages is now retained

dhensby:
This fixes an issue where pages duplicated with their children cause the new child pages to all have the same Sort value causing the child pages to jump around on save.


tractorcow:
Will merge on green.

PR #1641, opened 2016-10-08
Use the injector to instantiate a new Page.

Firesphere:
This way, a new project can have a namespaced Page Class.
It can then be mapped in the YML config with

```
SilverStripe\Core\Injector\Injector:
  Page:
    class: My\Namespace\Page
```

### DIFF
--- a/code/Model/ErrorPage.php
+++ b/code/Model/ErrorPage.php
@@ -143,7 +143,7 @@ class ErrorPage extends Page {
 		$page = ErrorPage::get()->filter('ErrorCode', $code)->first();
 		$pageExists = !empty($page);
 		if(!$pageExists) {
-			$page = new ErrorPage($defaultData);
+			$page = Injector::inst()->create(self::class, $defaultData);
 			$page->write();
 			$page->copyVersionToStage(Versioned::DRAFT, Versioned::LIVE);
 		}
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -15,6 +15,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\Deprecation;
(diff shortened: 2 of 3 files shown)


code/Model/SiteTree.php
```diff
@@ -1515,7 +1516,7 @@ public function requireDefaultRecords() {
 		// default pages
 		if(static::class == self::class && $this->config()->create_default_pages) {
 			if(!SiteTree::get_by_link(RootURLController::config()->default_homepage_link)) {
-				$homepage = new Page();
+				$homepage = Injector::inst()->get('Page', false);
```
dhensby: I think using `create` rather than `get` is a bit cleaner here

Firesphere: Ah, right, semantically better indeed.

Firesphere:
Updating as per @dhensby's comment, and also applied the same logic to the ErrorPage, to keep things consistent.

Firesphere:
Breaking of build is unrelated to this PR.

code/Model/ErrorPage.php
```diff
@@ -143,7 +143,7 @@ protected function requireDefaultRecordFixture($defaultData) {
 		$page = ErrorPage::get()->filter('ErrorCode', $code)->first();
 		$pageExists = !empty($page);
 		if(!$pageExists) {
-			$page = new ErrorPage($defaultData);
+			$page = Injector::inst()->create('SilverStripe\\CMS\\Model\\ErrorPage', $defaultData);
```
kinglozzer: `ErrorPage::class` :)

Firesphere: Does that include the namespace? I did not know that!
Thanks, will update :)

chillu:
Given the accepted RFC option is to use a small helper file which defines a namespaced `Page` class (https://github.com/silverstripe/silverstripe-framework/issues/5844), I don't see the need for this PR any more. There's going to be many other use cases where `new Page()` is used directly, and mandating `Injector` because of a relatively obscure API quirk for this particular instanciation feels a bit dirty. `Injector` should only be used for "services" in the wider sense - but not for every core object (e.g. `Page`).

Issue #1612, opened 2016-09-14
Double nested alert/message in history area


clarkepaul:
![image](https://cloud.githubusercontent.com/assets/555033/18501132/5074d920-7aa2-11e6-8a98-41db0f4ef26b.png)

tractorcow:
Fix at https://github.com/silverstripe/silverstripe-cms/pull/1646

PR #1646, opened 2016-10-12
Fix double nested alert message

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1612

I think this might be caused by a deeper bug in `ViewableData_Customised`, but it's abstract for me to get into. Maybe @tractorcow would be interested

### DIFF
--- a/code/Controllers/CMSPageHistoryController.php
+++ b/code/Controllers/CMSPageHistoryController.php
@@ -16,6 +16,7 @@ use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 use SilverStripe\ORM\Versioning\Versioned;
 use SilverStripe\Security\Security;
+use SilverStripe\View\ArrayData;
 use SilverStripe\View\ViewableData;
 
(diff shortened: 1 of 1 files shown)


code/Controllers/CMSPageHistoryController.php
```diff
@@ -174,8 +175,8 @@ public function getEditForm($id = null, $fields = null, $versionID = null, $comp
 		}
 
 		$fields->addFieldToTab('Root.Main',
-			new LiteralField('CurrentlyViewingMessage', $this->customise(array(
-				'Content' => DBField::create_field('HTMLFragment', $message),
+			new LiteralField('CurrentlyViewingMessage', ArrayData::create(array(
+                'Content' => DBField::create_field('HTMLFragment', $message),
```
dhensby: indentation has changed to spaces :/

dhensby:
@flamerohr if you make minor indentation amends, can you squash them, please? I've done it for you this time.

This should be merged on green

flamerohr:
Sure thing, will keep that in mind

Issue #1619, opened 2016-09-21
Toolbar extends across two different panels, they should be separated into respective panels


clarkepaul:
Acceptance criteria:

 * The north panel with the headers is split into two separate headers
 * The design implement is sufficient for 4.0-stable (no "short-term quick fix")

Recommend wait until after flexbox implementation is done

----

The top toolbar in the Pages section spans across the whole top but it would be better if they were separated out into the Pages panel and the Edit panel. This is one of the things which should be done before making the CMS work with Flex-box for cleanliness.
![pasted_image_21_09_16__5_09_pm](https://cloud.githubusercontent.com/assets/555033/18698550/52f99190-801f-11e6-8fbb-8410cab6f6f5.png)

sminnee:
It wouldn't surprise me if this was easier to do alongside or soon after flexbox, rather than before, as it's not worth trying to get it working with jLayout.

clarkepaul:
Yeah if I had dev assistance it could be done at the same time. I have had a conversation with @wilr  about this and he said he might be able to find the time to help, although there was no time pressure then. If this can be brought into the "Next" workstream and then one of the SilverStripe team could pick it up then great.

@wilr perhaps leave a comment here if you have an interest otherwise we would probably add to our upcoming stream of work. Cheers

wilr:
Yeah I was interested in fixing it up but sounded like @chillu wanted to remove jlayout completely which would be a bigger piece of work and change approach for this particular one. Happy to work on getting jlayout removed or implementing this in jlayout whichever way is best decided to go. My initial thought is its too hard to drop jLayout for 4.0.

chillu:
@wilr We're pretty far along removing jLayout, being worked on by @flamerohr as we speak :) https://github.com/silverstripe/silverstripe-framework/issues/6012 (there's five parts in five issues). You could help out with some peer review maybe?

flamerohr:
@wilr jLayout removal work is up, using flexbox now! :D (listed the pull requsts in @chillu's link)
We were looking doing this toolbar one after the flexbox work is merged, so we don't double up with conflicts all over the place.

flamerohr:
Pull requests that splits the toolbar to respective panels:
https://github.com/silverstripe/silverstripe-framework/pull/6189
https://github.com/silverstripe/silverstripe-cms/pull/1651

PR #1651, opened 2016-10-17
Split top toolbar for Pages section

flamerohr:
Requires https://github.com/silverstripe/silverstripe-framework/pull/6189

For more info check out https://github.com/silverstripe/silverstripe-cms/issues/1619


tractorcow:
Oh dear, this is what it looks like when I navigate from the pages section to editing a page.

image

flamerohr:
Hmm, trying to find what's causing this... quite a lot of magic.
I've found the bug to be related to if the Tab content was loaded once already, it won't load `.tabs()` on it again...

flamerohr:
Fixed, due to a crazy jQueryUI magic...

PR #1656, opened 2016-10-20
Correcting an issue where page controller methods

nglasl:
would end up with the error page response code, causing them to not work. This code only looks to be causing issues from what I can see.

An example is an autocomplete that is populated using a method on page controller. The moment you end up on a 404 error page, the autocomplete will attempt to populate and end up with a 404 response.

---

The only thing I noticed when removing this, is that hitting an error page **directly** ends up with a 200 response code (as opposed to the error page response code). Which is technically correct though, since this page **was** actually found. Hitting a URL that doesn't exist, however, continues to throw a 404.

The current state is quite nasty, and I've noticed it affecting a number of production sites, so it would be good to get this bug corrected across 3.X and 4.X moving forwards. Thoughts?

### DIFF
--- a/code/model/ErrorPage.php
+++ b/code/model/ErrorPage.php
@@ -323,22 +323,5 @@ class ErrorPage extends Page {
  * @package cms
  */
 class ErrorPage_Controller extends Page_Controller {
-
-	/**
-	 * Overload the provided {@link Controller::handleRequest()} to append the
-	 * correct status code post request since otherwise permission related error
-	 * pages such as 401 and 403 pages won't be rendered due to
-	  * {@link SS_HTTPResponse::isFinished() ignoring the response body.
-	 *
-	 * @param SS_HTTPRequest $request
-	 * @param DataModel $model
-	 * @return SS_HTTPResponse
-	 */
-	public function handleRequest(SS_HTTPRequest $request, DataModel $model = NULL) {
-		$body = parent::handleRequest($request, $model);
-		$this->getResponse()->setStatusCode($this->ErrorCode);
-
-		return $this->getResponse();
-	}
 }
 
--- a/tests/model/ErrorPageTest.php
+++ b/tests/model/ErrorPageTest.php
@@ -64,13 +64,19 @@ class ErrorPageTest extends FunctionalTest {
 	public function testBehaviourOf403() {
 		$page = $this->objFromFixture('ErrorPage', '403');
 		$page->publish('Stage', 'Live');
-		
-		$response = $this->get($page->RelativeLink());
-		
-		$this->assertEquals($response->getStatusCode(), '403');
-		$this->assertNotNull($response->getBody(), 'We have body text from the error page');
+
+		try {
+			$controller = singleton('ContentController');
+			$controller->httpError(403);
+			$this->fail('Expected exception to be thrown');
+		}
+		catch(SS_HTTPResponse_Exception $e) {
+			$response = $e->getResponse();
+			$this->assertEquals($response->getStatusCode(), '403');
+			$this->assertNotNull($response->getBody(), 'We have body text from the error page');
+		}
 	}
-	
+
 	public function testSecurityError() {
 		// Generate 404 page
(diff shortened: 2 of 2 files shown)


wilr:
Have you got sample code for your autocompleter? What is causing that code to hit ErrorPage::handleRequest(). Could the autocomplete be causing HTTP requests? In which case that could handle errors more gracefully.

nglasl:
The autocomplete is just a simple get objects, followed by a JSON encode for an AJAX request. The same problem also exists for a simple form submission.

I don't think it's really a matter of handling errors more gracefully, because the autocomplete still needs to be populated no differently than a page that actually exists. I guess my reasoning behind this change is more that I'm trying to correct the underlying issue. This is likely to be affecting many production websites as we speak :)

dhensby:
Can you squash these commits, please?

dhensby:
I *think* this seems reasonable, though removing a test is a big red flag here. how does it affect behaviour of error page?

I suppose *really* `ErrorPage` page types shouldn't respond to their URL at all, they aren't "real" pages and they shouldn't be accessibly by a URL they are just used to render a page when an error code response is made...

dhensby:
OK - this definitely can't go into 3.4 or even 3. Unfortunately core relies on error pages responding to requests by URL to generate static caches of them, which means we can't turn this off.

nglasl:
Okay, so I've put the test case back in, but updated it so it's no longer looking for the page directly. This now reflects a proper use case. The commits have also been squashed :)

This does not stop error pages from responding to requests by URL, it will just come back with a 200 response code if you happen to do so. Which is technically correct, since the page *was* found. I agree with your comment, in that they probably shouldn't respond to a request by URL, but maybe that's something that can be looked at further down the track, separate to this.

dhensby:
@silverstripe/core-team thoughts please?

We can see there's a change in behaviour here - should it go into 3?

tests/model/ErrorPageTest.php
```diff
@@ -64,13 +64,18 @@ public function testBehaviourOfShowInMenuAndShowInSearchFlags() {
...
 		$page->publish('Stage', 'Live');
-		
-		$response = $this->get($page->RelativeLink());
-		
-		$this->assertEquals($response->getStatusCode(), '403');
-		$this->assertNotNull($response->getBody(), 'We have body text from the error page');
+
+		try {
+			$controller = singleton('ContentController');
+			$controller->httpError(403);
```
tractorcow: After this line put a `$this->fail('Expected exception to be thrown')` as well. :)

code/model/ErrorPage.php
```diff
@@ -323,22 +323,5 @@ static public function get_static_filepath() {
...
-	 * Overload the provided {@link Controller::handleRequest()} to append the
-	 * correct status code post request since otherwise permission related error
-	 * pages such as 401 and 403 pages won't be rendered due to
-	  * {@link SS_HTTPResponse::isFinished() ignoring the response body.
-	 *
-	 * @param SS_HTTPRequest $request
-	 * @param DataModel $model
-	 * @return SS_HTTPResponse
-	 */
-	public function handleRequest(SS_HTTPRequest $request, DataModel $model = NULL) {
```
tractorcow: This is strictly a change in behaviour, but I think it's ok for 3 branch.

tests/model/ErrorPageTest.php
```diff
@@ -64,13 +64,18 @@ public function testBehaviourOfShowInMenuAndShowInSearchFlags() {
...
+		}
+		catch(Exception $e) {
```
tractorcow: I'd put `SS_HTTPResponse_Exception` to be more explicit.

tractorcow:
I left some review and requested some fixes to the tests.

tests/model/ErrorPageTest.php
nglasl: Fixed, cheers!

nglasl:
Cheers for the feedback :)

I have corrected those two things and squashed commits.

PR #1661, opened 2016-10-24
Simpler test run

sminnee:
This simplifies the test run in a similar manner to my work on framework.


tests/fixtures/Page.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\CMS\Controllers\ContentController;
+
+class Page extends SiteTree
```
tractorcow: Isn't this going to cause issues when `mysite` is installed with another Page.php? It'll also cause additional IDE and code inspection issues.

Suggested change: Rename to non-php extension, and perform a file copy in bootstrap.php if it detects no `$project` folder.

sminnee: The copy command is manual in `.travis.yml`; the travis test run will never make use of a mysite folder. It's kept in a manifest-excluded folder to prevent conflicts with other code.

I could perform the copy in bootstrap.php instead of travis.yml, although detecting if $project is set will get a chicken-egg bug: you'd need the manifest to be built to check that value, and the file needs to be copied before the manifest is built.

tractorcow: Could we at least rename the file in source control to a non-php extension so it doesn't interfere with code inspection? It's going to affect development when Page.php is defined elsewhere.

sminnee: Does it mess with PHPStorm?

sminnee: Will `Page.php.fixture` be okay?

tractorcow: Yes that would be fine. :)

sminnee:
Issues I can see with this:

 * `Page.php` should be renamed to `Page.php.fixture`
 * The Page.php class doesn't seem to be picked up by the manifest in the travis run. This may be because travis is using cached /tmp file content—you could try a `mkdir silverstripe-cache` call... :/
 * Even with that fixed I'm still getting test failures locally (details to follow)

sminnee:
Test failures when running locally

```
There were 2 errors:

1) CMSMainTest::testGetRecord
LogicException: The current member does not have permission to publish this ChangeSet.

/Users/sminnee/Projects/silverstripe-cms/framework/ORM/Versioning/ChangeSet.php:111
/Users/sminnee/Projects/silverstripe-cms/framework/ORM/Versioning/Versioned.php:1473
/Users/sminnee/Projects/silverstripe-cms/framework/Core/CustomMethods.php:71
/Users/sminnee/Projects/silverstripe-cms/tests/controller/CMSMainTest.php:262
/Users/sminnee/Projects/silverstripe-cms/vendor/phpunit/phpunit/phpunit:47

2) SiteTreeActionsTest::testActionsReadonly
LogicException: The current member does not have permission to publish this ChangeSet.

/Users/sminnee/Projects/silverstripe-cms/framework/ORM/Versioning/ChangeSet.php:111
/Users/sminnee/Projects/silverstripe-cms/framework/ORM/Versioning/Versioned.php:1473
/Users/sminnee/Projects/silverstripe-cms/framework/Core/CustomMethods.php:71
/Users/sminnee/Projects/silverstripe-cms/tests/model/SiteTreeActionsTest.php:36
/Users/sminnee/Projects/silverstripe-cms/vendor/phpunit/phpunit/phpunit:47

--

There were 3 failures:

1) ErrorPageTest::testStaticCaching
Failed asserting that a NULL is not empty.

/Users/sminnee/Projects/silverstripe-cms/tests/model/ErrorPageTest.php:103
/Users/sminnee/Projects/silverstripe-cms/vendor/phpunit/phpunit/phpunit:47

2) RedirectorPageTest::testReflexiveAndTransitiveInternalRedirectors
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'vendor/redirection-dest/'
+'vendor/vendor/redirection-dest/'

/Users/sminnee/Projects/silverstripe-cms/tests/model/RedirectorPageTest.php:53
/Users/sminnee/Projects/silverstripe-cms/vendor/phpunit/phpunit/phpunit:47

3) SiteTreeActionsTest::testActionsPublishedRecord
Failed asserting that null is not null.

/Users/sminnee/Projects/silverstripe-cms/tests/model/SiteTreeActionsTest.php:91
/Users/sminnee/Projects/silverstripe-cms/vendor/phpunit/phpunit/phpunit:47
```

tractorcow:
@sminnee I believe those issues are to do with unresolved errors upstream, not related to your changes; Permissions to do with changesets were changed slightly when we made Page::publishRecursive() use changesets.

composer.json
```diff
@@ -18,12 +18,16 @@
   "require": {
     "php": ">=5.5.0",
     "composer/installers": "*",
-    "silverstripe/framework": "^4.0",
-    "silverstripe/reports": "^4.0",
-    "silverstripe/siteconfig": "^4.0"
+    "silverstripe/framework": "^4.0-alpha3",
```
tractorcow: I've identified that once alpha3 is tagged, these builds won't install dev-master of framework anymore. There's a risk that cms builds will be stuck on old framework releases.

I suggest to change prefer-stable back to false, but leave it as true in the silverstripe/installer.

sminnee: This works:

```diff
diff --git a/composer.json b/composer.json
index 49fc956..2da2ffc 100644
--- a/composer.json
+++ b/composer.json
@@ -18,9 +18,9 @@
   "require": {
     "php": ">=5.5.0",
     "composer/installers": "*",
-    "silverstripe/framework": "^4.0-alpha3",
-    "silverstripe/reports": "^4.0-alpha3",
-    "silverstripe/siteconfig": "^4.0-alpha3"
+    "silverstripe/framework": "^4.0.9999-dev",
+    "silverstripe/reports": "^4.0.9999-dev",
+    "silverstripe/siteconfig": "^4.0.9999-dev"
   },
   "require-dev": {
     "phpunit/PHPUnit": "~4.8",
```

sminnee: Or this:

```diff
iff --git a/composer.json b/composer.json
index 49fc956..6c79dbc 100644
--- a/composer.json
+++ b/composer.json
@@ -18,9 +18,9 @@
   "require": {
     "php": ">=5.5.0",
     "composer/installers": "*",
-    "silverstripe/framework": "^4.0-alpha3",
-    "silverstripe/reports": "^4.0-alpha3",
-    "silverstripe/siteconfig": "^4.0-alpha3"
+    "silverstripe/framework": "^4.0@dev",
+    "silverstripe/reports": "^4.0@dev",
+    "silverstripe/siteconfig": "^4.0@dev"
   },
   "require-dev": {
     "phpunit/PHPUnit": "~4.8",
@@ -40,6 +40,5 @@
     },
     "classmap": ["tests/behat/"]
   },
-  "prefer-stable": true,
-  "minimum-stability": "dev"
+  "minimum-stability": "stable"
 }
```

tractorcow: 👍 for the second option; works well.

tractorcow:
I've finished my local testing; Pushed up a new config that should have workin behat + phpunit tests.

tractorcow:
Framework changes at https://github.com/silverstripe/silverstripe-framework/pull/6246

However, I have hacked the CMS build with temp code which should allow it to run before the framework changes are merged.

sminnee:
@tractorcow let's get 6246 merged so that we don't need a hacked CMS build

tractorcow:
Once it's green and proves it can pass CI ?

tests/bootstrap.php
```diff
@@ -0,0 +1,20 @@
...
+if (file_exists(FRAMEWORK_PATH . '/tests/bootstrap/init.php')) {
+	require __DIR__ . '/bootstrap/init.php';
+	require FRAMEWORK_PATH . '/tests/bootstrap/init.php';
+	require FRAMEWORK_PATH . '/tests/bootstrap/cli.php';
+	require FRAMEWORK_PATH . '/tests/bootstrap/environment.php';
+	require __DIR__ . '/bootstrap/mysite.php';
+	require FRAMEWORK_PATH . '/tests/bootstrap/mysite.php';
+	require FRAMEWORK_PATH . '/tests/bootstrap/phpunit.php';
+} else {
+	// @todo delete once framework changes are merged
```
sminnee: Let's get rid of this before we merge.

tests/bootstrap/mysite.php
```diff
@@ -0,0 +1,18 @@
...
+	mkdir($projectPath, 02775);
+	mkdir($projectPath.'/code', 02775);
+	mkdir($projectPath.'/_config', 02775);
+	copy(__DIR__.'/fixtures/Page.php.fixture', $projectPath . '/code/Page.php');
+	require_once $projectPath . '/code/Page.php';
+}
+
+if (file_exists(FRAMEWORK_PATH . '/tests/bootstrap/mysite.php')) {
+	require FRAMEWORK_PATH . '/tests/bootstrap/mysite.php';
+} else {
```
sminnee: Let's get rid of this before we merge.

tests/bootstrap/init.php
```diff
@@ -0,0 +1,15 @@
+<?php
```
sminnee: I think we can probably just reference `require FRAMEWORK_PATH . '/tests/bootstrap/init.php';` directly from 
`tests/behat/serve-bootstrap.php` and `tests/bootstrap.php`.

tests/bootstrap/mysite.php
tractorcow: Yep, just temp while we get behat green. :)

tractorcow:
The issue I'm running into is bootstrap_file being included after Core.php in behat-extension; It needs to be shifted to just before.

https://github.com/silverstripe/silverstripe-behat-extension/commit/2c4eb2050db7b4e97005ff734ae37ff0516bbc2d attempted to address this, however Core.php is actually included in two places in behat-extension, so this is too late.

sminnee:
Can you fix behat extension?

tractorcow:
I'm going to bootstrap before behat-extension for now in a separate `php ./tests/bootstrap/mysite.php` call in `.travis.yml`. behat-extension is troublesome to make reliable enough....

tractorcow:
Fix to postgres at https://github.com/silverstripe/silverstripe-postgresql/pull/59

tractorcow:
ALMOST THERE! :D

tractorcow:
Need to fix behat request hostname missing; maybe later. :)

tractorcow:
I've removed a @todo test that wasn't being run, and was previously polluting the output.

I suggest restarting travis a few times to make sure that the tests are stable, since I suspect that there might be an occasional failure (race condition?).

I've done about as much work as I think I can for the time being. :)

PR #1672, opened 2016-10-31
FIX: Remove unnecessary tests

sminnee:
Removed two test combinations from our travis builds, to reduce the
testing blockages.
- SQLite isn’t recommended for production and isn’t necessary on every
  PR / build
- We don’t need two separate PDO test runs (php 7 and 5.5)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,13 +22,9 @@ env:
 matrix:
   include:
     - php: 5.5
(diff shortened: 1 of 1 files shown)


kinglozzer:
Looks like this was sorted out in https://github.com/silverstripe/silverstripe-cms/pull/1801

PR #1675, opened 2016-11-04
FIX Page History 'Comparing versions' banner missing

mikenz:
The 'Comparing versions' banner was inserted before the "Title" field. If there is no "Title" field the banner is missing. If the "Title" field isn't the first field then the banner is in the incorrect place.

This change always adds it to the top of the Main tab


dhensby:
Seems like a very sensible fix. I don't see any reason this couldn't go into 3.4 branch

PR #1677, opened 2016-11-04
BUGFIX: Fixed issue on windows where the CMS_DIR constant would be set containing a backslash

UndefinedOffset:
This pull request fixes an issue found in SilverStripe 4 Alpha 3 where the cms would hard crash on windows because the CMS_DIR constant was being set containing a backslash. This seems to be caused by the trim call having a hard coded forward slash.


dhensby:
thanks

PR #1684, opened 2016-11-15
API Update tests to reflect renamed services

tractorcow:
Requires https://github.com/silverstripe/silverstripe-framework/pull/6315.

A later PR to namespace all CMS tests will be pushed up at a later date.


tractorcow:
Yay :D

PR #1687, opened 2016-11-22
FIX: content authors unable to duplicate top-level pages (fixes #1685)

### LINKED ISSUE #1685: Users without full admin rights unable to duplicate root level pages
I noticed this on 3.4, but also tried a fresh install of 3.2 and it has the same behaviour.

It seems that CMS users who have not been granted full admin rights are unable to duplicate a root-level page. Creating one works fine, but when duplicating there is a check in SiteTree.php `$parent->canAddChildren($member);` which ultimately get to a line in the *getIsDeletedFromStage* function `if(!$this->ID) return true;` Because the root pages parent ID is 0, this then causes canAddChildren to return false.

Speaking on IRC with @kinglozzer he suggested changing the if statement to ` if($parent && $parent->exists()) {` which has fixed the issue for me.

kinglozzer:


kinglozzer:
UGH whitespace 😢. Need me to fix it?

dhensby:
sorry to say this won't make the next 3.2 release, though :/

kinglozzer:
That’s fine, I’ll post the workaround in the issue (for reference, https://github.com/silverstripe/silverstripe-cms/issues/1685)

tractorcow:
Nice fix!

PR #1688, opened 2016-11-29
Initial steps towards PSR-2

sminnee:
This pull request addresses the phpcbf-provided PSR-2 compliance changes, introduces `composer run-script lint` and adds it as part of the travis build.


tractorcow:
I ran into this issue too... you need to do the `pyrus` install in a php 5.6 since it doesn't yet work in php7.

Everything else looks good to me. :)

sminnee:
Right... I wonder if that's a compelling reason to use composer over pyrus to install phpcs. If I understand correctly, developers running php7 will do the same thing.

sminnee:
It looks like we can make `phpcs.xml.dist` in our folder root and then we don't need any `standard` argument. There's also a `<file>./path/to/directory</file>` argument so that we don't need to pass src and tests on the command line.

I'll refactor this commit and also apply an update to framework, as I think this is a much better approach. It also means you shouldn't need to fiddle with PHPStorm.

tractorcow:
+1 sounds good.

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-cms/pull/1727

PR #1692, opened 2016-12-05
BUG Fix localisation issues in CMS

tractorcow:
This should resolve the majority of behat regressions caused within the last release.

Should fix most issues in this pull request only, but remainder of regressions fixed with https://github.com/silverstripe/silverstripe-framework/pull/6365

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -1317,10 +1317,14 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 			$token->updateFieldSet($fields);
 			$tokenField = $fields->First();
 			$tokenHtml = ($tokenField) ? $tokenField->FieldHolder() : '';
+			$publishAllDescription = _t(
+				'CMSMain.PUBALLFUN2',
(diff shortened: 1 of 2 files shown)


tractorcow:
Putting impact/high back; We need to get behat tests green for current pull requests.

kinglozzer:
silverstripe/silverstripe-framework#6365 merged, restarting behat...

kinglozzer:
Hmm, we still have [a failure](https://travis-ci.org/silverstripe/silverstripe-cms/jobs/181486238#L575). Perhaps it’s failing the `Then I should not see a ".ss-uploadfield-files .ss-uploadfield-item" element` check from the previous scenario? I can’t see why else it would fail immediately before even trying anything!

tractorcow:
It seems to be from this.

image

Wish we had artifacts working...

dhensby:
Wish no more!

kinglozzer:
Huh? Aren’t behat tests still failing? @dhensby

dhensby:
> Edit: Actually I guess this is a step closer, so that probably didn’t need to block this PR :)

yep

PR #1698, opened 2016-12-20
add warning about child pages being unpublished

andrewandante:
After multiple reports of content authors being unaware that unpublishing a page will unpublish all the children of that page also, this commit will offer a different warning to the user if there are children of the page.

Not sure if I've done the `lang` stuff correctly - any advice would be appreciated.

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -2217,6 +2217,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	 */
 	public function getCMSActions() {
 		$existsOnLive = $this->getExistsOnLive();
+		$numChildren = count($this->getDescendantIDList());
 
 		// Major actions appear as buttons immediately visible as page actions.
 		$majorActions = CompositeField::create()->setName('MajorActions')->setTag('fieldset')->addExtraClass('ss-ui-buttonset noborder');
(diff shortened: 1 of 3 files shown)


phalkunz:
Similar changes: https://github.com/silverstripe/silverstripe-cms/pull/1774/files

code/model/SiteTree.php
```diff
@@ -2217,6 +2217,7 @@ public function fieldLabels($includerelations = true) {
 	 */
 	public function getCMSActions() {
 		$existsOnLive = $this->getExistsOnLive();
+		$numChildren = count($this->getDescendantIDList());
```
tractorcow: My concern with this is that selecting all ids just to get a count is a heavily intensive process and could slow down some sites significantly. Can we use `numChildren` instead, and have a less precise (but faster) behaviour?

In master (https://github.com/silverstripe/silverstripe-cms/pull/1774) I think there is a similar problem, which I might go back to and address.

andrewandante:
Will try to re-do this as a backport of #1774  - keep it consistent.

tractorcow:
Just to catch up on this old PR, my feedback still stands. Let me know if you want me to take any action on this.

andrewandante:
@tractorcow have ported it over to #2087

dhensby:
closing in favour of #2087

PR #1699, opened 2016-12-20
Refresh CMS UI buttons to new flat style and bootstrap classes

wilr:
Also see https://github.com/silverstripe/silverstripe-framework/pull/6434

Requires https://github.com/silverstripe/silverstripe-cms/pull/1699

### DIFF
--- a/client/src/legacy/CMSMain.EditForm.js
+++ b/client/src/legacy/CMSMain.EditForm.js
@@ -377,51 +377,63 @@ $.entwine('ss', function($){
 	 * Enable save buttons upon detecting changes to content.
 	 * "changed" class is added by jQuery.changetracker.
 	 */
-	$('.cms-edit-form.changed').entwine({
-		onmatch: function(e) {
-			var save = this.find('button[name=action_save]');
+  $('.cms-edit-form.changed').entwine({
+    onmatch: function(e) {
+      var save = this.find('button[name=action_save]');
+
+      if(save.attr('data-text-alternate')) {
+        save.attr('data-text-standard', save.text());
+        save.text(save.attr('data-text-alternate'));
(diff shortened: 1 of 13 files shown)


client/src/legacy/CMSMain.EditForm.js
```diff
@@ -377,17 +377,65 @@ $.entwine('ss', function($){
...
+      }
+
+      if(publish.attr('data-btn-alternate')) {
+        publish.attr('data-btn-standard', publish.attr('class'));
+        publish.attr('class', publish.attr('data-btn-alternate'));
+      }
+
+      publish
+        .removeClass('btn-secondary-outline')
+        .addClass('btn-primary');
```
flamerohr: Took me a while to realise it was working with the `save` and `publish` button with the same actions... maybe select both in the `.find()` and do a `.forEach()` to reduce the repetition?

client/src/legacy/CMSMain.EditForm.js
```diff
@@ -377,17 +377,65 @@ $.entwine('ss', function($){
...
+
+      var publish = this.find('button[name=action_publish]');
+
+      if(publish.attr('data-text-standard')) {
+        publish.text(publish.attr('data-text-standard'));
+      }
+
+      if(publish.attr('data-btn-standard')) {
+        publish.attr('class', publish.attr('data-btn-standard'));
+      }
```
flamerohr: As above, perhaps select both and use `.forEach()`

tractorcow:
Ok, there are still some issues with cms tests.

I've fixed these with https://github.com/silverstripe/silverstripe-framework/pull/6475. Merge that PR and restart these to confirm.

PR #1702, opened 2016-12-29
BUILD Run PHP 7.1 as default 7.x

robbieaverill:


dhensby:
Why do we need 2 pgsql test suites?

robbieaverill:
I suppose we don't - removing it now...

PR #1703, opened 2016-12-29
API Allow controller discovery without underscore (PSR-2 compliance)

robbieaverill:
This pull request changes the default controller class lookup from `My_Controller` to `MyController`. It still supports `My_Controller` loading, however it will issue a deprecation notice from 4.0.0 onwards.

Issue: silverstripe/silverstripe-framework#6403

Blocks: silverstripe/silverstripe-framework#6446

### DIFF
--- a/tests/model/SiteTreeTest.php
+++ b/tests/model/SiteTreeTest.php
@@ -17,7 +17,6 @@ use SilverStripe\Control\Session;
 use SilverStripe\View\Parsers\ShortcodeParser;
 use SilverStripe\Control\Director;
 use SilverStripe\i18n\i18n;
-use SilverStripe\Dev\Deprecation;
 use SilverStripe\Dev\SapphireTest;
 use SilverStripe\Dev\TestOnly;
 use SilverStripe\View\Parsers\HTMLCleaner;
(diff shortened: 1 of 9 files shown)


tests/model/SiteTreeTest.php
```diff
@@ -1334,25 +1333,73 @@ public function testCanPublish() {
...
+	 * an exception is thrown will not allow further logic to execute after the exception is thrown.
+	 */
+	public function testGetControllerNameWithUnderscoresIsSupported()
+	{
+		$defaultEnabled = Deprecation::get_enabled();
+		Deprecation::set_enabled(true);
+		Deprecation::notification_version('4.0.0');
+
+		// Catch user_errors thrown by deprecation
+		set_error_handler(
```
dhensby: PHPUnit should automatically convert Errors/Warnings/Notices to exceptions and you can then use the docblock notation (`@expectedException PHPUnit_Framework_Error_Deprecated`)

robbieaverill: I don't think it does automatically, needs an arg set in the config to raise errors etc as exceptions. Adding now.

robbieaverill:
@dhensby I've updated the configuration and added the assertions to the doc block. I was initially a little weary of changing the Deprecation level to 4.0.0 and not changing it back since the exception halts further execution in the test method, but it doesn't appear to affect any of the rest of the suite so it's probably alright for now. I suppose it should be mocked if needed in a test anyway.

dhensby:
Hmm - that's a good point - and the actual assertion in that test won't be reached because the exception will be thrown instead...

Maybe that change wasn't for the better (sorry)? It may not even be worth asserting the deprecation notice is thrown - that's basically just testing that `Deprecation` works which isn't the point of this test...

robbieaverill:
@dhensby no problem. I'll remove the commit and test for the deprecation.

tractorcow:
@robbieaverill you can also use setup / teardown to restore any previous deprecation settings, so that you can modify them at-will in your individual tests.

See ControllerTest.php for example

```
public function setUp()
    {
        parent::setUp();
        $this->depSettings = Deprecation::dump_settings();
    }

    public function tearDown()
    {
        Deprecation::restore_settings($this->depSettings);
        parent::tearDown();
    }
```

tractorcow:
One more change in removing the underscore from the controller name is a small fix needed to SSViewer.php


```
// If the class is "Page_Controller", look for Page.ss
if (stripos($class, '_controller') !== false) {
    $templates[] = str_ireplace('_controller', '', $class) . $suffix;
}
```

We should probably support `MyPageController` mapping to `MyPage.ss` templates as well.

I would suggest, in order to get the best of both worlds by breaking as few backwards compatibilyt features as possible:
 - Case insensitive seach for `_controller` as-is.
 - Case sensitive seach for `Controller` on the END of the class name string (not simply anywhere). Comment as psr-2 compatible class lookup. :)

robbieaverill:
Sounds good. I'll make the changes tomorrow morning.

I'll put the deprecation tests back in I think- may as well have them

robbieaverill:
@tractorcow actually I covered the SSViewer change in silverstripe/silverstripe-framework#6446 - doesn't have a PSR-2 comment though :-)

tractorcow:
Sorry I didn't check your framework PR. :) Just trying to rush through my holiday emails!

robbieaverill:
All good!

tractorcow:
Added a comment on https://github.com/silverstripe/silverstripe-framework/pull/6446#pullrequestreview-15192815. You can ignore my earlier suggestion. :)

dhensby:
Whilst we aren't deprecating usercode, we are deprecating support for looking it up for them. If they want to use `_Controller` instead of `Controller` then they'll need to overload the `getControllerName` method on their object(s)....

robbieaverill:
Did I do something silly? This change is supposed to support both methods while issue deprecation on the underscored version. You shouldn't need to overload the method for it to work though?

dhensby:
I think what you have is perfect

robbieaverill:
Rebased onto latest master

PR #1704, opened 2017-01-03
ENHANCEMENT: Use SiteTree instead of Page class in more places.

mikenz:
This allows developers to extend SiteTree directly, instead of Page.

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -59,7 +59,6 @@ use SilverStripe\Security\SecurityToken;
 use SilverStripe\View\ArrayData;
 use SilverStripe\View\Requirements;
 use Translatable;
-use Page;
 use Zend_Cache;
 use InvalidArgumentException;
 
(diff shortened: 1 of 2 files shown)


dhensby:
I think this is sensible - we shouldn't be making references to classes that are outside the scope of this module, ideally, though `ErrorPage` and the other default pages all do (so hard to eliminate completely)

tractorcow:
Developers should be extending Page, not SiteTree. It's intended to be the main "root" class in the hierarchy, in order to work around the issue with SiteTree not being user-editable. Although I don't see the harm in the changes made in this particular PR, I fundamentally disagree with the suggested change in convention.

sminnee:
@tractorcow we'd likely review that in SS5. The necessity of the page class causes a lot of issues. But for SS4 the recommended pattern is for modules providing custom page types to extend from `Page` / `PageController`.

sminnee:
The two things that have been discussed to date are:

 * Modules stop extending Page and only provide content blocks. Under this model, site developers would still be free to make a custom Page class but it wouldn't be a necessity as nothing would sub-class page.
 * Modules extend SiteTree and side-wide modifications are done via extensions on SiteTree. Under this model, changes that a site developer makes in the Page class wouldn't be applied to other models.

dhensby:
I agree that we shouldn't encourage many classes to extend `SiteTree` but we also shouldn't be making assumptions about classes outside of our code base (I know, that's impossible given we extend `Page` for `ErrorPage` and others - but let's try to keep it to a minimum).

PR #1709, opened 2017-01-10
FIX: broken "Add new page here" context menu

colintucker:
Currently, the right-click "Add new page here" context menu for `SiteTree` is broken.  When a page type is selected from the context menu, the CMS does not receive the `ParentID` and `PageType` URL parameters.  This results in a regular `Page` being added to the root, rather than the selected page type being added to the selected parent.

The problem was traced to the double-escaping of ampersands in `SilverStripe\CMS\Controllers\CMSMain` and `CMSMain_TreeView.ss`.  Because output from the `LinkPageAdd` method is now properly escaped by SS4, the encoded `&amp;` is no longer necessary, and can be safely replaced with a regular `&`.  The context menu now functions correctly after the change.  👍


dhensby:
thanks

PR #1716, opened 2017-01-18
fixes load indicator position (fixes #1625)

### LINKED ISSUE #1625: URL segment loading indicator position in the wrong place
Wasn't quick enough to get a screenshot of it but edit the page name and click outside of the input to see indicator.
pasted_image_22_09_16__3_20_pm

hoKRa1:
The fix re-positions the load indicator to the left of the URL segment field when editing the page name and/or page URL. This prevents the indicator from being slightly overlapped by the URL segment field when loading page URL changes.
(Fixes #1625)


hoKRa1:
screen shot 2017-01-18 at 10 02 17 pm

See: #1716

hoKRa1:
screen shot 2017-01-18 at 10 02 17 pm

Here you go.

hoKRa1:
Hi @chillu,

Which browser are you using? I am using Chromium 55.0.2883.87 and Firefox 50.1.0. I do notice that your indicator is slightly more to the left than the indicator that's appearing on my screen:

screenshot from 2017-01-19 09-28-23

screenshot from 2017-01-19 09-29-53

tractorcow:
I've fixed this and pushed up a patch that works in a responsive manner.

The issue was not due to browser, but the fact that the background was fixed to the field container, and did not respond to the variable label size.

I've shifted this to the label, and adjusted the positioning to always sit on the right hand side of the label. It's not perfect, but it does ensure a consistent layout regardless of browser size.

tractorcow:
image

PR #1720, opened 2017-01-21
Changed init() visibility to protected

colintucker:
`Controller` has had it's `init()` method changed to `protected`, updating `Controller` subclasses in CMS to follow suit.


dhensby:
thanks - I probably should have done this when I made the original PR :/

PR #1722, opened 2017-01-22
API Add getFrontendControllerName - allows the controller to be configurable

### LINKED ISSUE #1721: Namespacing and configuring controllers for SiteTree subclasses
Currently, with the move to PSR-2 compliant code in v4, `Controller` subclasses should be separated into their own files, instead of being in the same file as their linked `SiteTree` subclasses.

The namespace changes also introduced in v4 allow for these controllers to exist in a different namespace from the model, however, currently they must exist in the same namespace in order for `SiteTree::getControllerName()` [to be able to locate them](https://github.com/silverstripe/silverstripe-cms/blob/3f53da050cecd47605575b4416abfc7fbe4f893c/code/Model/SiteTree.php#L2739-L2751).

This method could be overridden by subclasses to return a controller name in a different namespace, but it feels hacky, and would probably be better solved by adding a configuration option to `SiteTree` such as (courtesy @sminnee ):

```php
use App\Controllers\PageController;

class Page {
  private static $frontend_controller = PageController::class;
}
```

...and then modifying `SiteTree::getControllerName()` to look for this configuration option, with a fallback to the current `{$Page}Controller` and `{$Page}_Controller` (deprecated) functionality.

Thoughts?

robbieaverill:
This change will allow developers to set a configuration value for `frontend_controller` on a `SiteTree` instance to tell it to use a custom controller class, e.g. `Whatever\You\Want`.

If will fall back to the default behaviour if the config static is not set.

Resolves silverstripe/silverstripe-framework#6549 and resolves #1721

### DIFF
--- a/code/Controllers/ModelAsController.php
+++ b/code/Controllers/ModelAsController.php
@@ -37,7 +37,7 @@ class ModelAsController extends Controller implements NestedController {
 	 * @return ContentController
 	 */
 	public static function controller_for(SiteTree $sitetree, $action = null) {
-		$controller = $sitetree->getControllerName();
+		$controller = $sitetree->getFrontendControllerName();
 
 		if ($action && class_exists($controller . '_' . ucfirst($action))) {
 			$controller = $controller . '_' . ucfirst($action);
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -242,6 +242,15 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	private static $icon = null;
 
 	/**
+	 * Specify the fully qualfied class name for this model's controller. If not provided, the default behaviour
+	 * will be used (Fully\Qualified\PageController)
+	 *
+	 * @config
+	 * @var string
+	 */
+	private static $frontend_controller = null;
+
+	/**
 	 * @config
 	 * @var string Description of the class functionality, typically shown to a user
 	 * when selecting which page type to create. Translated through {@link provideI18nEntities()}.
@@ -2752,6 +2761,21 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 		}
 
 		return $controller;
+	}
+
+	/**
+	 * Find the frontend controller name, which can be the current fully qualified class name
+	 * with Controller or _Controller added to it within the same namespace, or a customised
+	 * filename as per the $frontend_controller configuration option.
+	 *
+	 * @return string
+	 */
+	public function getFrontendControllerName()
+	{
+		if ($customController = $this->config()->frontend_controller) {
+			return (string) $customController;
+		}
+		return $this->getControllerName();
 	}
 
 	/**
--- a/code/Model/VirtualPage.php
+++ b/code/Model/VirtualPage.php
@@ -455,4 +455,16 @@ class VirtualPage extends Page {
 		return parent::getControllerName();
 	}
 
+	/**
+	 * {@inheritDoc}
+	 */
+	public function getFrontendControllerName()
+	{
+		if ($copy = $this->CopyContentFrom()) {
+			return $copy->getFrontendControllerName();
+		}
+
+		return parent::getFrontendControllerName();
+	}
+
 }
--- a/tests/model/SiteTreeTest.php
+++ b/tests/model/SiteTreeTest.php
@@ -1351,6 +1351,21 @@ class SiteTreeTest extends SapphireTest {
 		$this->assertSame('SiteTreeTest_LegacyControllerName_Controller', $class->getControllerName());
 	}
 
+	/**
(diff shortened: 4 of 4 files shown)


kinglozzer:
@dhensby and I had discussed this when this was originally implemented, and we decided to leave it as we thought you can just override `getControllerName()` rather than add another static. I don’t object to this change (I can see that it could be useful to override this if a module provides a controller and you’d rather roll your own), but why the separate method? When would you ever _not_ want the controller that you’ve specified to be used?

robbieaverill:
Thoughts @sminnee?

I guess the benefit of the configuration property is less lines than a method override, but that also seems like a reasonable option too!

code/Controllers/ModelAsController.php
```diff
@@ -37,7 +37,7 @@ class ModelAsController extends Controller implements NestedController {
 	 * @return ContentController
 	 */
 	public static function controller_for(SiteTree $sitetree, $action = null) {
-		$controller = $sitetree->getControllerName();
+		$controller = $sitetree->getFrontendControllerName();
 
 		if ($action && class_exists($controller . '_' . ucfirst($action))) {
```
sminnee: Suggest we drop this functionality - AFAIK it's not heavily used.

code/Controllers/ModelAsController.php
```diff
@@ -37,7 +37,7 @@ class ModelAsController extends Controller implements NestedController {
 	 * @return ContentController
 	 */
 	public static function controller_for(SiteTree $sitetree, $action = null) {
-		$controller = $sitetree->getControllerName();
+		$controller = $sitetree->getFrontendControllerName();
```
sminnee: Suggest that this becomes `getFrontendController()` and includes the Injector call within that method.

code/Model/SiteTree.php
```diff
@@ -2755,6 +2764,21 @@ public function getControllerName() {
...
+	 * filename as per the $frontend_controller configuration option.
+	 *
+	 * @return string
+	 */
+	public function getFrontendControllerName()
+	{
+		if ($customController = $this->config()->frontend_controller) {
+			return (string) $customController;
+		}
+		return $this->getControllerName();
```
sminnee: Suggest `getControllerName()` is deprecated in favour of `getFrontendController()`. `getControllerName()` can return `get_class($this->getFrontEndController())` as deprecation shim, if needs be.

code/Controllers/ModelAsController.php
robbieaverill: Drop or deprecate?

code/Model/SiteTree.php
robbieaverill: This change would need to be made in `SSViewer::get_templates_by_class` as well. In terms of the existing functionality there (look for templates called `MyDataObject_?Controller`) - should we change this to look for a `->getFrontendController()` result, but fall back to the existing method if the class is not a `SiteTree` instance?

It seems to me that if we want to do away with the `_Controller` appending altogether, we should implement this logic higher up than `SiteTree`, probably in `DataObject` instead, then implement the same checks in `SSViewer`

code/Controllers/ModelAsController.php
dhensby: 4.0 is the time to drop it :)

Can put the deprecation warning in 3

code/Model/SiteTree.php
dhensby: Can't we do this the other way round, `getController` just does `Injector::inst()->create($this->getControllerName())`?

That seems more flexible to me, plus for templates don't we want the "controller name" and not the FQCN for the class the Injector instantiated?

dhensby:
I think @kinglozzer and I didn't take into account the full repercussions when implementing our solution. Whilst I don't really like the idea of having to define a FQCN for controllers on models, I think it offers the flexibility required when taking into account thirdparty code (modules) and being able to easily override that code.

One important thing to make sure works is if we have a module (eg: Blog module) and we want to override the `BlogController` that all the existing templates still work. This means if we change the class that is being used as the controller we don't actually change the class that the View *thinks* we are using - otherwise it'll start rendering with the wrong templates or failing to match the existing ones

kinglozzer:
> I think @kinglozzer and I didn't take into account the full repercussions when implementing our solution.

It was initially only implemented as a solution for problems with `VirtualPage` (namely so we could get rid of `VirtualPage_Controller`, and instead use the appropriate controller for the target page type), so it was literally just moving the 3.x behaviour from `ContentController` to `SiteTree`. Looks like I’m guilty of leaving it as “worry about that later”, sorry! https://github.com/silverstripe/silverstripe-cms/pull/1605#issuecomment-248909261

We need to weigh up what’s best practice against what makes life easier for SilverStripe developers here I think. Maybe we should brainstorm a few options and ping this over to the dev list for feedback?

robbieaverill:
Closing pending further discussion (on silverstripe/silverstripe-framework#6552)

PR #1724, opened 2017-01-24
Upgrade cms for new i18n backend

tractorcow:
Required for https://github.com/silverstripe/silverstripe-framework/pull/6558

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -85,7 +85,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
 	private static $menu_title = 'Edit Page';
 
-	private static $menu_icon_class = 'font-icon-sitemap'; 
+	private static $menu_icon_class = 'font-icon-sitemap';
 
 	private static $menu_priority = 10;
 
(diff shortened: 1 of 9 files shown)


.editorconfig
```diff
@@ -6,7 +6,7 @@
 charset = utf-8
 end_of_line = lf
 indent_size = 4
-indent_style = tab
+indent_style = space
```
mikenz: Yay! But was this really supposed to be in this PR?

tractorcow: We should be using psr2 now; I think some upgrades were missing from cms module though. I should run a formatter over it.

tractorcow: Oh look, we never merged it. =( https://github.com/silverstripe/silverstripe-cms/pull/1688

tractorcow: Rebuilt at https://github.com/silverstripe/silverstripe-cms/pull/1727. I'll revert from this PR.

tractorcow:
I've reverted any presumptuous psr2 reformatting on SiteTree.php; I'll wait until we merge the main psr2 PR instead :D

tractorcow:
Ok, I've updated the tests and fixed some german localisations for plurals. Should be passing now. :D

note: CMS pr can be merged ahead of the framework, since it only fixes problems, it doesn't RELY on the new api for it to work.

tractorcow:
This is ready to merge.

PR #1725, opened 2017-01-24
FIX: Javascript .addAttr() doesn't exist

mikenz:
When there's no templates that can be added in the root /admin/pages/add/ hits a JS error and never completes loading


dhensby:
d'oh - this change needs bundling into the dist file...

@chillu I thought the NPM test should catch when the bundled version hasn't been built?

dhensby:
I've built and pushed the files

PR #1728, opened 2017-01-27
CI Check dist files are correctly built

dhensby:
see https://github.com/silverstripe/silverstripe-framework/pull/6573


dhensby:
<3

PR #1734, opened 2017-02-08
Update CMS for new i18n API

tractorcow:
Requires https://github.com/silverstripe/silverstripe-framework/pull/6607


dhensby:
We're green :)

PR #1741, opened 2017-02-23
PSR-16 cache

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/6252

### DIFF
--- a/_config/cache.yml
+++ b/_config/cache.yml
@@ -1,0 +1,9 @@
+---
+After:
+  - '#corecache'
+---
+SilverStripe\Core\Injector\Injector:
+  Psr\SimpleCache\CacheInterface.CMSMain_SiteTreeHints:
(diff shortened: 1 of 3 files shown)


chillu:
@sminnee @tractorcow Thanks for merging! :) This merge broke the build, but not because of anything in the merge itself, but rather the fact that it's defaulting to `framework@4.0.0-alpha4`: https://travis-ci.org/silverstripe/silverstripe-cms/jobs/205416364. Given the constraint is set to `^4.0@dev` with `minimum-stability: stable`, I'd expect it to choose `4.0.x-dev` (we don't have `prefer-stable: true` set).

tractorcow:
Try removing `--prefer-dist` from composer install in cms tests `.travis.yml`. The `@dev` already overrides min-stability, but it doesn't override the prefer-dist.

PR #1742, opened 2017-02-24
Remove `ADMIN_THIRDPARTY_DIR . '/jquery/jquery.js'` requirement

oilee80:
As the framework's vendor.js has jQuery bundled in it already it is not required a second time

Causes issues where jQuery add ons being added to the wrong jQuery object


tractorcow:
However, contentcontroller doesn't require vendor.js, does it?

oilee80:
@tractorcow No but it does get included when in admin

tractorcow:
Yes, but when you aren't running contentcontroller in the admin (front end of the site) you'll end up with broken script dependencies.

tractorcow:
Sorry, it might be worth me pointing out that this can be used outside of the CMS too, since that's probably not clear. :)

However, it's more often used in the CMS, so maybe we need a conditional?

We could add a `['provides' => 'jquery']` to the options to ensure that requirements api only includes it if not already added.

chillu:
`ContentController` isn't instanciated in the CMS, only in the iframe that loads the page preview. In which case you do need the jquery dependency directly (outside of the CMS). Closing as invalid.

PR #1744, opened 2017-02-26
Fix issue with CI installing non-dev dependencies

tractorcow:


sminnee:
Ok this isn't ideal but good enough for now to fix the build.

PR #1748, opened 2017-02-27
FIX: address composer issues

sminnee:
We no longer need minimum-stability dev as the testsession,
behat-extension, siteconfig, and reports modules all have “@dev” on the
end of their framework dependencies now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,14 +20,17 @@
     "composer/installers": "*",
     "silverstripe/framework": "^4.0@dev",
     "silverstripe/reports": "^4.0@dev",
-    "silverstripe/siteconfig": "^4.0@dev"
+    "silverstripe/siteconfig": "^4.0@dev",
+    "silverstripe/asset-admin": "^1.0@dev"
   },
   "require-dev": {
     "phpunit/PHPUnit": "~4.8",
-    "silverstripe/behat-extension": "^2.1.0",
+    "silverstripe/behat-extension": "^2.3.1",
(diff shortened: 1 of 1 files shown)


tractorcow:
It's not working; Framework is getting installed as alpha tagged version, not dev. I guess we still do need this after all.

tractorcow:
Note: Treat closing this PR not as a dismissal... I still want a solution to this problem as well. Unfortunately this one doesn't seem to work. =(

sminnee:
I'm investigating this angle next: https://github.com/composer/composer/issues/6258

sminnee:
The underlying composer restriction will remain as-is: unless the root project decides to accept dev packages, no dev packages will appear.

tractorcow:
I'm not sure CMS should require asset-admin module at this stage. It still works fine without it.

dhensby:
if the `@dev` doesn't work on non-root composers we should remove them from here.

I agree with @tractorcow that we shouldn't require asset-admin if it's not needed and should move it to recommended package instead

tractorcow:
I think `@dev` only helps from a `show that we are aware of some unstable dependencies`.

We need to go over all CI and ensure we explicitly install dev dependencies rather than rely on composer.

I also need to go back and add some modules, such as silverstripe/config, to installer, so that users can check out non-stable versions.

tractorcow:
Working POC at https://github.com/silverstripe/silverstripe-framework/pull/6712

dhensby:
fixed at #6712

Issue #1746, opened 2017-02-27
Page "History" view renders HTML tags instead of diff view


chillu:
The diff views use `<ins>` and `<delete>`, which are being wrongly escaped now.

![image](https://cloud.githubusercontent.com/assets/111025/23377944/95b30e90-fd96-11e6-9511-561c05664c23.png)

Paul wants to redo the whole view, but for now we should fix the issue at hand (https://github.com/silverstripe/silverstripe-cms/issues/1493)

PR #1749, opened 2017-02-28
Fix history comparison fields

flamerohr:
so the fields no longer escape the `<ins>` and `<del>` tags injected by the Differencer in the History view for pages.

Original issue https://github.com/silverstripe/silverstripe-cms/issues/1746

Also added some unit tests for this.

### DIFF
--- a/code/Controllers/CMSPageHistoryController.php
+++ b/code/Controllers/CMSPageHistoryController.php
@@ -457,12 +457,12 @@ class CMSPageHistoryController extends CMSMain
             "ID" => $id,
             "Version" => $fromVersion,
         ));
-    
+
         // Comparison views shouldn't be editable.
         // As the comparison output is HTML and not valid values for the various field types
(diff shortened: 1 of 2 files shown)


code/Controllers/CMSPageHistoryController.php
```diff
@@ -475,4 +472,19 @@ public function Breadcrumbs($unlinked = false)
         $crumbs[0]->Title = _t('CMSPagesController.MENUTITLE', 'Pages');
         return $crumbs;
     }
+    
+    public function transformReadonly(FieldList $fields)
+    {
+        foreach($fields as &$field) {
```
tractorcow: The problem with iterating over FieldList directly is that it only loops over the top level of fields.

Try `foreach ($fields->dataFields() as $field) {`, which has the added benefit of selecting only data fields, so you don't need the CompositeField filter.

flamerohr: Thanks for fixing it :)

PR #1750, opened 2017-02-28
FIX Config regressions

tractorcow:
It turns out CMS relied on FIRST_SET for allowedChildren()

In a separate commit I've done some unit test class cleanup (rewrite from strings -> class constants, and other formatting improvements).


tractorcow:
Permission to merge from @sminnee :)

PR #1753, opened 2017-03-02
FIX function privacy level

andrewandante:
Prevents this on dev/build:

`PHP Fatal error:  Access level to SilverStripe\CMS\Model\RedirectorPageController::init() must be public (as in class PageController) in /vagrant/andante/cms/code/Model/RedirectorPageController.php on line 33`

### DIFF
--- a/code/Model/RedirectorPageController.php
+++ b/code/Model/RedirectorPageController.php
@@ -9,7 +9,7 @@ use PageController;
 class RedirectorPageController extends PageController
 {
 
-    protected function init()
+    public function init()
     {
         parent::init();
 


kinglozzer:
This should be protected I think, see https://github.com/silverstripe/silverstripe-installer/blob/master/mysite/code/PageController.php#L25

andrewandante:
Huh. Somehow I don't have a `PageController.php` - this occurred once I upgraded from alpha-4 to alpha-5.

Ah - I've found my own `Page.php` with PageController inside it (naughty) - and it's public. Will change that then. Carry on!

PR #1757, opened 2017-03-09
Fix non_virtual_fields being ignored

mikenz:
Fix non_virtual_fields and initially_copied_fields being ignored in classes that extent VirtualPage

### DIFF
--- a/code/Model/VirtualPage.php
+++ b/code/Model/VirtualPage.php
@@ -103,8 +103,8 @@ class VirtualPage extends Page
     public function getNonVirtualisedFields()
     {
         return array_merge(
-            VirtualPage::config()->non_virtual_fields,
-            VirtualPage::config()->initially_copied_fields
+            self::config()->non_virtual_fields,
+            self::config()->initially_copied_fields
         );
     }
 


dhensby:
@mikenz What's the difference here and what problem is it solving?

Perhaps a unit test would make it clearer?

mikenz:
I don't have a clean enough dev environment to test this, but....

```
use SilverStripe\CMS\Model\VirtualPage;

class MyVirtual extends VirtualPage 
{
    private static $non_virtual_fields = [
        'MyNonVirtualField',
   ];

   private static $initially_copied_fields = [];
}
```

```
$myVirtual = new MyVirtual();
var_dump($myVirtual->getNonVirtualisedFields());
```

It should return `['MyNonVirtualField']`. Instead it returns the non_virtual_fields and initially_copied_fields  from VirtualPage not MyVirtual class.

dhensby:
OK - I thought that we'd need `static` and not `self` for that... I can never remember the intracacies of the two but I thought `self` always referred to the class it was written in (`VirtualPage` in this instance) and `static` would refer to the class that is currently instantiated (`MyVirtual` in your example)...

mikenz:
I can never remember either, but it's consistent with the other ::config() call in VritualPage and it fixed my unit tests for my class that extends VirtualPage (ie it works for me...).

tractorcow:
In this case, it doesn't matter. `self::config()` always returns config for the late-bound class. Yes it's unexpected, but it does so anyway. This is because `get_called_class()` doesn't seem to respect it properly.

dhensby:
> This is because `get_called_class()` doesn't seem to respect it properly.

That's rubbish - I think we should aim to use the semantically correct syntax regardless

PR #1760, opened 2017-03-10
Add dependency on admin module

tractorcow:
Requires https://github.com/silverstripe/silverstripe-framework/pull/6695

We had to move CMSPreviewable to framework module, as otherwise framework classes couldn't inherit it without directly requiring admin module (undesirable).

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -4,7 +4,7 @@ namespace SilverStripe\CMS\Controllers;
 
 use SilverStripe\Admin\AdminRootController;
 use SilverStripe\Admin\CMSBatchActionHandler;
-use SilverStripe\Admin\CMSPreviewable;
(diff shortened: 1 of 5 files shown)


dhensby:
> We had to move CMSPreviewable to framework module, as otherwise framework classes couldn't inherit it without directly requiring admin module (undesirable).

this sounds bad, surely all of those classes should then be moved to the admin module? which classes are they?

tractorcow:
Because otherwise we'd need admin to be a dependency of framework, which is a more annoying coupling.

dhensby:
> Because otherwise we'd need admin to be a dependency of framework, which is a more annoying coupling.

why?

tractorcow:
Sorry I modified my comment. :)

dhensby:
hmm... ok :/

PR #1761, opened 2017-03-13
BUG Fix PHPUNIT_COVERAGE_TEST by moving to php 5 build

tractorcow:
See https://github.com/travis-ci/travis-ci/issues/5323#issuecomment-165536138

Apparently phpunit 4.8 was designed for use with php 5.5, but doesn't work properly with php7.

Once we upgrade to phpunit 5 we can do coverage again on php7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ env:
 
 matrix:
   include:
-    - php: 5.5
-      env: DB=MYSQL PHPUNIT_TEST=1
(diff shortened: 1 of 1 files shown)


tractorcow:
Note that the other suggestion in that thread (using phpdbg) was previously tried and didn't work. https://github.com/travis-ci/travis-ci/issues/5323#issuecomment-172000818

PR #1765, opened 2017-03-24
Update class namespace for campaign admin

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/6735


sminnee:
@tractorcow now that we've merged versioned, does this need a rebase?

tractorcow:
Rebased.

chillu:
Given the `cms` module depends on campaign admin functionality, should they require it in `composer.json`? e.g. for `AddToCampaignHandler`

chillu:
Damian agrees that we need that dependency, so I've added it here.

PR #1766, opened 2017-03-27
Check csrf timeout error and preserve unsaved form data

ishannz:


ishannz:
addressing CMS issue #1379, ensuring data doesn't get lost if a CSRF error is triggered on Form submission in the cms.

tractorcow:
This solution is still very fragile. A string comparison on a localised form error message doesn't feel like it's going to hold up well to age.

tractorcow:
You can put this into an extension on CMSMain


```
public function updateEditForm(Form $form) {
	// retain stored session information if a CSRF error timeout was triggered
	$data = Session::get("FormInfo.{$form->FormName()}.data");
	$message = $form->Message();
	// check against error, and whether data exists in the session
	// and override the form data if so.
	if($message === _t("Form.CSRF_EXPIRED_MESSAGE", "Your session has expired. Please re-submit the form.")
		&& $data) {
		$form->loadDataFrom($data);
	}
 }
 ```

PR #1772, opened 2017-03-30
Test fixes for upstream changes

dhensby:


dhensby:
Looks like we're going green :)

PR #1774, opened 2017-03-31
ENHANCEMENT Update archive warning message

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/6258

### DIFF
--- a/client/lang/en.js
+++ b/client/lang/en.js
@@ -30,7 +30,6 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "CMSMAIN.SELECTONEPAGE": "Please select at least 1 page.",
     "CMSMAIN.URLSEGMENTVALIDATION": "URLs can only be made up of letters, digits and hyphens.",
     "CMSMAIN.WARNINGSAVEPAGESBEFOREADDING": "You have to save a page before adding children underneath it",
-    "CMSMain.Archive": "Are you sure you want to archive this page and all of its children pages?\n\nThis page and all of its children will be unpublished and sent to the archive.",
     "CMSMain.ConfirmRestoreFromLive": "Are you sure you want to revert draft to when the page was last published?",
     "CMSMain.DeleteFromDraft": "Are you sure you want to remove your page from the draft site?\n\nThis page will remain on the published site.",
     "CMSMain.Restore": "Are you sure you want to restore this page from archive?",
--- a/client/src/legacy/CMSMain.EditForm.js
+++ b/client/src/legacy/CMSMain.EditForm.js
@@ -306,11 +306,11 @@ $.entwine('ss', function($){
 		 *  (Event) e
 		 */
 		onclick: function(e) {
-			var form = this.parents('form:first'), version = form.find(':input[name=Version]').val(), message = '';
-			message = i18n.sprintf(
-				i18n._t('CMSMain.Archive'),
-				version
-			);
+			var form = this.parents('form:first'), message = '';
+			message = form.find('input[name=ArchiveWarningMessage')
+				.val()
+				.replace(/\\n/g, '\n');
+
 			if(confirm(message)) {
 				return this._super(e);
 			} else {
--- a/client/src/legacy/CMSMain.EditForm.js
+++ b/client/src/legacy/CMSMain.EditForm.js
@@ -307,7 +307,7 @@ $.entwine('ss', function($){
 		 */
 		onclick: function(e) {
 			var form = this.parents('form:first'), message = '';
-			message = form.find('input[name=ArchiveWarningMessage')
+			message = form.find('input[name=ArchiveWarningMessage]')
 				.val()
 				.replace(/\\n/g, '\n');
 
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -60,6 +60,8 @@ use SilverStripe\View\ArrayData;
 use SilverStripe\View\Requirements;
 use Translatable;
 use InvalidArgumentException;
+use SilverStripe\Versioned\ChangeSet;
+use SilverStripe\Versioned\ChangeSetItem;
 
 /**
  * The main "content" area of the CMS.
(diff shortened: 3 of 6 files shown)


client/src/legacy/CMSMain.EditForm.js
```diff
@@ -306,11 +306,11 @@ $.entwine('ss', function($){
...
 		onclick: function(e) {
-			var form = this.parents('form:first'), version = form.find(':input[name=Version]').val(), message = '';
-			message = i18n.sprintf(
-				i18n._t('CMSMain.Archive'),
-				version
-			);
+			var form = this.parents('form:first'), message = '';
+			message = form.find('input[name=ArchiveWarningMessage')
+				.val()
+				.replace(/\\n/g, '\n');
```
phalkunz: I'd imagine with a major refactor of the CMS we don't inject the message in a `hidden` input so this will go away.

flamerohr: Yeah, the form schema has a `messages` property in the `state` part that could be used. For now retaining this is fine.

code/Controllers/CMSMain.php
```diff
@@ -744,8 +746,36 @@ public function getEditForm($id = null, $fields = null)
...
+                $archiveWarningTerm = 'CMSMain.ArchiveWarningWithChildren';
+            } else if ($inChangeSets->count() > 0) {
+                $archiveWarningTerm = 'CMSMain.ArchiveWarningWithCampaigns';
+            } else {
+                $archiveWarningTerm = 'CMSMain.ArchiveWarning';
+            }
+
+            $numCampaigns = ChangeSet::singleton()->i18n_pluralise($inChangeSets->count());
+            $numCampaigns = mb_strtolower($numCampaigns);
+            $archiveWarningMsg->setValue(_t($archiveWarningTerm, ' ', [ 'NumCampaigns' => $numCampaigns ]));
```
flamerohr: Please set proper translation fallbacks so we can see immediately what the translation is meant to be (and proper fallback if translation files were removed for some reason in the future).

code/Controllers/CMSMain.php
```diff
@@ -744,8 +746,36 @@ public function getEditForm($id = null, $fields = null)
...
 
+            // Get all page's descendants
+            $record->collateDescendants(true, $descendants);
+            if (!$descendants) {
+                $descendants = [];
+            }
+
+            // Get all campaigns that the page and its descendants belong to
+            $inChangeSetIDs = ChangeSetItem::get_for_object($record)->column('ChangeSetID');
+            foreach ($descendants as $page) {
```
flamerohr: Could we add a `count($descendents) > 0` check here

code/Controllers/CMSMain.php
```diff
@@ -744,8 +746,36 @@ public function getEditForm($id = null, $fields = null)
...
+                $inChangeSetIDs = array_merge($inChangeSetIDs, ChangeSetItem::get_for_object($page)->column('ChangeSetID'));
+            }
+            $inChangeSets = ChangeSet::get()->filter(['ID' => $inChangeSetIDs, 'State' => ChangeSet::STATE_OPEN]);
```
flamerohr: Add a `count($inChangeSetIDs)`, to reduce the unnecessary database calls.

phalkunz: Is it necessary here since the whole block will be irrelevant when the descendants is `[]`?

code/Controllers/CMSMain.php
phalkunz: Is it necessary here since the `foreach` block is not going to evaluate when the descendants is empty?

code/Controllers/CMSMain.php
phalkunz: Agree with this

code/Controllers/CMSMain.php
flamerohr: ah hah, you're right, I was supposed to remove this comment for the one below this about the database calls (I guess I forgot)

code/Controllers/CMSMain.php
```diff
@@ -744,8 +746,46 @@ public function getEditForm($id = null, $fields = null)
...
+            if (count($inChangeSetIDs) > 0) {
+                $inChangeSets = ChangeSet::get()->filter(['ID' => $inChangeSetIDs, 'State' => ChangeSet::STATE_OPEN]);
+            } else {
+                $inChangeSets = new ArrayList();
+            }
+
+            if (count($descendants) > 0 && $inChangeSets->count() > 0) {
+                $archiveWarningTerm = 'CMSMain.ArchiveWarningWithChildrenAndCampaigns';
+                $archiveWarningTransDefault = 'Warning: This page and all of its child pages will be unpublished and automatically removed from their associated {NumCampaigns} before being sent to the archive.\n\nAre you sure you want to proceed?';
+            } else if (count($descendants) > 0) {
```
flamerohr: linting (will fix) use `elseif` instead of `else if`

tractorcow:
This feature doesn't actually remove the page from any active campaigns on archive. Isn't that a primary acceptance criteria of this story?

code/Controllers/CMSMain.php
```diff
@@ -744,8 +746,46 @@ public function getEditForm($id = null, $fields = null)
...
+            }
+
+            if (count($inChangeSetIDs) > 0) {
+                $inChangeSets = ChangeSet::get()->filter(['ID' => $inChangeSetIDs, 'State' => ChangeSet::STATE_OPEN]);
+            } else {
+                $inChangeSets = new ArrayList();
+            }
+
+            if (count($descendants) > 0 && $inChangeSets->count() > 0) {
+                $archiveWarningTerm = 'CMSMain.ArchiveWarningWithChildrenAndCampaigns';
```
tractorcow: Please substitute these two variables with the actually localised value, and invoke `_t(` with the literal key and default values in place. Otherwise i18nTextCollector won't pick these up.

code/Controllers/CMSMain.php
```diff
@@ -744,8 +746,46 @@ public function getEditForm($id = null, $fields = null)
...
+                $archiveWarningTerm = 'CMSMain.ArchiveWarningWithCampaigns';
+                $archiveWarningTransDefault = 'Warning: This page will be unpublished and automatically removed from their associated {NumCampaigns} before being sent to the archive.\n\nAre you sure you want to proceed?';
+            } else {
+                $archiveWarningTerm = 'CMSMain.ArchiveWarning';
+                $archiveWarningTransDefault = 'Warning: This page will be unpublished before being sent to the archive.\n\nAre you sure you want to proceed?';
+            }
+
+            $numCampaigns = ChangeSet::singleton()->i18n_pluralise($inChangeSets->count());
+            $numCampaigns = mb_strtolower($numCampaigns);
+            $archiveWarningMsg->setValue(_t($archiveWarningTerm, $archiveWarningTransDefault, [ 'NumCampaigns' => $numCampaigns ]));
```
tractorcow: All of this message code would be best put into a separate method. getEditForm is getting huge as it is.

code/Controllers/CMSMain.php
```diff
@@ -744,8 +746,46 @@ public function getEditForm($id = null, $fields = null)
...
+
+            if (count($inChangeSetIDs) > 0) {
+                $inChangeSets = ChangeSet::get()->filter(['ID' => $inChangeSetIDs, 'State' => ChangeSet::STATE_OPEN]);
+            } else {
+                $inChangeSets = new ArrayList();
+            }
+
+            if (count($descendants) > 0 && $inChangeSets->count() > 0) {
+                $archiveWarningTerm = 'CMSMain.ArchiveWarningWithChildrenAndCampaigns';
+                $archiveWarningTransDefault = 'Warning: This page and all of its child pages will be unpublished and automatically removed from their associated {NumCampaigns} before being sent to the archive.\n\nAre you sure you want to proceed?';
```
tractorcow: This feature doesn't actually remove the page from any active campaigns on archive. Isn't that a primary acceptance criteria of this story?

phalkunz: My mistake. I thought the functionality was in place so I only updated the message. I've added in a new task for that. Thanks!

code/Controllers/CMSMain.php
phalkunz: Thanks! I'll fix this.

code/Controllers/CMSMain.php
phalkunz: Agreed!

code/Controllers/CMSMain.php
```diff
@@ -828,6 +833,44 @@ public function getEditForm($id = null, $fields = null)
...
+        } else {
+            $inChangeSets = new ArrayList();
+        }
+
+        $numCampaigns = ChangeSet::singleton()->i18n_pluralise($inChangeSets->count());
+        $numCampaigns = mb_strtolower($numCampaigns);
+
+        if (count($descendants) > 0 && $inChangeSets->count() > 0) {
+            $archiveWarningMsg = _t('CMSMain.ArchiveWarningWithChildrenAndCampaigns', 'Warning: This page and all of its child pages will be unpublished and automatically removed from their associated {NumCampaigns} before being sent to the archive.\n\nAre you sure you want to proceed?', [ 'NumCampaigns' => $numCampaigns ]);
+        } else if (count($descendants) > 0) {
```
flamerohr: Please be careful of rebasing your work, I think you force pushed my CSLint fixes for updating `else if` to be `elseif` ;)

phalkunz: Sorry, I forgot that you pushed your change to this branch.

client/lang/en.js
```diff
@@ -30,7 +30,6 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "CMSMAIN.SELECTONEPAGE": "Please select at least 1 page.",
     "CMSMAIN.URLSEGMENTVALIDATION": "URLs can only be made up of letters, digits and hyphens.",
     "CMSMAIN.WARNINGSAVEPAGESBEFOREADDING": "You have to save a page before adding children underneath it",
-    "CMSMain.Archive": "Are you sure you want to archive this page and all of its children pages?\n\nThis page and all of its children will be unpublished and sent to the archive.",
```
tractorcow: We can't remove strings from master keys, it'll delete it from 3.x branch too, since transifex is used as a single source for both branches.

code/Controllers/CMSMain.php
```diff
@@ -828,6 +833,43 @@ public function getEditForm($id = null, $fields = null)
         }
     }
 
+    protected function getArchiveWarningMessage($record)
+    {
+        // Get all page's descendants
+        $record->collateDescendants(true, $descendants);
```
tractorcow: Personal note: Investigate if this is a cause of significant CMS slowdown.

chillu: @tractorcow @flamerohr @phalkunz Yes, it is. Which we've found out painfully by clients 9 months later in a large pre-production codebase, and is now causing us some pretty awkward discussions and team impact :/ https://github.com/silverstripe/silverstripe-admin/issues/571#issuecomment-416794591

flamerohr: A good discussion for accuracy vs performance :)

chillu: Hm, I think that's missing the point - in this case you can get accuracy *and* performance (by computing on demand when the info is actually used), but only if you think about this use case with performance in mind from the start

PR #1776, opened 2017-04-02
Remove extraneous gitignore declarations

tractorcow:
This was causing the https://github.com/silverstripe/silverstripe-cms/tree/master/code/Reports folder to appear as ignored (in some IDE). Modules aren't normally installed within CMS dir, so removing these from gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,5 @@
-.sass-cache
-.DS_Store
 node_modules
-framework
-siteconfig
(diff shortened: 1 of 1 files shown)


dhensby:
I'd also remove `.DS_Store` ;)

tractorcow:
Also removed `.sass-cache` which isn't used anymore.

PR #1781, opened 2017-04-04
Fixing VirtualPage for php7 patch

dhensby:
see https://github.com/silverstripe/silverstripe-framework/pull/6771

### DIFF
--- a/code/model/VirtualPage.php
+++ b/code/model/VirtualPage.php
@@ -416,7 +416,7 @@ class VirtualPage extends Page {
 		} else if(parent::hasField($field) || ($field === 'ID' && !$this->exists())) {
 			return $this->getField($field);
 		} else {
-			return $this->copyContentFrom()->$field;
+			return $this->CopyContentFrom()->$field;
 		}
 	}
 	
@@ -449,7 +449,7 @@ class VirtualPage extends Page {
 		if(parent::hasMethod($method)) {
 			return parent::__call($method, $args);
 		} else {
-			return call_user_func_array(array($this->copyContentFrom(), $method), $args);
+			return call_user_func_array(array($this->CopyContentFrom(), $method), $args);
 		}
 	}
 
@@ -485,8 +485,8 @@ class VirtualPage extends Page {
 	 * @return string
 	 */
 	public function castingHelper($field) {
-		if($this->copyContentFrom()) {
-			return $this->copyContentFrom()->castingHelper($field);
+		if($this->CopyContentFrom()) {
+			return $this->CopyContentFrom()->castingHelper($field);
 		} else {
 			return parent::castingHelper($field);
 		}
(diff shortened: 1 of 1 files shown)


code/model/VirtualPage.php
```diff
@@ -510,8 +497,17 @@ public function getViewer($action) {
 	 * We can't load the content without an ID or record to copy it from.
 	 */
 	public function init(){
-		if(isset($this->record) && $this->record->ID){
-			if($this->record->VersionID != $this->failover->CopyContentFrom()->Version){
+		if (isset($this->record)) {
+			if (is_array($this->record)) {
```
dhensby: This is a bad sign - tests used to pass on the assumption `$this->record` was an object, so there is some kind of API change going on here.

I have no idea how `$this->record` was ever an object... it's always an array in SS :/

dhensby: You know what, I bet the original isset was never working before and so it never actually evaluated the second part of the if...

kinglozzer: Test failures look related to these changes, I can’t work out how it was supposed to work either 😅

kinglozzer: If your second comment is true (it never got past the first `isset()`), can we just remove this entire block of code?

code/model/VirtualPage.php
```diff
@@ -467,8 +454,8 @@ public function hasMethod($method) {
 	 * @return string
 	 */
 	public function castingHelper($field) {
-		if($this->copyContentFrom()) {
-			return $this->copyContentFrom()->castingHelper($field);
```
sminnee: if this was necessary to fix a bug it's a bad sign - methods should be case-insensitive. Recommend reverting this change and fixing upstream if needed.

code/model/VirtualPage.php
sminnee: I expect that this is a bug; it would be good to fix the bug but outside the scope of the PHP7 work.

I recommend:

 *  calling a PHP5.6-based or `3` mainline test run and seeing if the inner block gets called. I want to confirm that `ViewableData::__get() / __isset()` changes haven't mucked with this.

 * commenting out the block and seeing if tests pass. If they do, I'd raise a separate ticket for a bugfix that replaces `&& $this->record->ID` with `&& $this->record['ID']`.`$this->record` should never be an object and we don't need to test for that. I suggest making it a separate PR/issue since we'd probably want to include a test for it and get it merged to SS4, both of which are a distraction from the task at hand.

That being the case, I think we can move forward with the code commented out with a link to the aforementioned ticket.

code/model/VirtualPage.php
```diff
@@ -441,22 +441,9 @@ public function __call($method, $args) {
...
 	public function hasField($field) {
 		return (
-			array_key_exists($field, $this->record)
-			|| $this->hasDatabaseField($field)
-			|| array_key_exists($field, $this->db()) // Needed for composite fields
-			|| parent::hasMethod("get{$field}")
+			parent::hasField($field)
 			|| $this->CopyContentFrom()->hasField($field)
 		);
-	}
```
sminnee: I don't understand why you've deleted hasMethod...

tractorcow: Yes I think it's needed. DataObject::hasMethod respects `get{$field}` so this needs to also.

sminnee:
FYI I've confirmed that the case-correction commit wasn't necessary to get the tests to pass.

I made some tweaks to this and now I think it's ready for merge.

PR #1784, opened 2017-04-07
Make page urls bookmarkable

timkung:
3.x backport from @tractorcow's work to move page ids into the url

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -15,6 +15,10 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
 	private static $url_rule = '/$Action/$ID/$OtherID';
 
+	private static $url_handlers = array(
+		'EditForm/$ID' => 'EditForm',
+	);
+
 	// Maintain a lower priority than other administration sections
 	// so that Director does not think they are actions of CMSMain
 	private static $url_priority = 39;
@@ -65,6 +69,13 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 	 * @config
 	 */
 	private static $enabled_legacy_actions = array();
+
+	/**
+	 * The page id for this request
+	 *
+	 * @var int|null
+	 */
+	protected $pageID = null;
 
 	public function init() {
 		// set reading lang
@@ -593,6 +604,29 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 	}
 
 	/**
+	 * Ensuring we set the current page id from the $ID url parameter.
+	 *
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return Form
+	 */
+	public function EditForm($request = null) {
+		// set page ID from request
+		if ($request) {
+			// validate id is present
+			$id = $request->param('ID');
+
+			if (!isset($id)) {
+				return $this->httpError(400);
+			}
+
+			$this->setCurrentPageID($id);
+		}
+
+		return $this->getEditForm();
+	}
+
+	/**
 	 * @param int $id
 	 * @param FieldList $fields
 	 * @return CMSForm
@@ -689,6 +723,9 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 				$readonlyFields = $form->Fields()->makeReadonly();
 				$form->setFields($readonlyFields);
 			}
+
+			// update form action to include $pageID
+			$form->setFormAction(Controller::join_links($form->FormAction(), $id));
 
 			$this->extend('updateEditForm', $form);
 			return $form;
@@ -866,8 +903,25 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 		return $listview;
 	}
 
+	/**
+	 * Set the page id into $pageID rather than into {@link Session}.
+	 *
+	 * @param string|int $id
+	 *
+	 * @return void
+	 */
+	public function setCurrentPageID($id) {
+		$id = (int)$id;
+		$this->pageID = $id;
+	}
+
+	/**
+	 * Get the page id from this request
+	 *
+	 * @return int
+	 */
 	public function currentPageID() {
-		$id = parent::currentPageID();
+		$id = $this->pageID;
 
 		$this->extend('updateCurrentPageID', $id);
 
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -725,7 +725,11 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 			}
 
 			// update form action to include $pageID
-			$form->setFormAction(Controller::join_links($form->FormAction(), $id));
+			$form->setFormAction(Controller::join_links(
+				$this->Link(),
+				$form->getName(),
+				$id
+			));
 
 			$this->extend('updateEditForm', $form);
 			return $form;
--- a/code/controllers/CMSPageHistoryController.php
+++ b/code/controllers/CMSPageHistoryController.php
@@ -12,56 +12,81 @@ class CMSPageHistoryController extends CMSMain {
 	private static $menu_title = 'History';
 	private static $required_permission_codes = 'CMS_ACCESS_CMSMain';
 	private static $session_namespace = 'CMSMain';
-	
+
 	private static $allowed_actions = array(
+		'EditForm',
 		'VersionsForm',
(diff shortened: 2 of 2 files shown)


code/controllers/CMSPageHistoryController.php
```diff
@@ -241,120 +308,25 @@ public function VersionsForm() {
...
-	
-	/**
-	 * Process the {@link VersionsForm} compare function between two pages.
-	 *
-	 * @param array
-	 * @param Form
-	 *
-	 * @return html
-	 */
-	public function doCompare($data, $form) {
```
tractorcow: I think for the 3.x version of this PR, we can't remove this method (semver). Can you add this back please?

code/controllers/CMSPageHistoryController.php
```diff
@@ -241,120 +308,25 @@ public function VersionsForm() {
...
-	/**
-	 * Process the {@link VersionsForm} show version function. Only requires
-	 * one page to be selected.
-	 *
-	 * @param array
-	 * @param Form
-	 *
-	 * @return html
-	 */	
-	public function doShowVersion($data, $form) {
```
tractorcow: As above, revert removal please. :)

code/controllers/CMSPageHistoryController.php
```diff
@@ -241,120 +308,25 @@ public function VersionsForm() {
...
-			$this->Link('version'),
-			$versionID
-		));
-	}
-
-	/**
-	 * @param int|null $versionID
-	 * @return Form
-	 */
-	public function ShowVersionForm($versionID = null) {
```
tractorcow: And this. :P

code/controllers/CMSPageHistoryController.php
```diff
@@ -169,17 +234,19 @@ public function getEditForm($id = null, $fields = null, $versionID = null, $comp
...
-		if(($record && $record->isLatestVersion())) {
+
+		if ($record->isLatestVersion()) {
 			$revert->setReadonly(true);
 		}
-		
+
 		$form->removeExtraClass('cms-content');
 
+		$form->setFormAction(Controller::join_links($form->FormAction(), $id, $versionID));
```
tractorcow: The risk here is that $form->FormAction() Could already have the ID in it set by CMSmain::getEditForm(). Repeating this action could end up with double-ID.

Safer fix here is.

```php
$form->setFormAction(Controller::join_links($this->Link(), $form->getName(), $id, $versionID));
```

code/controllers/CMSMain.php
```diff
@@ -690,6 +724,9 @@ public function getEditForm($id = null, $fields = null) {
 				$form->setFields($readonlyFields);
 			}
 
+			// update form action to include $pageID
+			$form->setFormAction(Controller::join_links($form->FormAction(), $id));
```
tractorcow: As mentioned in CMSPageHistoryController.php a less-risky alternative:

```php
$form->setFormAction(Controller::join_links($this->Link(), $form->getName(), $id));
```

code/controllers/CMSPageHistoryController.php
timkung: sure done :)

code/controllers/CMSPageHistoryController.php
timkung: sure done :)

code/controllers/CMSPageHistoryController.php
timkung: sure done :)

code/controllers/CMSPageHistoryController.php
timkung: sure done :)

code/controllers/CMSMain.php
timkung: sure done :)

tractorcow:
beta1 time?

Zauberfisch:
I absolutely love the idea behind this pull request, it is in fact something I have brought up recently in silverstripe/silverstripe-framework/issues/6675.

However, this seems to break the Settings and History tab of pages. 
Both [Settings->getEditForm()](https://github.com/silverstripe/silverstripe-cms/blob/51c8e8639e9b5f1c258f21f640df2a6cf2af12df/code/controllers/CMSPageSettingsController.php#L17) and [History->getEditForm()](https://github.com/silverstripe/silverstripe-cms/blob/51c8e8639e9b5f1c258f21f640df2a6cf2af12df/code/controllers/CMSPageHistoryController.php#L109) seem to invoke `parent::getEditForm($record, ...)`, but [CMSMain->getEditForm()](https://github.com/silverstripe/silverstripe-cms/blob/bb3c09c359158b046c4e06a023d6f77f0052d8fc/code/controllers/CMSMain.php#L634) actually expects an integer as first paramter (`public function getEditForm($id = null, $fields = null) {}`).


This in turn results in the following code breaking:

    // update form action to include $pageID
    $form->setFormAction(Controller::join_links(
        $this->Link(),
        $form->getName(),
        $id
    ));
([link to above code in CMSMain->getEditForm()](https://github.com/silverstripe/silverstripe-cms/blob/bb3c09c359158b046c4e06a023d6f77f0052d8fc/code/controllers/CMSMain.php#L728))


So, this is obvious: We need to fix Settings->getEditForm() and History->getEditForm() to call with ID. 
But I am now wondering, why is this issue tackled in CMS rather than in framework? As far as I can tell this problem of non-unique URLs is universal across all LeftAndMain controllers if they don't fix it themselves.
LeftAndMain has a bad "state management", and now all sub classes individually work around it. Shouldn't we fix this in the base class instead?

tractorcow:
> why is this issue tackled in CMS rather than in framework? 

The inclusion of ID into the form url is opt-in, not opt-out. It's baked into CMS because page forms need that ID context, whereas certain other sections (e.g. modeladmin, which has a gridfield as a root form field) don't. The context in these cases is nested within the gridfield, which has it's own record-ID url-ification.

In any case, we need to fix the broken settings forms for 3.6.0 stable, if broken.

Zauberfisch:
Non the less LeftAndMain still has [LeftAndMain->currentPageID()](https://github.com/silverstripe/silverstripe-framework/blob/3362e15a2978918c7d3f2b72b30844e2f5ee53eb/admin/code/LeftAndMain.php#L1918), which contradicts that Page ID handling is opt in.

Some LeftAndMains may not use it, but it's there anyway. So we should fix it there instead of overwriting it in all sub classes.

tractorcow:
> Non the less LeftAndMain still has LeftAndMain->currentPageID(), which contradicts that Page ID handling is opt in.

I know, but this is 3.x, so a lot of legacy architecture needs to be tolerated. LeftAndMain shouldn't be enforcing a "page id" context in any case, but it's a major version breakage to refactor.

tractorcow:
Problem is with this commit https://github.com/silverstripe/silverstripe-cms/commit/487235f9918b762f9eadcaaa655602da884807a6

image

It needs to fail over to currentPageID(), otherwise it'll break if the id is in url_params.

Zauberfisch:
But that would just be a hotfix and workaround for the fact that a record is passed to EditForm instead of an ID

tractorcow:
> But that would just be a hotfix and workaround for the fact that a record is passed to EditForm instead of an ID

Yes, but that's the way it's always been, and it is poor behaviour. However, it's not the cause of the regressions in this PR.

Zauberfisch:
I do understand the problem that we can't fix LeftAndMain in 3.6.

But when we fix it in CMSMain, could we at least fix it in a way that Settings and History still work with the new URL instead of falling back to session again?

One alternative approach I can think of is checking if $id is actually a object instead and then `$id = $id->ID`.
This would keep it backwards compatible, yet allows us to clean up Settings and History.

tractorcow:
I've pinged @timkung to maybe have a look at it... otherwise I'll check it out some time later this week. :D I'll see if I can cleanup the record vs id inconsistency, unless @timkung is awesome and gets to it before me. I'm pretty busy though!

tractorcow:
ok, this is why we pass in `$record`, after some digging.

```
public function getEditForm($id = null, $fields = null) {
		if(!$id) $id = $this->currentPageID();

		if(is_object($id)) {
			$record = $id;
		} else {
			$record = $this->getRecord($id);
			if($record && !$record->canView()) return Security::permissionFailure($this);
		}
```

the issue is that ID isn't enough to identify the record in the case of, for example, page history. We force to a record for a specific version by passing in the record, instead of the ID. By passing in the ID we could end up getting the form for the draft record, instead of the one at the correct version.

Zauberfisch:
I can confirm that the issue in settings is now fixed, thanks for the quick reaction @tractorcow

Issue #1785, opened 2017-04-07
Hidden list


flamerohr:
I speculate that it's caused by bootstrap's collapsible feature...

![screen shot 2017-04-07 at 4 27 34 pm](https://cloud.githubusercontent.com/assets/1064889/24785584/5c6e735e-1baf-11e7-9f82-be1122d9eed5.png)

chillu:
Pointers to fix: This is in the "Settings" tab of a page, at the bottom on "who can edit this page". It's a `ListboxField` created in `SiteTree.php`. Styles are in `admin/client/src/styles/legacy/_form.scss`

phalkunz:
Assigned to Ishan J.

flamerohr:
PR here https://github.com/silverstripe/silverstripe-cms/pull/1799/files

PR #1799, opened 2017-04-21
fix Hidden listbox #1785

ishannz:


client/src/legacy/CMSMain.EditForm.js
```diff
@@ -232,7 +232,7 @@ $.entwine('ss', function($){
 
       holder.next().filter('.listbox')[instant ? 'hide' : 'slideUp'](() => {
         holder.removeClass('field--merge-below');
-      });
+      }).css('overflow','visible'); // jquery adding overflow:hidden in hide, this will break listbox display
```
flamerohr: awesome, could we add this to the line above as well?
`.slideUp()` has a delay which means this line ends up being overwritten

ishannz: if we added line above like this 
holder.next().filter('.listbox').css('overflow','visible')[instant ? 'hide' : 'slideUp'](() 
then it will be overwitten from the 'hide' class again, then it will the same as previous

flamerohr: After testing, I think this still needs work, if there are text below the hidden field, it'll overlap the textbox as it appears and slides down...

flamerohr:
@ishannz I've toggled the `.css()` calls around
This does something similar to your change but fixes the text overlapping the textbox issue.

Could you have a review of this when you are able to? :) really appreciate it

dhensby:
There are some code style issues that need resolving.

chillu:
Fixing those code style issues myself so I can merge :)

PR #1800, opened 2017-04-21
FIX intl test

dhensby:
It appears that a change to the translation strings has caused this test to fail.

I've also rearranged the parameters to be correct.


dhensby:
merged so we don't get test failures on PRs

PR #1801, opened 2017-04-22
API Upgrade to behat 3

tractorcow:
See https://github.com/silverstripe/silverstripe-behat-extension/pull/152


tractorcow:
It's green!

sminnee:
And now it's purple @tractorcow!

PR #1803, opened 2017-04-26
BUG Fix page history / settings forms

tractorcow:
Fixes issue noted in #1784

@Zauberfisch are you happy with how I've addressed the id / record issue?

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -627,23 +627,25 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 	}
 
 	/**
-	 * @param int $id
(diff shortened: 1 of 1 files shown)


Zauberfisch:
Of course not happy, because I would love to see the whole thing refactored. 

But since we need to satisfy semver if we want to get it into 3.6 and in light of the findings you just posted in #1784, I would say this is the best possible solution, thanks a lot for looking into this so quickly.

tractorcow:
> Of course not happy, because I would love to see the whole thing refactored.

Yes that would be nice. If I had more time, I'd go as far as refactoring framework LeftAndMain, but trying to keep the scope of this patch as small as possible.

PR #1804, opened 2017-04-26
fixing doArchive

pitchandtone:
for users who can delete, but not unpublish, they should be able to delete draft pages that they’ve created.

Old code used to unpublish even if the page wasn’t published, which meant users who couldn’t publish, also could delete draft pages. They could create draft pages, but not delete them.


tractorcow:
@dhensby since this is going into 3.6.0 is this ok to merge? If not please change and merge into 3.6.

PR #1814, opened 2017-05-08
Prevent installation from removing behat.yml

tractorcow:
With the current export-ignore in place, behat.yml is deleted when installed within framework, breaking integration tests when framework runs BEHAT_TEST = cms.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,5 @@
 docs/                 export-ignore
 javascript/src/       export-ignore
-behat.yml             export-ignore
 *.dist                export-ignore
 
 # Hide diffs


sminnee:
Good for now.

In time I question whether it's in users' interests to bundle tests with the dist files, in which case I would put a lot more stuff into export-ignore. At that point we'd need to make the installer test run --prefer-source.

However, that's a future change and this is appropriate now.

tractorcow:
Thanks, and sorry for the sub-optimal work-around. It looks like any time we use --prefer-dist we'll run into this issue.

sminnee:
Yeah it's a bigger issue of "what functionality do we accept losing in our dists?" and deserves discussion outside of a single-issue PR. I'll raise when I get a chance.

PR #1828, opened 2017-05-19
Remove create_function implementations, replace with closures

robbieaverill:
Resolves https://github.com/silverstripe/silverstripe-framework/issues/2842


robbieaverill:
I think we should wait until the builds are stable to be sure

PR #1834, opened 2017-05-26
API Add insert internal link modal

tractorcow:
Required for https://github.com/silverstripe/silverstripe-framework/issues/6740

Ready for review, not ready for merge (need to update tests)


webpack.config.js
```diff
@@ -40,8 +41,16 @@ module.exports = [
     },
     externals: {
       i18n: 'i18n',
+      jquery: 'jQuery',
```
tractorcow: Case sensitivity issues; Copied this solution from admin.

client/src/legacy/TinyMCE_sslink-internal.js
```diff
@@ -0,0 +1,128 @@
...
+    });
+  },
+};
+
+const modalId = 'insert-link__dialog-wrapper--internal';
+const sectionConfigKey = 'SilverStripe\\CMS\\Controllers\\CMSPageEditController';
+const formName = 'editorInternalLink';
+const InsertLinkInternalModal = provideInjector(createInsertLinkModal(sectionConfigKey, formName));
+
+jQuery.entwine('ss', ($) => {
```
tractorcow: Unfortunately html editor still relies on entwine.

code/Forms/InternalLinkFormFactory.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
tractorcow: admin doesn't know anything about page type (the internal link); This code is all CMS specific.

webpack.config.js
```diff
@@ -26,14 +26,19 @@ const config = [
...
     },
     output: {
       path: PATHS.DIST,
       filename: 'js/[name].js',
     },
     devtool: (ENV !== 'production') ? 'source-map' : '',
     resolve: resolveJS(ENV, PATHS),
-    externals: externalJS(ENV, PATHS),
+    externals: Object.assign({}, externalJS(ENV, PATHS), {
+      'containers/InsertLinkModal/InsertLinkModal': 'InsertLinkModal',
```
tractorcow: This can be removed after https://github.com/silverstripe/webpack-config/pull/2 is on NPM.

client/src/legacy/TinyMCE_sslink-internal.js
```diff
@@ -0,0 +1,128 @@
...
+TinyMCEActionRegistrar.addAction('sslink', {
+  text: i18n._t('CMS.LINKLABEL_PAGE', 'Link to page'),
+  // eslint-disable-next-line no-console
+  onclick: (editor) => editor.execCommand('sslinkinternal'),
+});
+
+const plugin = {
+  init(editor) {
+    editor.addCommand('sslinkinternal', () => {
+      const field = window.jQuery(`#${editor.id}`).entwine('ss');
```
tractorcow: Fixed in the rebase.

webpack.config.js
tractorcow: Fixed in rebase.

code/Forms/InternalLinkFormFactory.php
phalkunz: Ah, got ya! :)

client/src/legacy/TinyMCE_sslink-internal.js
```diff
@@ -0,0 +1,128 @@
...
+TinyMCEActionRegistrar.addAction('sslink', {
+  text: i18n._t('CMS.LINKLABEL_PAGE', 'Link to page'),
+  // eslint-disable-next-line no-console
+  onclick: (editor) => editor.execCommand('sslinkinternal'),
+});
+
+const plugin = {
+  init(editor) {
+    editor.addCommand('sslinkinternal', () => {
+      const field = $(`#${editor.id}`).entwine('ss');
```
phalkunz: Still similar issue as before. `$` here might be different from the imported `jQuery`.

client/src/legacy/TinyMCE_sslink-internal.js
phalkunz: From what I can see the use of entwine here is to add `openLinkInternalDialog()` to the DOM and `openLinkInternalDialog()` is only being used in at line 24 in the same file so instead of attached a function to the DOM using entwine, it can just be a normal function.

flamerohr: I think this is fine to use entwine.
This keeps the extensibility that entwine provides, e.g. if a user wants to modify a behaviour somewhere along the lines.

We can look at removing entwine when we get to moving this to react completely.

Issue #776, opened 2013-06-19
Infinite loop in tree display when B childof A on live and A childof B on stage.


sminnee:
If you have a situation in the tree like this:
- A
  - B

And then you make changes on stage so that it's like this:
- B
  - A

Then you trigger an infinite loop when trying to load this subtree in the CMS.

The underlying cause is that the live-children of A are included in the tree, which includes B.  This means that children deleted from stage but still on live are displayed.  This "get the live nodes" script needs to exclude any pages that are still present on stage and have merely been moved to another location.

This occurred in 3.0.

tractorcow:
@dhensby were you going to help back-port the fix for this to 3.x? 

See https://github.com/silverstripe/silverstripe-framework/issues/5876 for main issue.

tractorcow:
See my gist at https://gist.github.com/tractorcow/540a3056cfce5d45f0d5c680372dad0c

dhensby:
fixed by #1838

PR #1838, opened 2017-05-30
FIX Issue recursion when parent/child records are swapped in SiteTree

dhensby:
see #776

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -1158,10 +1158,11 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 	 * @param string $siteConfigMethod Method to call on {@link SiteConfig} for toplevel items, e.g. "canEdit"
 	 * @param string $globalPermission If the member doesn't have this permission code, don't bother iterating deeper
 	 * @param bool   $useCached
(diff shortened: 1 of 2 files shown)


dhensby:
Don't merge

code/model/SiteTree.php
```diff
@@ -1158,10 +1158,11 @@ static public function prepopulate_permission_cache($permission = 'CanEditType',
 	 * @param string $siteConfigMethod Method to call on {@link SiteConfig} for toplevel items, e.g. "canEdit"
 	 * @param string $globalPermission If the member doesn't have this permission code, don't bother iterating deeper
 	 * @param bool   $useCached
+	 * @param array $stages            Which stages to check permissions against, defaults to both Stage and Live
 	 * @return array An map of {@link SiteTree} ID keys to boolean values
 	 */
 	public static function batch_permission_check($ids, $memberID, $typeField, $groupJoinTable, $siteConfigMethod,
-												  $globalPermission = null, $useCached = true) {
+												  $globalPermission = null, $useCached = true, $stages = array('Stage', 'Live')) {
```
tractorcow: If you add $stages you'll need to add this to the cache key.

dhensby: done

frankmullenger:
@dhensby @tractorcow many thanks!

PR #1849, opened 2017-06-13
Removed ErrorPage

jonom:
Removed ErrorPage class and references so it can be split off in to a separate module.

May require supporting Framework PR for tests to pass. https://github.com/silverstripe/silverstripe-framework/pull/7017

Not sure if there is any point removing the translations? Does that need to happen in Transifex?

Related RFC: https://github.com/silverstripe/silverstripe-framework/issues/4149


flamerohr:
Amazing effort, only test failing is a behat test which targets an ErrorPage :)

tractorcow:
Hi @jonom in order to get this over the line before beta1, I think we will need an errorpage module (suggest `silverstripe/errorpage`) and add this to the installer at the same time as merging this. Do you feel like putting this together?

tractorcow:
Blank repo with necessary permissions setup at https://github.com/silverstripe/silverstripe-errorpage :D

chillu:
Needs a rebase

jonom:
@chillu rebased. Just a couple of behat tests failing - same ones as in the Framework PR so must be unrelated.

jonom:
Thanks for the merge!

Issue #1846, opened 2017-06-08
SS4: Right click, add page fails


mikenz:
AddForm receives POSTs from /admin/pages/add and GETs from right clicking a page in site tree and adding a child.

With https://github.com/silverstripe/silverstripe-framework/pull/6998 the GET is now `405 Method Not Allowed`

dhensby:
Thanks @mikenz for finding and reporting this. I think the best solution is to turn off `strictFormMethodCheck` for this form...

tractorcow:
@AntonyThorpe will you PR that fix ?

AntonyThorpe:
@tractorcow PR submitted as requested.  All checks have passed.  Thank you.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-cms/pull/1852

Thanks @AntonyThorpe you are great :D

PR #1852, opened 2017-06-15
Bugfix: SS4 Right click, add page fails

AntonyThorpe:
`addForm` function receives both POSTs and GETs.  Set the form's `strictFormMethodCheck` to false to avoid `405 Method Not Allowed` error responses.  Issue #1846.


dhensby:
thanks

Issue #1780, opened 2017-04-04
Visual updates to page view toggle


clarkepaul:
We now have a newer pattern for 2 option toggles within the CMS which has been established in the Files area but the Pages area still has the older functionality.

Only show the option to go to the view you do not have currently (ie if you are on tree view show the icon for list view).
Move the toggle to align with panel options toolbar, far right of any actions (as per files area).

Replace batch action label text to just "Actions" so that there is sufficient space on smaller panel sizes.

I hope a few benefits come from these changes:
Increase spacing in the top toolbar which will be useful for when the search is updated to newer pattern, as well as more space for custom section headers ("Pages" obviously doesn't need the space but there might be modal admins with toggle?)
The toggle will take up less space in general especially as it is not surrounded with the border.
Toggle will be in closer proximity to the tree or list.
Simplified labelling for batch actions, I think this one is relatively low risk but would be good to hear a few opinions/testing.

<img width="178" alt="pasted_image_4_04_17__5_47_pm" src="https://cloud.githubusercontent.com/assets/555033/24643418/f30fc740-1961-11e7-974c-8d1b7e90d771.png">

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-cms/pull/1858
- [x] https://github.com/silverstripe/silverstripe-admin/pull/111

clarkepaul:
I've tested this with the full label of "batch actions" and moving the view button inline should still fit with a few other changes.

* Buttons in the content toolbar would need to have a reduced spacing on the right side (from 8px to 4px
* The view toggle would need to be set to the far right of the content toolbar so that it is 10px from the far right side (possibly with margin-right: -10px or positioned absolute and right:10px)

Before:
<img width="191" alt="pasted_image_19_06_17__2_24_pm" src="https://user-images.githubusercontent.com/555033/27267346-d3c0884a-54fb-11e7-9d54-9b3ad6200a08.png">

After:
<img width="171" alt="pasted_image_19_06_17__2_25_pm" src="https://user-images.githubusercontent.com/555033/27267338-c775b844-54fb-11e7-87ac-73d2db974f5f.png">

PR #1858, opened 2017-06-20
Visual updates to page view toggle

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-cms/issues/1780


dhensby:
I've rebased this because behat was failing and I'm hoping that the latest commits fix that ;)

dhensby:
looks like it did :)

PR #1869, opened 2017-06-27
BUG Don’t click hidden button

tractorcow:


tractorcow:
Nice green, thanks @kinglozzer

PR #1870, opened 2017-06-27
FIX namespaces in docs

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/5757

### DIFF
--- a/code/Controllers/CMSPageHistoryController.php
+++ b/code/Controllers/CMSPageHistoryController.php
@@ -159,7 +159,7 @@ class CMSPageHistoryController extends CMSMain
      * Returns the read only version of the edit form. Detaches all {@link FormAction}
      * instances attached since only action relates to revert.
      *
-     * Permission checking is done at the {@link CMSMain::getEditForm()} level.
+     * Permission checking is done at the {@link SilverStripe\CMS\Controllers\CMSMain::getEditForm()} level.
      *
      * @param int $id ID of the record to show
      * @param array $fields optional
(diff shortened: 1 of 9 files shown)


code/Controllers/CMSPageHistoryController.php
```diff
@@ -159,7 +159,7 @@ public function EditForm($request = null)
      * Returns the read only version of the edit form. Detaches all {@link FormAction}
      * instances attached since only action relates to revert.
      *
-     * Permission checking is done at the {@link CMSMain::getEditForm()} level.
+     * Permission checking is done at the {@link SilverStripe\CMS\Controllers\CMSMain::getEditForm()} level.
```
flamerohr: Did we want to have the FQN all over the docs or perhaps we could just `use SilverStripe\CMS\Controllers\CMSMain` at the top?

phalkunz: No all over but only when use FQN when there's no `use` mentioned.

flamerohr: yeah, that's my concern that if we end up adding a `use` later on, then the Docs are out of date using the FQN when not needed...

I'll leave the change as optional, but nice-to-have :)

@tractorcow mentions:
> please leave them short, and add a `use` if needed
> we don't need unreadable PHPDoc

phalkunz:
This is not priority at the moment and should use short name as much as possible by using use at the top even for {@link}.

PR #1871, opened 2017-06-27
NEW: Remove use of MODULE_DIR constants 

sminnee:
This relies on
https://github.com/silverstripe/silverstripe-framework/pull/7065


_config.php
```diff
@@ -17,14 +17,10 @@
  * - CMS_PATH: Absolute filepath, e.g. "/var/www/my-webroot/cms"
  */
 call_user_func(function () {
-    define('CMS_PATH', __DIR__);
```
chillu: Hmmm, by no longer defining these we need to mention the change in the upgrading guide - it's no longer an opt-in feature. I'll write up something

Issue #1755, opened 2017-03-06
CMSMain::EditForm()'s "archive" action is not validation exempt


jakxnz:
The 'delete' action is exempt from validation, so it would make sense to also include the 'archive' action.

*CMSMain.php line 680*
```
// Set validation exemptions for specific actions
$form->setValidationExemptActions(array('restore', 'revert', 'deletefromlive', 'delete', 'unpublish', 'rollback', 'doRollback'));
```

dhensby:
agreed - PR?

chillu:
Sounds reasonable, happy to review a PR @jakxnz  - closing this as a minor enhancement in the meantime.

PR #1884, opened 2017-07-05
Add archive to list of validation-exempt actions

kinglozzer:
See https://github.com/silverstripe/silverstripe-cms/issues/1755


dhensby:
should this target 3.5?

kinglozzer:
Good point, will rebase

dhensby:
thanks

Issue #948, opened 2014-02-18
Don't let user define CMS page that conflicts with controller class URL segment


Rexeh:
Currently you are able to define a controller with the URL of /book. You can then create a page in the CMS of the same URL segment.

Should be message along the lines of "This is a reserved page name" or similar. Creating page seems to have no negative affects, controller takes precedence when visited.

chillu:
Agreed that it's bad UX - the authors will be left wondering why their page doesn't work.

dhensby:
Fix being worked on here https://github.com/silverstripe/silverstripe-cms/pull/1886

maxime-rainville:
That's been fixed in 4 and we won't be backporting in SS3

PR #1886, opened 2017-07-09
Disable the ability to use reserved URL's

Firesphere:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7139

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -1559,6 +1559,22 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      */
     public function validURLSegment()
     {
+        $excludes = Director::config()->get('rules');
+        $excludes = array_keys($excludes);
+        $disallowedSegments = array_map(function ($key) {
+            $route = explode('/', $key);
+            if (!empty($route) && strpos($route[0], '$') === false) {
+                return $route[0];
+            }
+            return;
+        }, $excludes);
+
+        if (!$this->ParentID && in_array($this->URLSegment, $disallowedSegments)) {
+            // Default to '-2', onBeforeWrite takes care of further possible clashes
+            return false;
+        }
+
+
         if (self::config()->nested_urls && $parent = $this->Parent()) {
             if ($controller = ModelAsController::controller_for($parent)) {
                 if ($controller instanceof Controller && $controller->hasAction($this->URLSegment)) {
@@ -1611,17 +1627,17 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     public function generateURLSegment($title)
     {
         $filter = URLSegmentFilter::create();
-        $t = $filter->filter($title);
+        $filteredTitle = $filter->filter($title);
 
         // Fallback to generic page name if path is empty (= no valid, convertable characters)
-        if (!$t || $t == '-' || $t == '-1') {
-            $t = "page-$this->ID";
+        if (!$filteredTitle || $filteredTitle == '-' || $filteredTitle == '-1') {
+            $filteredTitle = "page-$this->ID";
         }
 
         // Hook for extensions
-        $this->extend('updateURLSegment', $t, $title);
+        $this->extend('updateURLSegment', $filteredTitle, $title);
 
-        return $t;
+        return $filteredTitle;
     }
 
     /**
--- a/tests/php/Model/SiteTreeTest.php
+++ b/tests/php/Model/SiteTreeTest.php
@@ -46,6 +46,15 @@ class SiteTreeTest extends SapphireTest
         SiteTreeTest_DataObject::class,
     );
 
+    public function reservedSegmentsProvider()
+    {
+        return [
+            ['Admin', 'admin-2'],
+            ['Dev', 'dev-2'],
+            ['Robots in disguise', 'robots-in-disguise']
+        ];
+    }
+
     public function testCreateDefaultpages()
     {
             $remove = SiteTree::get();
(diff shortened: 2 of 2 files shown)


code/Model/SiteTree.php
```diff
@@ -18,6 +18,7 @@
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Resettable;
+use SilverStripe\Dev\Debug;
```
dhensby: not used

tests/model/SiteTreeTest.php
```diff
@@ -4,6 +4,7 @@
 use SilverStripe\CMS\Model\VirtualPage;
 use SilverStripe\Control\ContentNegotiator;
 use SilverStripe\Control\Controller;
+use SilverStripe\Dev\Debug;
```
dhensby: not used

tests/model/SiteTreeTest.php
```diff
@@ -108,6 +109,47 @@ public function testURLGeneration()
     }
 
     /**
+     * Check if reserved URL's are properly appended with a number at top level
+     */
+    public function testDisallowedURLGeneration()
+    {
+        // Expected URL defaults to '-2'
+        $expectedURLs = [
```
dhensby: you might want to consider the use of data providers (https://phpunit.de/manual/5.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers)

Firesphere: Good idea, but I think it would be better to do such a thing in all the tests at once, instead of having some of the tests use inline expectation arrays (as `testURLGeneration` on line 88) and others use providers

tests/model/SiteTreeTest.php
Firesphere: Fixed :) I like using debug dump in tests, forgot to remove them.

code/Model/SiteTree.php
```diff
@@ -1604,6 +1604,21 @@ public function generateURLSegment($title)
         // Hook for extensions
         $this->extend('updateURLSegment', $t, $title);
 
+        $excludes = Director::config()->get('rules');
+        $excludes = array_keys($excludes);
+        $disallowed = array_map(function ($key) {
```
tractorcow: You can also use `strtok($key, '/');` as shorthand too. :) (not a change request, just an aside).

code/Model/SiteTree.php
```diff
@@ -1604,6 +1604,21 @@ public function generateURLSegment($title)
...
+            $route = explode('/', $key);
+            return $route[0];
```
tractorcow: Could you do the `$` check in here, and just put array_filter around the entire array_map? It would clean up the complexity of the code here a bit.

tests/model/SiteTreeTest.php
tractorcow: I wouldn't mind it if you just added these for these new tests and ignored the older ones. ;)

tractorcow (changes requested):

dhensby:
Fixes #948

code/Model/SiteTree.php
Firesphere: Yep, can do :)

Firesphere:
@dhensby @tractorcow Applied your feedback. It has no conflicts, but do you want me to rebase as well? (Or GitHub could do that for us ;) )

tests/model/SiteTreeTest.php
Firesphere: After applying... I have immediately fallen in love with data providers! 👍 for that one @dhensby !

code/Model/SiteTree.php
Firesphere: `strto()` is pretty neat! But it fails if there's no slash in the segment, so I'm sticking to the explode :D

tests/model/SiteTreeTest.php
```diff
@@ -823,6 +859,14 @@ public function testIsSection()
         $this->assertTrue($ceo->isSection());
     }
 
+    public function testURLSegmentReserved()
```
Firesphere: Although this is a double-up from previous tests, I believe it is a good test to have, to isolate `admin` testing and have it explicitly tested

Firesphere:
Rebased anyway, just to make sure nothing breaks :)

Firesphere:
I'll rebase this PR this week :)

Any other requests of changes?

dhensby:
I'm happy

dhensby (approved):

Firesphere:
> I'm happy
Is that related to this PR, or just in general :D

code/Model/SiteTree.php
```diff
@@ -1595,17 +1595,32 @@ function ($response) {
...
+            $filteredTitle = "page-$this->ID";
         }
 
         // Hook for extensions
-        $this->extend('updateURLSegment', $t, $title);
+        $this->extend('updateURLSegment', $filteredTitle, $title);
+
+        $excludes = Director::config()->get('rules');
+        $excludes = array_keys($excludes);
+        $disallowed = array_map(function ($key) {
```
tractorcow: This logic should be in validURLSegment().

tractorcow (changes requested):

tractorcow:
Sorry, one more feedback. You need to validate this in the right method. generateURLSegment() isn't the right place.

dhensby:
needs a rebase and final feedback dealing with

Firesphere:
Done, @dhensby @tractorcow

tractorcow (approved):

PR #1894, opened 2017-07-16
FIX Ensure tab states are reflected when switching on page edit screen

### LINKED ISSUE #1883: SS4 (minor) : tab bar indicator in Pages doesn't change when switching tabs
In the Pages section of the CMS, the underlining of the tabs remains on "Content" rather than changing to "Settings" or "History" when navigating to those tabs:

The underline should be under "Settings" in this one:
screenshot from 2017-07-05 14-00-53

The underlining should be under "History" in this one:
screenshot from 2017-07-05 14-01-16

### LINKED ISSUE #1887: Tabs in pages don't update
Tabs don't seem to be updating visually when you click between content and settings
 image

robbieaverill:
Resolves #1883 and resolves #1887

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -350,6 +350,21 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         }
     }
 
+    /**
+     * Decide which is the active primary tab in a CMS edit screen
+     *
+     * @return string
+     */
+    public function getActiveTab()
+    {
+        if ($this instanceof CMSPageSettingsController) {
+            return 'settings';
+        } elseif ($this instanceof CMSPageHistoryController) {
+            return 'history';
+        }
+        return 'edit';
+    }
+
     public function LinkWithSearch($link)
     {
         // Whitelist to avoid side effects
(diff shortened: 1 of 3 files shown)


robbieaverill:
I've added a behat test for this in 9bde76b, it requires https://github.com/silverstripe/silverstripe-framework/pull/7177 to pass since the CMS UI context already had _some_ functionality for the CMS toolbar tabs, but no step definitions for them.

@tractorcow should the `CmsUiContext` class either be moved to the CMS module, or be renamed if its used for other areas?

tractorcow:
> @tractorcow should the CmsUiContext class either be moved to the CMS module, or be renamed if its used for other areas?

The trouble is that CMS is really a euphemism for "admin" in this case. They should stay in the framework module, for now, as they are to be moved to admin module in the near future.

See https://github.com/silverstripe/silverstripe-framework/issues/6702

code/Controllers/CMSMain.php
```diff
@@ -350,6 +350,21 @@ public function LinkPageHistory()
...
     }
 
+    /**
+     * Decide which is the active primary tab in a CMS edit screen
+     *
+     * @return string
+     */
+    public function getActiveTab()
+    {
+        if ($this instanceof CMSPageSettingsController) {
```
tractorcow: Better coding style is for this method to return a literal string, and for those subclasses to override this to return the specific key.

Parent classes, as a principle, should not know about specific subclasses.

robbieaverill: You're right. PR at https://github.com/silverstripe/silverstripe-cms/pull/1901

PR #1899, opened 2017-07-17
FIX Make sure VirtualPage renders correct templates

### LINKED ISSUE #1889: VirtualPage does not render content from orignal page correctly
**Per contrib guidelines:**
- Ask for assistance: Reported on forum by others users in 2009. See https://www.silverstripe.org/community/forums/customising-the-cms/show/5702 and https://www.silverstripe.org/community/forums/general-questions/show/8579
- Existing tickets: None I could find open or closed that describe the problem in these terms.
- Security issue: No.
- Reproduced on clean installation: Yes.
- Enhancement: No.

**Description:**
SS CMS: 3.6.1
OS: Ubuntu 14 & 16
PHP 5.5.9-1ubuntu4.21
mysqld  Ver 5.5.55-0ubuntu0.14.04.1

Pages created with VirtualPage only reproduce content from the default "Content" field and ignore the original pages template and custom fields. User expectation is that template and custom fields of original page will be used.

**Reproduction:**
Create: code/AdvancedPage.php
```php
class AdvancedPage extends Page {
	private static $description = 'A more advanced Page';

	private static $db = array(
					'Section2'    => 'HTMLText'
			);
		
	public function getCMSFields(){
			$fields = parent::getCMSFields();
			
			$fields->addFieldToTab('Root.Main', HtmlEditorField::create('Section2', 'Content for second part of the page'));
			
			return $fields;
	}
}

class AdvancedPage_Controller extends Page_Controller {

}
```

Create: templates/Layout/AdvancedPage.ss
```
<h1>Content</h1>
<div>$Content</div>
<h1>Section 2</h1>
<div>$Section2</div>
```
Create a page using the AdvancedPage layout and add content to both fields.
Create a VirtualPage pointing to the one you just made. Observe that only some of the content appears.

**Probable Related Issues:**
https://github.com/silverstripe/silverstripe-cms/issues/1336

dhensby:
fixes #1889

### DIFF
--- a/code/model/VirtualPage.php
+++ b/code/model/VirtualPage.php
@@ -496,10 +496,8 @@ class VirtualPage_Controller extends Page_Controller {
 	}
 	
 	public function getViewer($action) {
-		$originalClass = get_class($this->CopyContentFrom());
-		if ($originalClass == 'SiteTree') $name = 'Page_Controller';
-		else $name = $originalClass."_Controller";
-		$controller = new $name();
+		$object = $this->CopyContentFrom();
+		$controller = ModelAsController::controller_for($object, $action);
 		return $controller->getViewer($action);
 	}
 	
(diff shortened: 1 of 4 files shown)


heyrex:
Thank you @dhensby !

Issue #632, opened 2013-04-03
Page Search opens unexpanded tree view by default (should be list view)


silverstripe-issues:
created by: nicolaas
created at: 2013-01-01
original ticket: http://open.silverstripe.org/ticket/8163

---

when doing a search (filter) for a page type in the CMS (3.0.3), you find the relevant page, BUT the sitetree does not automatically open to the relevant page.

In some ways, I think it would be better if any searches open the gridfield view rather than the sitetree view. 

## AC

- List view is switched to when user searches for pages
- Display breadcrumb for search items
- Remove "Add page" action
- User has the ability to use the view toggle to navigate away from list view into tree view

### Excluded:

- Search field component as per files area to be implemented as separate issue
- Search notification updates which information about what has been searched to be implemented as separate issue

## Tasks

- [x] Switch to list view when in search view
- [x] Remove "Add page" in search view
- [x] Add breadrumbs

## Related

 * [Retain list view state](https://github.com/silverstripe/silverstripe-cms/issues/503)
 * [Page search should show filter criteria and highlight search filters](https://github.com/silverstripe/silverstripe-cms/issues/1891)

## Pull request

- [x] https://github.com/silverstripe/silverstripe-cms/pull/1902
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7196

chillu:
Deep default result:
![image](https://user-images.githubusercontent.com/111025/27845445-a7ff324e-6183-11e7-8300-fbea61fea790.png)

After FIVE expansion clicks:
![image](https://user-images.githubusercontent.com/111025/27845453-b32329e6-6183-11e7-8abf-7e10ca456dbb.png)

List view:
![image](https://user-images.githubusercontent.com/111025/27845460-c0c61d1a-6183-11e7-9c33-4c405a378f69.png)

I think we should fix this for 4.0, it makes page search on non-trivial page trees unusable by default. @clarkepaul What do you think?

clarkepaul:
Yeah if it doesn't overload the system, like if thousands of pages exist. Maybe if it was too many pages then we could alert the user and then allow them to view them in list view. Either way needs improving.

chillu:
[...]

PR #1902, opened 2017-07-19
FIX Pages search now defaults to list view

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-cms/issues/632

https://github.com/silverstripe/silverstripe-framework/pull/7196 needs to be merged first before the tests are passing.


code/Controllers/CMSMain.php
```diff
@@ -838,7 +838,7 @@ public function SearchForm()
         $dateGroup->setTitle(_t('SilverStripe\\CMS\\Search\\SearchForm.PAGEFILTERDATEHEADING', 'Last edited'));
 
         // view mode
-        $viewMode = HiddenField::create('view', false, $this->ViewState());
+        $viewMode = HiddenField::create('view', false, 'listview');
```
tractorcow: Instead of hard-coding this, we should make ViewState() default to listview, shouldn't we?

tractorcow:
Had to rebase

PR #1910, opened 2017-07-26
API Use setTitleField to use MenuTitle for selecting internal links

tractorcow:
For https://github.com/silverstripe/silverstripe-admin/issues/53

Depends on https://github.com/silverstripe/silverstripe-framework/pull/7206


dhensby:
behat test failure is consistent. I've restarted it twice and same error each time

tractorcow:
Yes because it's calling API that isn't merged yet. :)

dhensby:
D'oh

PR #1914, opened 2017-07-28
BUGFIX check if parent context is SiteTree instance

### LINKED ISSUE #1913: BUG SiteTree::canCreate() assumes parent context is SiteTree instance
In upgrading one of our tools we've found that applying a `many_many` relation  to `SiteTree` from a DataObject class causes `allowedChildren` to be checked on the DataObject class. [Tracing this back](https://github.com/silverstripe/silverstripe-framework/blob/4/src/Forms/GridField/GridFieldAddNewButton.php#L39-L50), I found that passing `$record` as `$context['Parent']` in this case causes `SiteTree::canCreate()` to assume the parent context is an instance of `SiteTree` and checks for `allowedChildren`.

I don't have the namespacing in here, but this is the gist of it:

```php
<?php

class MyObject extends DataObject{

    private static $many_many = [
        'Pages' => SiteTree::class,
    ];

    public function getCMSFields(){
        $fields = Parent::getCMSFields();
        $fields->removeByName('Pages');
        if ($this->owner->ID) {
            $config = GridFieldConfig_RelationEditor::create();
            $pages = $this->Pages();
            $linksField = GridField::create(
                'Pages',
                'Pages',
                $pages,
                $config
            );

            $fields->addFieldsToTab('Root.Pages', array(
                $linksField,
            ));
        }
        return $fields;
    }
}
```

muskie9:
fixes #1913

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -1155,7 +1155,8 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         // Check parent (custom canCreate option for SiteTree)
         // Block children not allowed for this parent type
         $parent = isset($context['Parent']) ? $context['Parent'] : null;
-        if ($parent && !in_array(static::class, $parent->allowedChildren())) {
+        $strictParentInstance = ($parent && $parent instanceof SiteTree);
+        if ($strictParentInstance && !in_array(static::class, $parent->allowedChildren())) {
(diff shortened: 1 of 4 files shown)


flamerohr:
thanks @muskie9 for the pull request, I would like to see this expanded a bit more for extensibility before it's merged :)

PR #1919, opened 2017-08-08
API Virtual pages now respect cascade_deletes on source page

tractorcow:
Fixes broken CMS tests on https://github.com/silverstripe/silverstripe-framework/pull/7257, but neither will pass in isolation.

Parent story: https://github.com/silverstripe/silverstripe-versioned/issues/23

### DIFF
--- a/tests/model/VirtualPageTest.php
+++ b/tests/model/VirtualPageTest.php
@@ -1,18 +1,16 @@
 <?php
 
-use SilverStripe\ORM\DataObject;
-use SilverStripe\Versioned\Versioned;
-use SilverStripe\ORM\DB;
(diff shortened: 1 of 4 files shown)


tests/model/VirtualPageTest.php
```diff
@@ -200,51 +198,12 @@ public function testCantPublishVirtualPagesBeforeTheirSource()
         $this->assertTrue($vp->canPublish());
     }
 
-    public function testCanDeleteOrphanedVirtualPagesFromLive()
```
flamerohr: ooh, no more orphaned virtual pages, no wonder it's no longer needed in the "broken links" report

PR #1930, opened 2017-08-23
BUG Fix “Show unpublished versions”

### LINKED ISSUE #1648: Show unpublished versions checkbox doesn't work in history
Occurs in master.

The checkbox to show unpublished versions no longer works in the history tab.

You can still view the unpublished versions by going direct to the URL of that version number.

tractorcow:
Update CMSPageHistoryController.js to fit coding conventions
Fixes #1648


tractorcow:
@phalkunz I've pushed up a fix that should address the issue you noted. It looks like the template was mixing $Published and $WasPublished, when $Published looks like the most reliable flag to use.

PR #1935, opened 2017-08-24
sanitize namespaced classes for page icon css

andrewandante:
Fixes #980

PR #1936, opened 2017-08-24
sanitize class names for cms icons

andrewandante:
Backport of https://github.com/silverstripe/silverstripe-cms/pull/1935 for 3.5


dhensby:
postgres failure appears to be a regression from the postgres module

PR #1948, opened 2017-09-01
FIX Update unused button classes for url segment field

sachajudd:


tractorcow:
Thanks, looks good @sachajudd !

Issue #1939, opened 2017-08-29
SS4: Cached sitetree is being displayed after editing a page


mikenz:
On a stock SS4 install:

Log in as admin and click on a couple of pages in the tree until it refreshes the tree (tree will flash).
Click on a published page (eg "Contact Us")
Edit the Page name (eg "Contact Us edited") and "Save" this.
The tree will show the new page name and badge (eg "Contact Us edited  MODIFIED")
Click on another page in the tree (eg "Page not found")
The tree will revert back to before the edit (eg "Contact Us")
Click on the edited page again the tree will be correct,
Click on another page the tree will return to the unedited cached version.

Caching/refreshing issue after editing?

chillu:
Hey Mike, thanks for reporting! I've had a hard time reproducing this a couple of days ago, hence didn't respond immediately. Note that this issue might currently be masked by a regression which triggers the `treeview` load on every page edit view: https://github.com/silverstripe/silverstripe-cms/issues/1949

FYI, you might be interested in installing the Zenhub.com Chrome Extension, which shows you a bit more activity around how we prioritise work :) I've moved this issue into our "Next" column 4 days ago, which means we'll get around to it very soon.

![image](https://user-images.githubusercontent.com/111025/30009464-af9634a2-917c-11e7-8a6b-eaefa0461c71.png)

flamerohr:
to replicate, you can do these steps:
1) open a page "Contact"
2) open a different page "About"
3) make a change and save on "About" so that the label "modified" appears
4) browse to "Contact" and you'll see the "modified" disappears for the "About" page

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-cms/pull/1950

sminnee:
@mikenz are you able to confirm that this is fixed for you in 4.x-dev now, in your project as well as a clean install?

PR #1950, opened 2017-09-04
BUG Prevent treeview loading repeatedly on each page edit form

### LINKED ISSUE #1949: Treeview XHR triggered on every page switch
It reloads the tree on every page edit view in the CMS, which will kill server performance. Likely a regression from recent performance-related fixed by @tractorcow 

image

Potentially related: https://github.com/silverstripe/silverstripe-cms/issues/1939

PR https://github.com/silverstripe/silverstripe-cms/pull/1950

tractorcow:
Fixes #1949 and #1939

Previously I'd added in the page ID to force the "current" page and avoid a session access, but this broke the caching of treeview so it's been reverted.

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -138,7 +138,6 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
     private static $url_handlers = [
         'EditForm/$ID' => 'EditForm',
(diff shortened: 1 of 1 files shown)


phalkunz:
You can get into a state where there are 2 items being highlighted.

silverstripe_-_edit_page

tractorcow:
I've fixed this. the `current` class is now provided by client only, not on the server.

PR #1952, opened 2017-09-04
API Allow SiteTree::Link to be extended

tractorcow:
This is necessary to allow post-baseurl applied links to be modified.

Useful for modules such as subsites / fluent which may need to prepend domain names.


tests/php/Model/SiteTreeTest_ExtensionA.php
```diff
@@ -13,4 +14,9 @@ public function canPublish($member)
     {
         return static::$can_publish;
     }
+
+    public function updateLink(&$link, $action = null)
+    {
+        $link = Controller::join_links($link, '?extra=1');
```
flamerohr: I'd like to see a unit test prepending to the link, to cover what you're intending for this extension point as well.

tractorcow:
Done @flamerohr :D

PR #1954, opened 2017-09-05
BUG Fix race condition with change detection / loading animation

### LINKED ISSUE #1859: "Save and Publish" button label overlays with loading indicators
When loading a page, editing it, and hitting "save" before change detection has a chance to kick in, the 
save button momentarily appears garbled.

image

# PRs

 - [x] https://github.com/silverstripe/silverstripe-admin/pull/222
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/1954

tractorcow:
Fixes #1859

Requires https://github.com/silverstripe/silverstripe-admin/pull/222 first


code/Model/SiteTree.php
```diff
@@ -2246,11 +2246,13 @@ public function getCMSActions()
         }
 
         // "save", supports an alternate state that is still clickable, but notifies the user that the action is not needed.
+        $noChangesClasses = 'btn-outline-secondary font-icon-check-mark';
```
flamerohr: does `btn-outline-secondary` need to be here if it's still hardcoded in the javascript?

code/Model/SiteTree.php
```diff
@@ -2246,11 +2246,13 @@ public function getCMSActions()
...
         if ($canEdit && $isOnDraft) {
             $majorActions->push(
                 FormAction::create('save', _t(__CLASS__.'.BUTTONSAVED', 'Saved'))
-                    ->addExtraClass('btn-outline-secondary font-icon-check-mark')
-                    ->setAttribute('data-btn-alternate', 'btn action btn-primary font-icon-save')
+                    ->addExtraClass($noChangesClasses)
+                    ->setAttribute('data-btn-alternate-add', 'btn-primary font-icon-save')
```
flamerohr: same for `btn-primary`

code/Model/SiteTree.php
tractorcow: The javascript code is there for legacy reasons; This is the canonical source of these class names.

code/Model/SiteTree.php
tractorcow: Soft coding over hard coding.

PR #1959, opened 2017-09-11
FIX Ensure client config merging includes existing parent "form" attributes

robbieaverill:
The existing code will overwrite any `$config['form']` values with its own. This change merges them instead.

Note - current state returns AddToCampaignForm and editorInternalLink. With this change it also returns EditorEmailLink and EditorExternalLink


tractorcow:
I wonder if it wouldn't just be safer to use a recursive array merge? (note: neither array_merge_recursive / array_replace_recursive would do this, it would need to be a custom replacer).

robbieaverill:
Good call - updated

tractorcow:
Sorry, I just corrected a typo

```dif
-would do this
+wouldn't do this
```

tractorcow:
Rebased, can be merged on green.

PR #1969, opened 2017-09-16
Reports: VirtualPages pointing to deleted pages report is incorrect

mikenz:
A virtual pointing to a deleted page is not the only way HasBrokenLink could be 1. An page extending VirtualPage could have a HTML field with a broken link in it but still point to a valid source page - this shouldn't be returned in the report.

### DIFF
--- a/code/Reports/BrokenVirtualPagesReport.php
+++ b/code/Reports/BrokenVirtualPagesReport.php
@@ -2,42 +2,45 @@
 
 namespace SilverStripe\CMS\Reports;
 
+use SilverStripe\CMS\Model\VirtualPage;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Forms\CheckboxField;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\ORM\DB;
-use SilverStripe\Versioned\Versioned;
 use SilverStripe\Reports\Report;
+use SilverStripe\Versioned\Versioned;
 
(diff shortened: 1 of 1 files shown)


code/Reports/BrokenVirtualPagesReport.php
```diff
@@ -2,42 +2,45 @@
...
     {
-        $classes = ClassInfo::subclassesFor('SilverStripe\\CMS\\Model\\VirtualPage');
+        $classes = ClassInfo::subclassesFor(VirtualPage::class);
         $classParams = DB::placeholders($classes);
-        $classFilter = array(
-            "\"ClassName\" IN ($classParams) AND \"HasBrokenLink\" = 1" => $classes
-        );
+        $classFilter = [
+            "\"ClassName\" IN ($classParams)" => $classes,
+            "\"VirtualPage\".\"CopyContentFromID\" NOT IN (SELECT \"ID\" FROM SiteTree)",
```
dhensby: `SiteTree` should be wrapped in double quotes `"`

maxime-rainville:
This sounds like a really edge casy scenario. I'd be curious to see a use case where you would want a virtual page with an HTML field.

maxime-rainville:
I'll close this for now as having a Virtual Page with an HTML Field sounds like a contradiction.

I'm happy to reopen if someone wants to discuss it further.

PR #1973, opened 2017-09-20
Update usages of ClassInfo

tractorcow:
Add missing unit tests for SiteTreeFolderExtension

Required for https://github.com/silverstripe/silverstripe-framework/pull/7386, but should work in isolation as well.

Parent stories:
- https://github.com/silverstripe/silverstripe-framework/issues/7385
- https://github.com/silverstripe/silverstripe-admin/issues/199

### DIFF
--- a/code/Controllers/SilverStripeNavigator.php
+++ b/code/Controllers/SilverStripeNavigator.php
@@ -43,8 +43,8 @@ class SilverStripeNavigator extends ViewableData
     {
         $items = array();
 
-        $classes = ClassInfo::subclassesFor('SilverStripe\\CMS\\Controllers\\SilverStripeNavigatorItem');
-        unset($classes['SilverStripe\\CMS\\Controllers\\SilverStripeNavigatorItem']);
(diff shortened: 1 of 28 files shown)


code/Controllers/SilverStripeNavigator.php
```diff
@@ -43,8 +43,8 @@ public function getItems()
     {
         $items = array();
 
-        $classes = ClassInfo::subclassesFor('SilverStripe\\CMS\\Controllers\\SilverStripeNavigatorItem');
-        unset($classes['SilverStripe\\CMS\\Controllers\\SilverStripeNavigatorItem']);
+        $classes = ClassInfo::subclassesFor(SilverStripeNavigatorItem::class);
+        array_shift($classes);
```
flamerohr: nice one, so it doesn't need to be aware of the case sensitivity work

tractorcow: Yes that's how all the code should be.

PR #1983, opened 2017-10-04
Fix: TreeDropdownField showing broken page icons (fixes silverstripe/framework#7420)

kinglozzer:
Removing this class hides the page type icons in the “unexpanded” TreeDropdownField (they still show up when you expand it), restoring the old behaviour.

Looks like a regression from https://github.com/silverstripe/silverstripe-cms/pull/1936


tractorcow:
PG issue doesn't seem related.

PR #1985, opened 2017-10-08
Encode breadcrumbs links as plain text

wilr:
Requires https://github.com/silverstripe/silverstripe-admin/pull/291/


tractorcow:
Text is already encoded as plain by default; Explicit cast here isn't needed.

PR #1988, opened 2017-10-11
Move cascading delete to onAfterDelete for SiteTree

unclecheese:

### DIFF
--- a/code/model/SiteTree.php
+++ b/code/model/SiteTree.php
@@ -1575,7 +1575,7 @@ class SiteTree extends DataObject implements PermissionProvider,i18nEntityProvid
 		parent::onBeforeDelete();
 
 		// If deleting this page, delete all its children.
-		if(SiteTree::config()->enforce_strict_hierarchy && $children = $this->AllChildren()) {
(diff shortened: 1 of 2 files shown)


dhensby:
@kinglozzer good question. I think the bug is quite severe and should probably be patched in 3.5+ but this does feel like an API change...

Very few people have reported this being a problem so whilst the bug has the capability to be very damaging it seems to surface very infrequently.

On balance, I'd probably stick with 3.x branch given that it feels like more than a simple patch and that the bug seems to be relatively infrequently encountered.

tractorcow:
Moving the cascade delete into the "after" introduces a rather concerning shift in integrity I must point out. I can't see how this is necessary in order to simply avoid deleting all records, when it seems all that's necessary is an `isInDB()` check on children enumeration.

If you delete parents first, but fail to delete a child, you end up with your application in an inconsistent state. It needs to be leaf-first not node-first to ensure this consistency.

unclecheese:
Closing. Change needs to be made to `Hierarchy`, not site tree.

PR #1989, opened 2017-10-12
Move cascading delete to onAfterDelete for SiteTree

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-framework/issues/5014#issuecomment-335975688

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -1481,9 +1481,10 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         $this->extend('augmentSyncLinkTracking');
     }
 
-    public function onBeforeDelete()
+    public function onAfterDelete()
     {
-        parent::onBeforeDelete();
+        $this->updateDependentPages();
+        parent::onAfterDelete();
 
         // If deleting this page, delete all its children.
         if (SiteTree::config()->enforce_strict_hierarchy && $children = $this->AllChildren()) {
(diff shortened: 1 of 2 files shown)


tests/php/Model/SiteTreeTest.php
```diff
@@ -332,6 +332,26 @@ public function testRestoreToStage()
...
+        $fail = true;
+        try {
+            $obj->delete();
+        } catch (\LogicException $e) {
+            $fail = false;
+        }
+        if ($fail) {
+            $this->fail('Failed to throw delete exception');
+        }
+        $this->assertEquals($count, SiteTree::get()->count());
```
dhensby: `assertCount` please

tests/php/Model/SiteTreeTest.php
```diff
@@ -332,6 +332,26 @@ public function testRestoreToStage()
         $this->assertInstanceOf('Page', $requeriedPage);
     }
 
+    public function testNoCascadingDeleteWithoutID()
+    {
+        Config::nest();
```
dhensby: `Config` nesting is not needed as the test suite does this itself

code/Model/SiteTree.php
```diff
@@ -1481,9 +1481,10 @@ public function syncLinkTracking()
         $this->extend('augmentSyncLinkTracking');
     }
 
-    public function onBeforeDelete()
+    public function onAfterDelete()
```
tractorcow: As commented on #1988 

Moving the cascade delete into the "after" introduces a rather concerning shift in integrity I must point out. I can't see how this is necessary in order to simply avoid deleting all records, when it seems all that's necessary is an `isInDB()` check on children enumeration.

If you delete parents first, but fail to delete a child, you end up with your application in an inconsistent state. It needs to be leaf-first not node-first to ensure this consistency.

unclecheese:
Closing as the overall approach violates data integrity. The only change needed to fix this is a change to hierarchy.

PR #1990, opened 2017-10-12
BUG UploadField overwriteWarning isn't working in AssetAdmin

jason-zz:
Enable UploadField overwriteWarning

```yaml
Upload:
  replaceFile: true
UploadField:
  defaultConfig:
    overwriteWarning: true
```

In assets admin, when you upload files with same name in non-root folder, it just replaces the file and there is no overwrite warning message.

This fix will let CMSFileAddController know the current folder when 'fileexists' AJAX request is called.

### DIFF
--- a/code/controllers/CMSFileAddController.php
+++ b/code/controllers/CMSFileAddController.php
@@ -102,6 +102,11 @@ class CMSFileAddController extends LeftAndMain {
 			)
 		);
 		$form->loadDataFrom($folder);
+		
+		if($this->currentPageID()){
+		    // Make sure this controller know current folder when AJAX 'fileexists' is fired.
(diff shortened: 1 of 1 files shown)


dhensby:
Hi @jason-zz - thanks a lot for this contribution! Can you confirm if this is an issue on 3.5 or just 3.6 and up?

jason-zz:
Hi @dhensby , this is an issue for 3.5 as well. I had a look at the history of CMSFileAddController.php. It seems this problem exists since version 3. 

Tested with 3.1 and it's having this problem as well.

dhensby:
OK - thanks, 3.1 is end-of-life and not supported anymore, 3.5 is the lowest supported release at the moment, so I'll rebase this work onto 3.5 for you.

dhensby:
test failures are pre-existing

PR #1993, opened 2017-10-16
Force SiteTree to check in DB before deleting children

unclecheese:


tractorcow:
Rebased.

PR #1994, opened 2017-10-16
Force SiteTree to check in DB before deleting children

unclecheese:


tractorcow:
Closed with https://github.com/silverstripe/silverstripe-cms/pull/1988, which seems to have the correct commit rebased.

PR #1996, opened 2017-10-17
BUG Fix page icons in vendor modules

tractorcow:
cc @axllent


tractorcow:
Tests weren't working if cms installed as root, and there was a path case sensitivity issue. These are fixed now. :)

tractorcow:
Ah, tests didn't work when cms was module root. Fixed now.

tractorcow:
Fixed. :)

code/Controllers/LeftAndMainPageIconsExtension.php
```diff
@@ -32,28 +35,25 @@ public function generatePageIconsCss()
...
+                $iconURL = $iconResource->getURL();
+                $css .= "$selector { background: transparent url('$iconURL') 0 0 no-repeat; }\n";
+            } elseif (Director::fileExists($icon)) {
+                // Support direct path
+                $iconURL = ModuleResourceLoader::resourceURL($icon);
+                $css .= "$selector { background: transparent url('$iconURL') 0 0 no-repeat; }\n";
             } else {
                 // Support for more sophisticated rules, e.g. sprited icons
-                $css .= "$selector { $iconFile }\n";
+                $css .= "$selector { $icon }\n";
```
flamerohr: should this be sanitised?
I'm a bit wary of someone injecting css code here

tractorcow: I think we may need this in Requirements_Backend

```diff
// Literal custom CSS content
foreach ($this->getCustomCSS() as $css) {
+    $css = Convert::raw2xml($css);
    $requirements .= HTML::createTag('style', ['type' => 'text/css'], "\n{$css}\n");
    $requirements .= "\n";
}
```

But separate story to this.

tractorcow: Looks like it's not that easy.

Possible answer https://stackoverflow.com/questions/3241616/sanitize-user-defined-css-in-php, but requires a thirdparty lib.

tractorcow: Let's dump this risky behaviour. It's a risk identified on owasp so let's at least minimise it https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet#HTML_entity_encoding

PR #2001, opened 2017-10-24
Separate logic for whether to show the link text from the renderModal method

robbieaverill:
Separates some heavy TinyMCE coupling from the modal popup entwine logic so it can be overloaded easily in other contexts.


client/src/legacy/TinyMCE_sslink-internal.js
```diff
@@ -86,6 +83,20 @@ jQuery.entwine('ss', ($) => {
     },
 
     /**
+     * Determine whether to show the link text field
+     *
+     * @return {Boolean}
+     */
+    getRequireLinkText() {
+      const selection = tinymce.activeEditor.selection;
```
flamerohr: noticed this bug here
I'll rebase and correct this too - should be `this.getElement().getEditor().getInstance().selection`

PR #2004, opened 2017-10-27
FIX Add warning state to revert action in CMS page history

robbieaverill:


tractorcow:
Before:

image

New:

image

@clarkepaul can you please approve this change?

Issue #1977, opened 2017-09-26
SiteTree load extremely inefficient when you have many page types + many pages in tree (~2k calls to ->canCreate)


patricknelson:
My site currently has **54** page types and **36** pages visible at the first two levels. This means that SilverStripe performs **1,944** unique hits to `->canCreate($member, $context)`. It calls that method for every single page at those first two levels for every single page type.  The end result is that page load times *just* for the requests going to `/admin/pages/treeview` end up being between 3 to 7 seconds, depending on the server (more on local dev, less on prod). I was able to verify this as the bottle neck when I could reduce the high 7 second figure down to about 1.5 seconds when I cached the first response from `->canCreate()` (which of course is not properly accounting for parent context).

I can see this getting even more out of hand. As time goes by, the number of calls to `->canCreate()` will only continue to increase by 36x for every new page type and 54x for every new page at the first two levels.  Ultimately, this seems inefficient to me because the only apparent purpose of this is to facilitate visibility of options in a right click menu. This is a relatively rarely used option, given that this overhead occurs repeatedly on _every single load_ of the site tree.

Possible solutions:

- Perform some sort of **caching**, which varies (that is, `Vary` in the caching sense) by:
    - Page type
    - Parent page instance (or just type? currently instance is passed)
    - Member
    - Also: Need to think about how to bust cache in case of custom implementations  of `->canCreate(...)`  which may rely on unknown external factors.
- Generate **on-demand**: Completely re-think the use of this sort of option in the right click menu. Maybe only generate it on-demand after the user right clicks?

This may not be worthwhile for most smaller sites, however it becomes noticeable for us as our site is fairly large both in complexity (number of page types) and in scale (number of pages) as well as the number of back-end users (and thus impacting overall server load).

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2048
- [x] https://github.com/silverstripe/silverstripe-frameworktest/pull/45

patricknelson:
BTW: The main culprit is here in `SiteTree->getTreeTitle()` and it affects both `v3` and `v4`:

https://github.com/silverstripe/silverstripe-cms/blob/4/code/Model/SiteTree.php#L2637

chillu:
I've marked this as `impact/high`, but since it's not a regression from 3.x and we're got our hands pretty full with getting 4.0 stable at the moment, the implementation of this would likely have to wait until later in the 4.x release cycle (assuming it can be done without API changes).

[...]

PR #2029, opened 2017-11-16
Make `CMSMain::getSiteTreeFor()` default to `null` nodeCount

oilee80:
This speeds the AJAX request from 3.81s to 1.7s

Also should only return the top level pages otherwise its just assuming and thus wasting resources

This also allows it to be fully controlled via config whereas for the `/admin/pages/treeview` AJAX call it was defaulting to 30 as no parameter was passed

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -53,6 +53,7 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 use SilverStripe\ORM\HiddenClass;
+use SilverStripe\ORM\Hierarchy;
 use SilverStripe\ORM\Hierarchy\MarkedSet;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\ValidationResult;
@@ -108,8 +109,6 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
     private static $session_namespace = self::class;
 
     private static $required_permission_codes = 'CMS_ACCESS_CMSMain';
-
-    private static $default_node_count_threshold = 30;
 
     /**
      * Amount of results showing on a single page.
@@ -467,7 +466,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         $filterFunction = null,
         $nodeCountThreshold = null
     ) {
-        $nodeCountThreshold = is_null($nodeCountThreshold) ? Config::inst()->get(get_class(), 'default_node_count_threshold') : $nodeCountThreshold;
+        $nodeCountThreshold = is_null($nodeCountThreshold) ? Config::inst()->get(Hierarchy::class, 'node_threshold_total') : $nodeCountThreshold;
         // Provide better defaults from filter
         $filter = $this->getSearchFilter();
         if ($filter) {
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -21,6 +21,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Environment;
 use SilverStripe\Core\Injector\Injector;
@@ -107,6 +108,8 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
     private static $session_namespace = self::class;
 
     private static $required_permission_codes = 'CMS_ACCESS_CMSMain';
+
+    private static $default_node_count_threshold = 30;
 
     /**
      * Amount of results showing on a single page.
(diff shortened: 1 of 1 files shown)


oilee80:
@ninty9notout @dhensby @andrewandante for your consideration

andrewandante:
I mean, it seems sensible to me - but I'm sure there was a reason 30 was set as the minimum. I guess it pre-loads the SiteTree, so we are trading up-front loading time for loading time per-click? Am I understanding it correctly?

Also good that it can be controlled via config with:

```yml
SilverStripe\CMS\Model\SiteTree:
  node_threshold_total: 30
```

And maybe we want to set this config value on SiteTree so that it's a BC change, and document how to adjust?

oilee80:
Hi @andrewandante 
Yes this does me trading upfront load to a per-click but at least the time is being spent on what you're asking for.
For the site I'm working on I've also set
```yml
SilverStripe\ORM\Hierarchy\Hierarchy:
  node_threshold_total: 1
```
so as to not load the children on the subtree, but it's not possible with `CMSMain::treeview` as `CMSMain::SiteTreeAsUL()` doesn't pass the extra parameters.

dhensby:
@silverstripe/core-team

jonom:
3.81s is a bit of an eye-opener... can we optimise the rendering of each node instead of reducing the volume?

tractorcow:
I can't reproduce the supposed performance gains of 1.1 seconds, so I'm hesitant to approve or critique this change without knowing what's actually being improved.

The default `node_threshold_total` in Hierarchy is 50, so is the change from 30 to (supposedly) 50 an actual performance benefit?

tractorcow:
@oilee80 maybe if you give me the scale of your site tree locally I can reproduce your testing and verify the performance gain. :)

I will test with the node_threshold_total = 1 config you've supplied also.

sminnee:
The specific time benefits will depend on your tree. It's assuming that it's okay to pre-fetch 2nd level pages and so on up to a threshold because otherwise you have a 2nd load when you expand the tree, and when you're opening the CMS it's pretty rare that you want to edit the top-level page.

The current design was built on the assumption that the round trip to the server, SS bootstrap, etc, was non-trivial compared to the cost of rendering the tree nodes.

Like @jonom I think the underlying issue here is a sluggish tree-node render. As such, this gets a :-1: from me.

sminnee:
It looks like #1977 and #7442 already cover the underlying issue.

oilee80:
Hi All,
I agree that there is a lower level issue with the rendering of the individual nodes that should also be dealt with probably via the 2 Issues raised by @sminnee.

However in the meantime I've done an amend that will resolve the issue for the short term by making the `$nodeCountThreshold` property controllable via Config as we and no doubt others will still not want the second level rendered straight away in the Tree and a second call will be quicker overall.

code/Controllers/CMSMain.php
```diff
@@ -462,8 +465,9 @@ public function getSiteTreeFor(
         $childrenMethod = null,
         $numChildrenMethod = null,
         $filterFunction = null,
-        $nodeCountThreshold = 30
+        $nodeCountThreshold = null
     ) {
+        $nodeCountThreshold = is_null($nodeCountThreshold) ? Config::inst()->get(get_class(), 'default_node_count_threshold') : $nodeCountThreshold;
```
tractorcow: Instead of adding a new config to CMSMain (new API) can we use the SiteTree config? This is pre-existing and means fewer changes CMS developers need to make to update a hierarchy's behaviour.

```
if (!isset($nodeCountThreshold)) {
	$nodeCountThreshold = Config::inst()->get($className, 'node_threshold_total');
}
```

code/Controllers/CMSMain.php
```diff
@@ -462,8 +464,9 @@ public function getSiteTreeFor(
         $childrenMethod = null,
         $numChildrenMethod = null,
         $filterFunction = null,
-        $nodeCountThreshold = 30
+        $nodeCountThreshold = null
     ) {
+        $nodeCountThreshold = is_null($nodeCountThreshold) ? Config::inst()->get(Hierarchy::class, 'node_threshold_total') : $nodeCountThreshold;
```
tractorcow: Hierarchy will only provide the default, which can be overridden on a $className basis, hence the suggestion.

oilee80:
Hi @tractorcow I've pushed the change you requested but it doesn't seem to be overridable on a `$className` basis.

e.g. with the following in my `site.yml` file I would expect the value to to 2 as `SilverStripe\CMS\Model\SiteTree` is the value of `$className` but's using 1. Or do YAML configs work differently for extensions?
```yml
SilverStripe\ORM\Hierarchy\Hierarchy:
  node_threshold_total: 1
SilverStripe\CMS\Model\SiteTree:
  node_threshold_total: 2
```

tractorcow:
No, it should be 2. Did you flush?

tractorcow:
Sorry, you are right... extensions override the class they are directly applied to.

However, you can still use Config::modify() to override both base and extension values, so I think the change is still an improvement. :)

tractorcow:
Proof of concept: This gets you the exact per-sitetree config you want.

```php
<?php

use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\Core\Config\Config;

Config::modify()->set(SiteTree::class, 'node_threshold_total', 2);
```

tractorcow:
Another proof of concept, add another extension on SiteTree to modify the threshold :) Not perfect I'm afraid!

Issue #2031, opened 2017-11-17
Add page filter doesn't update when new PageType selected


andrewandante:
4.0.0 Stable

When adding a page through the Add Page dialogue, if you select "Under another page", the pages you can select will not update based on `$allowed_children`. 

To reproduce:

- Blank install of SS4 stable.
- Create Article Holder and Article Page classes as per lessons, i.e.

```php
class ArticleHolder extends Page
{
    private static $allowed_children = [
        ArticlePage::class
    ];

}

class ArticlePage extends Page
{
    private static $can_be_root = false;
}
```

- Create a top-level ArticleHolder
- Attempt to create an ArticlePage below the ArticleHolder page.

From what I can see, there is no JS firing when the pagetype is selected from the tree-dropdown field.

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-cms/pull/2032

PR #2032, opened 2017-11-19
BUG Fix allowed children types now load properly

flamerohr:
FIxes https://github.com/silverstripe/silverstripe-cms/issues/2031


tractorcow:
updateTypeList doesn't trigger on change anymore, which seems to be a bug... I'm looking into this.

PR #2033, opened 2017-11-19
Update CMSMain.php

SpiritLevel:
Tidy up :)

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -784,7 +784,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
     public function CanOrganiseSitetree()
     {
-        return !Permission::check('SITETREE_REORGANISE') && !Permission::check('ADMIN') ? false : true;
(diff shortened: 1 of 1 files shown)


code/Controllers/CMSMain.php
```diff
@@ -784,7 +784,7 @@ public function savetreenode($request)
 
     public function CanOrganiseSitetree()
     {
-        return !Permission::check('SITETREE_REORGANISE') && !Permission::check('ADMIN') ? false : true;
+        return ( Permission::check('SITETREE_REORGANISE') || Permission::check('ADMIN') );
```
tractorcow: There is no case where SITETREE_REORGANISE would be false but ADMIN would return true; You should just make it the simple case.

maxime-rainville:
I'll close this for now as it looks like the matching repo for the PR is gone. So there's no way to answer the feedback.

It's still a pretty good fix, so I created a PR implementing @tractorcow's feedback https://github.com/silverstripe/silverstripe-cms/pull/2363.

Issue #2036, opened 2017-11-23
travis cms behat tests currently broken


tractorcow:
As per artifacts available at https://s3.amazonaws.com/silverstripe-travis-artifacts/silverstripe/silverstripe-cms/304489358/304489362/index.html

Error output from selenium shows:

```
Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
Nov 23, 2017 1:01:29 AM org.openqa.grid.selenium.GridLauncher main
INFO: Launching a standalone server
01:01:29.967 INFO - Java: Oracle Corporation 25.144-b01
01:01:29.969 INFO - OS: Linux 4.9.6-040906-generic amd64
01:01:29.983 INFO - v2.41.0, with Core v2.41.0. Built from revision 3192d8a
01:01:30.074 INFO - Default driver org.openqa.selenium.ie.InternetExplorerDriver registration is skipped: registration capabilities Capabilities [{ensureCleanSession=true, browserName=internet explorer, version=, platform=WINDOWS}] does not match with current platform: LINUX
01:01:30.177 INFO - RemoteWebDriver instances should connect to: http://127.0.0.1:4444/wd/hub
01:01:30.178 INFO - Version Jetty/5.1.x
01:01:30.181 INFO - Started HttpContext[/selenium-server,/selenium-server]
01:01:30.242 INFO - Started org.openqa.jetty.jetty.servlet.ServletHandler@2eee9593
01:01:30.242 INFO - Started HttpContext[/wd,/wd]
01:01:30.242 INFO - Started HttpContext[/selenium-server/driver,/selenium-server/driver]
01:01:30.242 INFO - Started HttpContext[/,/]
01:01:30.247 INFO - Started SocketListener on 0.0.0.0:4444
01:01:30.247 INFO - Started org.openqa.jetty.jetty.Server@30dae81
01:01:31.876 INFO - Executing: [new session: Capabilities [{build=3146, browser=firefox, tunnel-identifier=3146.4, name=Behat feature suite, browserName=firefox, ignoreZoomSetting=false, tags=[Travis-CI, PHP 7.0.22]}]] at URL: /session)
01:01:31.895 INFO - Creating a new session for Capabilities [{build=3146, browser=firefox, tunnel-identifier=3146.4, name=Behat feature suite, browserName=firefox, ignoreZoomSetting=false, tags=[Travis-CI, PHP 7.0.22]}]
01:01:34.104 WARN - Exception thrown
[...]

PR #2037, opened 2017-11-23
Update .travis.yml in line with framework .travis.yml

tractorcow:
Might fix https://github.com/silverstripe/silverstripe-cms/issues/2036 by minimising differences between broken CMS build and working framework build.


tractorcow:
I have no idea how but this fixes behat?

PR #2041, opened 2017-11-28
FIX: VirtualPage not using target page's template (fixes #2039)

### LINKED ISSUE #2039: silverstripe 4 VirtualPage doesn't find correct template
_From @cansozeri on November 27, 2017 21:27_

I have created a page from Article page type and I have Article.ss Layout. 

I have created a virtual page that points to Article page. But VirtualPage doesn't find correct template and use Page.ss when loading this virtual link. Content loaded but with page.ss not with article.ss.

_Copied from original issue: silverstripe/silverstripe-framework#7652_

kinglozzer:
Requires: https://github.com/silverstripe/silverstripe-framework/pull/7668#discussion_r154855427

Two commits/fixes here.

First one (ref. `refreshFromCopied()`) is because when you write a `VirtualPage` for the first time, `SiteTree::onBeforeWrite()` will check for a duplicate URL and fix it, but then `refreshFromCopied()` was running _afterwards_ and overwriting it with the target page’s URL.

Second one is the actual fix for #2039. Introduces a new API `SiteTree::getViewerTemplates()` which I’m happy to take feedback on, I just can’t see any way of solving this without that extra API (or hardcoded checks for `instanceof VirtualPage` in `ContentController`).

Given that this is a bugfix (and it severely impacts the usefulness of `VirtualPage`) I think a new API is acceptable in a patch release to fix this.

### DIFF
--- a/code/Model/VirtualPage.php
+++ b/code/Model/VirtualPage.php
@@ -277,8 +277,8 @@ class VirtualPage extends Page
 
     public function onBeforeWrite()
     {
-        parent::onBeforeWrite();
         $this->refreshFromCopied();
+        parent::onBeforeWrite();
     }
 
     /**
--- a/code/Model/VirtualPage.php
+++ b/code/Model/VirtualPage.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\CMS\Model;
 
+use Page;
 use SilverStripe\Core\Convert;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\LiteralField;
@@ -9,9 +10,9 @@ use SilverStripe\Forms\ReadonlyTransformation;
 use SilverStripe\Forms\TreeDropdownField;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\ValidationResult;
-use SilverStripe\Versioned\Versioned;
 use SilverStripe\Security\Member;
-use Page;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\View\SSViewer;
 
 /**
  * Virtual Page creates an instance of a  page, with the same fields that the original page had, but readonly.
@@ -366,6 +367,22 @@ class VirtualPage extends Page
     public function CMSTreeClasses()
     {
         return parent::CMSTreeClasses() . ' VirtualPage-' . $this->CopyContentFrom()->ClassName;
+    }
+
+    /**
+     * Use the target page's class name for fetching templates - as we need to take on its appearance
+     *
+     * @param string $suffix
+     * @return array
+     */
+    public function getViewerTemplates($suffix = '')
+    {
+        $copy = $this->CopyContentFrom();
+        if ($copy && $copy->exists()) {
+            return $copy->getViewerTemplates($suffix);
+        }
+
+        return parent::getViewerTemplates($suffix);
     }
 
     /**
--- a/code/Controllers/ContentController.php
+++ b/code/Controllers/ContentController.php
@@ -440,16 +440,24 @@ HTML;
             $action = '_' . $action;
         }
 
-        $templates = array_merge(
-            // Find templates by dataRecord
-            SSViewer::get_templates_by_class(get_class($this->dataRecord), $action, "SilverStripe\\CMS\\Model\\SiteTree"),
-            // Next, we need to add templates for all controllers
-            SSViewer::get_templates_by_class(static::class, $action, "SilverStripe\\Control\\Controller"),
-            // Fail-over to the same for the "index" action
-            SSViewer::get_templates_by_class(get_class($this->dataRecord), "", "SilverStripe\\CMS\\Model\\SiteTree"),
-            SSViewer::get_templates_by_class(static::class, "", "SilverStripe\\Control\\Controller")
-        );
+        $templatesFound = [];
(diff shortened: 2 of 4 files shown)


dhensby:
Why can't we solve this on the controller level alone? I'm don't see why we need the method on `SiteTree` and `VirtualPage` - can't we have it on `ContentController` and `VirtualPageController`?

kinglozzer:
@dhensby Because there is no `VirtualPageController` 😉

cansozeri:
I am waiting for composer update :=)

dhensby:
> @dhensby Because there is no VirtualPageController 😉 . The other option is to hard-code instanceof VirtualPage checks in ContentController::getViewer(), but... yuck...

I think adding this API to the model isn't the right approach, if there's no controller, maybe there needs to be one??

kinglozzer:
> I think adding this API to the model isn't the right approach, if there's no controller, maybe there needs to be one??

`VirtualPage_Controller` was _removed_ for 4.x, because it caused so many issues with `$allowed_actions`, `init()` and so on, so that would be a step backward IMO. It is a little strange to be adding extra view-related API to the model, but given that it explicitly _depends_ on the model (because the templates we need will vary based on chosen page) I can’t see a better way of doing it

code/Model/VirtualPage.php
```diff
@@ -368,6 +369,22 @@ public function CMSTreeClasses()
...
+     * Use the target page's class name for fetching templates - as we need to take on its appearance
+     *
+     * @param string $action
+     * @return array
+     */
+    public function getViewerTemplates($action = '')
+    {
+        $copy = $this->CopyContentFrom();
+        if ($copy && $copy->exists()) {
+            return SSViewer::get_templates_by_class($copy->ClassName, $action, SiteTree::class);
```
flamerohr: would `get_class($copy)` work for this case?

oh, possibly because of VirtualPage inception?

kinglozzer: No I think that should work actually, I’ll give it a shot tomorrow morning unless someone wants to push to my fork in the meantime

tractorcow: Looks good overall, maybe just make that change and we'll approve.

dhensby:
Ok, if we are adding this API to the model, should it not be on viewable data rather than sitetree?

kinglozzer:
> Ok, if we are adding this API to the model, should it not be on viewable data rather than sitetree?

We’d still need it in `SiteTree` so that we can pass `SiteTree::class` as the third arg to `get_templates_by_class()`, but I can’t see any reason we couldn’t introduce it to `ViewableData` too as a helper. I’ll make that change

tractorcow:
> We’d still need it in SiteTree so that we can pass SiteTree::class as the third arg to get_templates_by_class(), but I can’t see any reason we couldn’t introduce it to ViewableData too as a helper. I’ll make that change in a framework PR for 4.1

How about: 

```
ViewableData::getViewerTemplates() // uses self::class
DataObject::getViewableTemplates() // uses $this->baseClass();
SiteTree::getViewerTemplates() // remove
VirtualPage::getViewerTemplates() // as-is
```

code/Model/VirtualPage.php
```diff
@@ -368,6 +369,22 @@ public function CMSTreeClasses()
...
+     * Use the target page's class name for fetching templates - as we need to take on its appearance
+     *
+     * @param string $suffix
+     * @return array
+     */
+    public function getViewerTemplates($suffix = '')
+    {
+        $copy = $this->CopyContentFrom();
+        if ($copy && $copy->exists()) {
+            return SSViewer::get_templates_by_class(get_class($copy), $suffix, SiteTree::class);
```
tractorcow: `$copy->getViewerTemplates($suffix);` would be more respectful to custom implementations also.

code/Model/SiteTree.php
```diff
@@ -2730,6 +2730,17 @@ public function getControllerName()
...
 
+    /**
+     * Find appropriate templates for SSViewer to use to render this page
+     *
+     * @param string $suffix
+     * @return array
+     */
+    public function getViewerTemplates($suffix = '')
+    {
+        return SSViewer::get_templates_by_class(get_class($this), $suffix, self::class);
```
tractorcow: As per my comment, maybe a ViewableData method that stops at self::class, and DataObject method that stops at `$this->baseClass()`

tractorcow:
@kinglozzer if you want to make a bugfix that requires a new API, I don't mind if you make the entire PR at 4.1. It may be less urgent, as you can monkey-patch this issue in 4.0 with code in PageController::getViewer(), can we not? I'm happy for you to make the final call on the branch though. :)

kinglozzer:
Framework PR: https://github.com/silverstripe/silverstripe-framework/pull/7668, builds are going to fail until that is merged (tests are passing locally).

I’ve left this against 4.0. VirtualPage is rendered useless without this fix, and I don’t think semver is intended to block bugfixes (or make them hack-ier than they need to be) 😄

tractorcow:
I've left a comment on https://github.com/silverstripe/silverstripe-framework/pull/7668 requesting feedback on the semver concern.

tractorcow:
Framework merged, I'll restart cms soon.

kinglozzer:
❤️

PR #2044, opened 2017-11-29
File Form missing CMS Usage information

mikenz:


flamerohr:
Thanks @mikenz :D

We had originally disabled it in hopes of (in the planning stages currently) standardising an API/data for react to render the content properly - as opposed to dropping in server-rendered HTML.

We should probably get this in as a temporary measure anyway... @chillu ?

mikenz:
@flamerohr History was moved behind a config option for that reason. I think Usage was probably lost accidentally. The code was still present but extending a form that no longer exists.

tractorcow:
I rebased this and gave it a tiny cleanup so that it looked a bit better in the admin section. Approved and can merge on green.

PR #2048, opened 2017-12-04
ENHANCEMENT: Cache canCreate in CMSMain

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/1977

## Benchmarks

(-) CMSMain cache
(-) SiteTree cache
* 15.5s wall time
* 32.1MB memory
* 1.45s response time

(+) CMSMain cache
(-) SiteTree cache
* 15.3s wal
* 30.6MB
* 1.2s

(+) CMSMain cache
(+) SiteTree cache
* 7.5s
* 0.9s
* 31MB

### DIFF
--- a/_config/permissions.yml
+++ b/_config/permissions.yml
@@ -6,7 +6,6 @@ SilverStripe\Core\Injector\Injector:
     class: SilverStripe\Security\InheritedPermissions
     constructor:
       BaseClass: SilverStripe\CMS\Model\SiteTree
-      CacheService: '%$Psr\SimpleCache\CacheInterface.InheritedPermissions'
     properties:
       DefaultPermissions: '%$SilverStripe\SiteConfig\SiteConfigPagePermissions'
       GlobalEditPermissions:
--- a/_config/permissions.yml
+++ b/_config/permissions.yml
@@ -6,6 +6,7 @@ SilverStripe\Core\Injector\Injector:
     class: SilverStripe\Security\InheritedPermissions
     constructor:
       BaseClass: SilverStripe\CMS\Model\SiteTree
+      CacheService: '%$Psr\SimpleCache\CacheInterface.InheritedPermissions'
     properties:
       DefaultPermissions: '%$SilverStripe\SiteConfig\SiteConfigPagePermissions'
       GlobalEditPermissions:
@@ -15,3 +16,5 @@ SilverStripe\Core\Injector\Injector:
     properties:
       Services:
         - '%$SilverStripe\Security\PermissionChecker.sitetree'
+        - '%$SilverStripe\CMS\Controllers\CMSMain'
+        - '%$SilverStripe\CMS\Model\SiteTree'
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -25,8 +25,6 @@ use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Environment;
 use SilverStripe\Core\Injector\Injector;
-use SilverStripe\Core\Manifest\ModuleResource;
-use SilverStripe\Core\Manifest\ModuleResourceLoader;
 use SilverStripe\Forms\DateField;
 use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\FieldGroup;
@@ -53,7 +51,6 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 use SilverStripe\ORM\HiddenClass;
-use SilverStripe\ORM\Hierarchy;
 use SilverStripe\ORM\Hierarchy\MarkedSet;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\ValidationResult;
@@ -69,6 +66,8 @@ use SilverStripe\Versioned\ChangeSetItem;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\View\ArrayData;
 use SilverStripe\View\Requirements;
+use SilverStripe\Core\Flushable;
+use SilverStripe\Core\Cache\MemberCacheFlusher;
 use Translatable;
 
 /**
@@ -81,7 +80,7 @@ use Translatable;
  *
  * @mixin LeftAndMainPageIconsExtension
  */
-class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionProvider
+class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionProvider, Flushable, MemberCacheFlusher
 {
     /**
      * Unique ID for page icons CSS block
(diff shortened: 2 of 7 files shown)


unclecheese:
Could use some unit tests once overall approach is approved.

dhensby:
This needs to go against `4` branch due to new APIs

code/Controllers/CMSMain.php
```diff
@@ -379,6 +388,22 @@ public function getTabIdentifier()
         return 'edit';
     }
 
+    /**
+     * @param CacheInterface $cache
+     */
+    public function setHintsCache(CacheInterface $cache)
```
tractorcow: The setter needs a getter.

Also please `return $this;` to make it fluent.

code/Controllers/CMSMain.php
```diff
@@ -160,6 +160,15 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         'SiteTreeAsUL' => 'HTMLFragment',
     );
 
+    private static $dependencies = [
```
tractorcow: Can you make the case of the property match the public getter / setter.

```php
'HintsCache'
```

code/Controllers/CMSMain.php
```diff
@@ -974,19 +999,25 @@ public function Breadcrumbs($unlinked = false)
...
 
-        // Generate basic cache key. Too complex to encompass all variations
-        $cache = Injector::inst()->get(CacheInterface::class . '.CMSMain_SiteTreeHints');
-        $cacheKey = md5(implode('_', array(Security::getCurrentUser()->ID, implode(',', $cacheCanCreate), implode(',', $classes))));
-        if ($this->getRequest()->getVar('flush')) {
-            $cache->clear();
-        }
-        $json = $cache->get($cacheKey);
+        // Cache the JSON string representing allowed children
+        $hintsCacheKey = md5(implode('_', [$memberID, implode(',', $cacheCanCreate), $classList]));
```
tractorcow: We don't really need two cache keys here; Can we consolidate this together?

code/Controllers/CMSMain.php
```diff
@@ -974,19 +999,25 @@ public function Breadcrumbs($unlinked = false)
...
-
-        $cacheCanCreate = array();
-        foreach ($classes as $class) {
-            $cacheCanCreate[$class] = singleton($class)->canCreate();
+        $memberID = Security::getCurrentUser()->ID;
+        $cache = $this->hintsCache;
+        $classList = implode(',', $classes);
+
+        // Cache the results of canCreate()
+        $permissionsCacheKey = md5(implode('_', [$memberID, $classList]));
```
tractorcow: $classlist isn't a variable, it's a fact (fixed) for each project. I know it was there earlier but it's probably redundant, now that we have `flush` to clear (should be done when adding new classes).

code/Model/SiteTree.php
```diff
@@ -2940,6 +2970,15 @@ public static function reset()
         }
     }
 
+    /**
+     * Clear the creatableChildren cache on flush
+     */
+    public static function flush()
+    {
+        Injector::inst()->get(CacheInterface::class . '.SiteTree_CreatableChildren')
```
tractorcow: This won't respect `$dependencies` config; Could we use `static::singleton()->getCreatableChildrenCache()` instead?

code/Model/SiteTree.php
```diff
@@ -904,6 +917,14 @@ public function getParent()
         return null;
     }
 
+    /**
+     * @param CacheInterface $cache
+     */
+    public function setCreatableChildrenCache(CacheInterface $cache)
```
tractorcow: Need a getter for this setter.

code/Model/SiteTree.php
```diff
@@ -2512,6 +2533,26 @@ public function allowedChildren()
...
+            $children = array();
+            $candidates = static::page_type_classes();
+            foreach ($candidates as $childClass) {
+                $child = singleton($childClass);
+                if ($child->canCreate(null, array('Parent' => $this))) {
+                    $children[$childClass] = $child->i18n_singular_name();
+                }
+            }
+            $cache->set($cacheKey, $children);
+        }
```
tractorcow: Is this method meant to return something?

code/Controllers/CMSMain.php
```diff
@@ -974,19 +999,25 @@ public function Breadcrumbs($unlinked = false)
     public function SiteTreeHints()
     {
         $classes = SiteTree::page_type_classes();
-
-        $cacheCanCreate = array();
-        foreach ($classes as $class) {
-            $cacheCanCreate[$class] = singleton($class)->canCreate();
+        $memberID = Security::getCurrentUser()->ID;
+        $cache = $this->hintsCache;
```
tractorcow: Once we add a getter we should use that instead of direct properties.

code/Controllers/CMSMain.php
```diff
@@ -974,19 +999,25 @@ public function Breadcrumbs($unlinked = false)
     public function SiteTreeHints()
     {
         $classes = SiteTree::page_type_classes();
-
-        $cacheCanCreate = array();
-        foreach ($classes as $class) {
-            $cacheCanCreate[$class] = singleton($class)->canCreate();
+        $memberID = Security::getCurrentUser()->ID;
```
tractorcow: Assumes getCurrentUser() isn't null.

code/Model/SiteTree.php
```diff
@@ -2512,6 +2533,26 @@ public function allowedChildren()
         return $allowedChildren;
     }
 
+    public function creatableChildren()
```
tractorcow: PHPDoc on return type and purpose please. :)

Could this be protected?

tractorcow:
Good job overall. :)

Are you able to get some kind of performance metric on SiteTreeHints() and getTreeTitle() to see how much faster it is?

unclecheese:
Thanks for those. Good points.

One glaring omission is that it's missing a cache-busting mechanism on Member/Group write. Here's the options, as I see it:

* We could make `CMSMain` and `SiteTree` implementors of `CacheFlusher` (`flushCache($ids = [])` ), but that would put the `flushCache($ids)` method in conflict with the native `DataObject::flushCache()` method. Possible solution: change `CacheFlusher` to define `flushBackendCache($ids)` or something along those lines.

* We create a new layer of abstraction and hand off the entire implementation to another service, which would scale better:

```
class CreateableChildrenCache implements CacheFlusher
{
  __construct(CacheInterface $cache)
  
  set($memberID, $siteTreeClass, $pageID)

  flushCache($memberIDs = []))  
}
```

tractorcow:
Pretty damn awesome. :)

I had to re-run `phpcbf` over it again to fix linting issues, and added a few phpdoc fixes, plus some minor code cleanup. :)

It's all in one commit if you want to review @unclecheese.

Otherwise, can be merged on green.

_config/permissions.yml
```diff
@@ -6,7 +6,6 @@ SilverStripe\Core\Injector\Injector:
     class: SilverStripe\Security\InheritedPermissions
     constructor:
       BaseClass: SilverStripe\CMS\Model\SiteTree
-      CacheService: '%$Psr\SimpleCache\CacheInterface.InheritedPermissions'
```
tractorcow: Why was this removed here?

tractorcow:
I've squashed a bunch of work, and fixed some caching and service bugs.

code/Model/SiteTree.php
```diff
@@ -1508,6 +1541,24 @@ public function flushCache($persistent = true)
...
 
+    /**
+     * Flushes the member specific cache for creatable children
+     * @param null $memberIDs
+     */
+    public function flushMemberCache($memberIDs = null)
+    {
+        $cache = static::singleton()->getCreatableChildrenCache();
+
+        if ($memberIDs) {
```
tractorcow: I think you had this condition inverted

Issue #2012, opened 2017-11-02
Remember "list" view preference in "pages" section


chillu:
# Overview

The pages list view is a very useful way to navigate websites, particularly when they have a large amount of pages (both with a shallow or deep IA). These lists can be filtered and sorted in ways that the tree can't. 

# Acceptance Criteria

 * As a content author, when choosing the list view in the "pages" section, my preference gets remembers the next time I'm navigating to the section
 * When navigating back from an edit page view, the preference is remembered
 * When viewing the tree, the view preference is unset again
 * When loading the CMS, I can interact with pages as quickly as before (meaning the list view gets lazy loaded before tree view)
 * The preference is remembered per device

# Notes

 * Excludes remembering sort/filter criteria
 * This shouldn't be implemented through cookies because it's adding data to every request, incl. assets. We should do this via local storage.
 * This is one piece of the puzzle for decreasing our reliance on a full-featured tree (after replacing TreeDropdownField with a list-based approach). The tree is really hard to maintain, and the jstree library is outdated and too complex for us to keep up with.

# Related

 * [Page edit mode - list/tree toggle](https://github.com/silverstripe/silverstripe-cms/issues/1664)

## Pull requests

- [ ] https://github.com/silverstripe/silverstripe-cms/pull/2049
- [x] https://github.com/silverstripe/silverstripe-admin/pull/375

tractorcow:
Approved and tested; Can be merged on green. Will check back later. :)

PR #2049, opened 2017-12-06
Remember pages view type

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-cms/issues/2012


client/src/legacy/CMSMain.js
```diff
@@ -24,20 +24,49 @@ $.entwine('ss', function ($) {
 		}
 	});
 
+  // Page edit mode
+  $('.cms-edit-form').entwine({
+      onadd: function() {
+        // Always use treeview when in page edit mode
+        localStorage.setItem('pages-view-type', 'treeview');
```
flamerohr: perhaps it'd be worth using a `ss` prefix e.g. `ss.pages-view-type`
In retrospect, I should've done this for the campaign localStorage work I did earlier as well.

client/src/legacy/CMSMain.js
```diff
@@ -24,20 +24,49 @@ $.entwine('ss', function ($) {
...
+
+  /**
+   * Override super's `onadd` to modify url base the previously select view
+   * type.
+   *
+   * See $('.cms .cms-panel-link') in LeftAndMain.js
+   */
+  $('.cms-panel-deferred.cms-content-view').entwine({
+    onadd: function() {
+      const viewType = localStorage.getItem('pages-view-type');
```
flamerohr: I think it'd be a good item to have a fallback, in case there isn't an entry in localStorage

client/src/legacy/CMSMain.js
```diff
@@ -24,20 +24,49 @@ $.entwine('ss', function ($) {
...
-  $('.cms .cms-panel-link.page-view-link').entwine({
+  $('.cms .page-view-link').entwine({
     onclick: function(e){
-      // Toggle 'active' flag
-      this.siblings().removeClass('active');
-      this.addClass('active');
+      e.preventDefault();
+
+      const viewType = $(this).data('view');
+      const $contentView = this.parents('.cms-content-view');
```
flamerohr: use `.closest();` rather than `.parents()`

client/src/legacy/CMSMain.js
```diff
@@ -24,20 +24,49 @@ $.entwine('ss', function ($) {
...
+      var url = $contentView.data(`url-${viewType}`);
+      const isContentViewInSidebar = $contentView.parents('.cms-content-tools').length !== 0;
```
flamerohr: `.closest()` again :)

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_PageList.ss
```diff
@@ -1,7 +1,9 @@
...
-    </div>
-<% else %>
-    <% include SilverStripe\\CMS\\Controllers\\CMSMain_TreeView_Deferred %>
-<% end_if %>
+<div
+    class="panel panel--padded panel--scrollable flexbox-area-grow fill-height flexbox-display cms-content-view cms-tree-view-sidebar cms-panel-deferred"
+    data-url="$LinkTreeViewDeferred"
+    data-url-treeview="$LinkTreeViewDeferred"
+    data-url-listview="$LinkListViewDeferred"
+    data-url-listviewchildren="$LinkListViewChildren(0)"
```
flamerohr: Could we have `0` as the default for `LinkListViewChildren`?
Or a new method like `LinkListViewRoot()` which returns `LinkListViewChildren(0)`?

This one is not a blocker, but I'm not a fan of params in templates unless unavoidable :)

phalkunz:
@flamerohr Thanks a lot for your feedback. I've fixed them. However, I still have failed behat tests to fix.

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_PageList_Sidebar.ss
```diff
@@ -1 +1 @@
-<% include SilverStripe\\CMS\\Controllers\\CMSMain_TreeView_Deferred %>
+<% include SilverStripe\\CMS\\Controllers\\CMSMain_PageList %>
```
tractorcow: Are we not deferring the tree view anymore?

tractorcow: I'd removed deferred intentionally earlier, but was requested to put it back because it increased the overall "until editable" wait for page edit forms.

phalkunz: @tractorcow You are still deferring tree view and as well as list view. We load either tree or list view dynamically via JS using the urls provided in `CMSMain_PageList.ss`.

phalkunz:
@flamerohr I've fixed the tests so it's ready for another review.

tractorcow:
Feedback from testing:

If I am in tree view editing a page, the "list view" icon has the url '/admin/pages/listview'. Clicking it takes me to another page `/admin/pages/listviewchildren/0`. Not sure why it goes to a different url now.

However, if I click "tree view" again the url stays `/admin/pages/listviewchildren/0` when it's not the list view.

I think if you want to control the list / tree via local storage, you need to take the view type out of the url. Thus:

 - admin/pages -> shows the view last remembered by localstorage
 - if in admin/pages in tree view, clicking the list-view button keeps the URL `admin/pages`
  
I would recommend dropping the deferred `listview` action, since deferring this particular view doesn't provide the same value that deferring the `treeview` action does; I.e. letting you edit the page while the tree view is loading. It just makes the overall loading time longer.

For listview for a particular level, you could go back to querystring args `?ParentID=3` instead of trying to use a particular url action, i.e. `/admin/pages/listviewchildren/3`. That way whatever current view can choose to either ignore or respect the querystring arg.

tractorcow:
ok, had a discussion with @phalkunz. Switching to non-pjax for some requests is hard (since there's a lack of context sharing between back / front end), so we need more pjax for all requests.

I would still like to fix the URLs so they don't mis-represent the content they are rendering.

phalkunz:
@tractorcow I've pushed fixes for URL segment

tractorcow:
Nice work @phalkunz I'll probably test it out tomorrow early. :) I appreciate your getting into the difficult feedback and sorting this out.

phalkunz:
@tractorcow Actually, thanks for your effort in putting long and descriptive feedbacks :)

tractorcow:
Can be merged on green.

tractorcow:
@phalkunz sorry I found the search behaviour was incorrect, and "clear search" wasn't working. I couldn't figure out your logic, so it's just a hard page refresh for now, sorry.

Issue #2055, opened 2017-12-13
BUG Breadcrumbs in add-page section missing "pages" top level


tractorcow:
![image](https://user-images.githubusercontent.com/936064/33919160-fef1a354-e01b-11e7-8b9f-5f8618685339.png)

It looks like this should be:

```
Pages /
Add page
```

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-cms/pull/2056

tractorcow:
Changes requested on that PR.

PR #2056, opened 2017-12-14
Add root breadcrumb title to add page form

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-cms/issues/2055

### DIFF
--- a/templates/SilverStripe/CMS/Controllers/Includes/CMSPageAddController_Content.ss
+++ b/templates/SilverStripe/CMS/Controllers/Includes/CMSPageAddController_Content.ss
@@ -4,7 +4,10 @@
 			<div class="toolbar toolbar--north">
 				<div class="toolbar__navigation">
 					<ol class="breadcrumb">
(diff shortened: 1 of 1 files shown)


lang/en.yml
```diff
@@ -72,6 +72,7 @@ en:
     ParentMode_child: 'Under another page'
     ParentMode_top: 'Top level'
     Title: 'Add page'
+    ROOT_BREADCRUMB_TITLE: 'Pages'
```
tractorcow: Duplicate of `SilverStripe\CMS\Controllers\CMSPagesController.MENUTITLE`. Please use the existing key instead.

flamerohr: Update made

tractorcow:
It's not clickable, but it still is an improvement. :)

tractorcow:
Can be merged on green.

dhensby:
linting issues

tractorcow:
Fixed linting issues on 4.0 branch and rebased this on top.

PR #2070, opened 2018-01-10
Update page history javascript to fix big performance issue

emteknetnz:
On pages with a large amount of versions was getting extremely bad performance when clicking on the page history tab.  After the page loaded, javascript was causing the page to be unresponsive for a very long time, up to 120 seconds in chrome.

The issue boiled down to element.compareDocumentPosition being erroneously called a huge number of times by jQuery, like it was getting totally confused by itself

Have updated to use regular JS instead of jQuery for a .removeClass call, and make another jQuery selector more explicit

### DIFF
--- a/javascript/CMSPageHistoryController.js
+++ b/javascript/CMSPageHistoryController.js
@@ -156,10 +156,16 @@
 			 * Function: _unselect()
 			 *
 			 * Unselects the row from the form selection.
+			 *
+			 * Using regular js to update the class rather than this.removeClass('active')
+			 * because the latter causes the browser to continuously call
+			 * element.compareDocumentPosition, causing the browser to hang for long
+			 * periods of time, especially on pages with lots of versions (e.g. 100+)
 			 */
 			_unselect: function() {
-				this.removeClass('active');
-				this.find(":input[type=checkbox]").attr("checked", false);
+				var tr = this.get(0);
+				tr.className = $.trim(tr.className.replace('active', ''));
+				this.find(":input[type=checkbox][checked]").attr("checked", false);
 			},
 			
 			/**
--- a/javascript/CMSPageHistoryController.js
+++ b/javascript/CMSPageHistoryController.js
@@ -164,7 +164,7 @@
 			 */
 			_unselect: function() {
 				var tr = this.get(0);
-				tr.className = $.trim(tr.className.replace('active', ''));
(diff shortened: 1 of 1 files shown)


javascript/CMSPageHistoryController.js
```diff
@@ -156,10 +156,16 @@
...
+			 * Using regular js to update the class rather than this.removeClass('active')
+			 * because the latter causes the browser to continuously call
+			 * element.compareDocumentPosition, causing the browser to hang for long
+			 * periods of time, especially on pages with lots of versions (e.g. 100+)
 			 */
 			_unselect: function() {
-				this.removeClass('active');
-				this.find(":input[type=checkbox]").attr("checked", false);
+				var tr = this.get(0);
+				tr.className = $.trim(tr.className.replace('active', ''));
```
dhensby: I think this could result in `inactive` becoming `in`, correct? maybe regex with a word boundary assertion would be better?

emteknetnz:
@dhensby have updated to use regex word boundaries.

Has been tested in IE8,9,10,11, Latest Edge, FireFox, Chrome and Safari

javascript/CMSPageHistoryController.js
```diff
@@ -156,10 +156,16 @@
...
+			 * Using regular js to update the class rather than this.removeClass('active')
+			 * because the latter causes the browser to continuously call
+			 * element.compareDocumentPosition, causing the browser to hang for long
+			 * periods of time, especially on pages with lots of versions (e.g. 100+)
 			 */
 			_unselect: function() {
-				this.removeClass('active');
-				this.find(":input[type=checkbox]").attr("checked", false);
+				var tr = this.get(0);
+				tr.className = $.trim(tr.className.replace(/\bactive\b/, ''));
```
jacobbuck: `Element.classList` support is pretty common these days, so you could use it like so:

```js
this.get(0).classList.remove('active');
```

dhensby: Only IE11 according to https://caniuse.com/#feat=classlist

jacobbuck: IE 10 and 11 😄

dhensby: Ah, yes. So not quite enough. Thanks for the suggestion anyway. We could use that in 4.x instead.

dhensby: merged up into 4.x using classList

PR #2074, opened 2018-01-17
FIX Add bootstrap styles to url segment field

sachajudd:
I have added styles in place for now before silverstripe-admin is using the lastest bootstap version which we could then implement .input-group-append classes.

https://github.com/silverstripe/silverstripe-admin/issues/395

Before: 
image

After:
image


dhensby:
Yes yes yes! Looks great

robbieaverill:
@sachajudd looks like you might need to rebase your pull request and rebuild it - see https://travis-ci.org/silverstripe/silverstripe-cms/jobs/329734719#L920

PR #2077, opened 2018-01-24
ENHANCEMENT Ensure all base folders are excluded

### LINKED ISSUE #2059: Add reserved keywords
Such as `assets`, `vendor` and `resources`, these paths are now pretty much hardcoded to be blocked or redirected/handled elsewhere.
Top-level pages with these keywords should give a validation error or warning to keep users informed of what happened to their page.

# PRs:

 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/2077

tractorcow:
Fixes #2059

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -1614,43 +1614,19 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      */
     public function validURLSegment()
     {
-        $excludes = Director::config()->get('rules');
-        $excludes = array_keys($excludes);
-        $disallowedSegments = array_map(function ($key) {
-            $route = explode('/', $key);
-            if (!empty($route) && strpos($route[0], '$') === false) {
-                return $route[0];
-            }
-            return;
-        }, $excludes);
-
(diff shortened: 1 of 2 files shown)


code/Model/SiteTree.php
```diff
@@ -1614,36 +1614,27 @@ public function validate()
...
-
-        if (!$this->ParentID && in_array($this->URLSegment, $disallowedSegments)) {
-            // Default to '-2', onBeforeWrite takes care of further possible clashes
-            return false;
-        }
-
-
-        if (self::config()->nested_urls && $parent = $this->Parent()) {
+            // Check if urlsegment matches a class name
+            if (class_exists($this->URLSegment) && is_subclass_of($this->URLSegment, RequestHandler::class)) {
```
flamerohr: I thought Controller urlsegment access was removed during alpha?

tractorcow: Yes but this seems a prudent check in any case.

I was worried if I removed it, it would be a risk.

tractorcow:
I think I can remove it, now. I'll do it after lunch. ;0

tractorcow:
Updated @flamerohr

PR #2085, opened 2018-02-02
NEW Use Bootstrap alerts instead of legacy message classes for install.php warning

robbieaverill:
Related issue: https://github.com/silverstripe/silverstripe-admin/issues/201

cc @sachajudd

Before:

image

After:

image


robbieaverill:
Related pull request (removes the original classes we use at the moment) https://github.com/silverstripe/silverstripe-admin/pull/412

PR #2087, opened 2018-02-05
backport #1774 (archive children warning) to 3.5

andrewandante:
Update to #1698 with backport from #1774


code/controllers/CMSMain.php
```diff
@@ -710,6 +713,20 @@ public function getEditForm($id = null, $fields = null) {
 		}
 	}
 
+	/**
+	 * @param SiteTree $record
+	 */
+	protected function getArchiveWarningMessage($record)
+	{
+		// Get all page descendants
+		$record->collateDescendants(true, $descendants);
```
flamerohr: @tractorcow mentioned if we could use `numChildren` instead, do you think that's still a reasonable change to have?

robbieaverill:
We're in limited support (security and critical bugfixes only) for SS3 now, so this is unlikely to make it in. Thanks for the PR though =)

PR #2089, opened 2018-02-07
Fix warning appearing when button is disabled for rollback

flamerohr:


tractorcow:
fix at https://github.com/silverstripe/silverstripe-cms/pull/2145

PR #2090, opened 2018-02-07
NEW Convert page history notice to use Bootstrap 4 info alert

robbieaverill:
Before:

image

After:

image

### DIFF
--- a/code/Controllers/CMSPageHistoryController.php
+++ b/code/Controllers/CMSPageHistoryController.php
@@ -220,10 +220,10 @@ class CMSPageHistoryController extends CMSMain
                 $id
             );
 
-            $view = _t('SilverStripe\\CMS\\Controllers\\CMSPageHistoryController.VIEW', "view");
+            $view = _t(__CLASS__ . '.VIEW', "view");
 
             $message = _t(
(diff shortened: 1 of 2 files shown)


code/Controllers/CMSPageHistoryController.php
```diff
@@ -247,9 +247,9 @@ public function getEditForm($id = null, $fields = null, $versionID = null, $comp
         /** @var Tab $mainTab */
         $mainTab = $fields->fieldByName('Root.Main');
         $mainTab->unshift(
-            new LiteralField('CurrentlyViewingMessage', ArrayData::create(array(
+            LiteralField::create('CurrentlyViewingMessage', ArrayData::create(array(
                 'Content' => DBField::create_field('HTMLFragment', $message),
-                'Classes' => 'notice'
+                'Classes' => 'alert alert-info'
```
kinglozzer: Should we just move these to the template? I don’t see any advantage to keeping the classes in PHP

robbieaverill: The only drawback I see is that it can’t be reused for other alert types then, however this is the only place it’s used currently

kinglozzer: Truedat, let’s leave it then

PR #2100, opened 2018-02-19
NEW Add extension point to CMSMain::Breadcrumbs

robbieaverill:
Part of silverstripe/silverstripe-admin#439

Allows, for example, user code to push a badge in the "Extra" attribute for the last breadcrumbs link.

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -992,6 +992,9 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
                 'Title' => CMSPagesController::menu_title(),
                 'Link' => ($unlinked) ? false : $this->LinkPages()
             )));
+
+            $this->extend('updateBreadcrumbs', $items);
+
             return $items;
         }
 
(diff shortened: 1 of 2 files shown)


code/Controllers/CMSMain.php
```diff
@@ -1009,6 +1009,8 @@ public function Breadcrumbs($unlinked = false)
             )));
         }
 
+        $this->extend('updateBreadcrumbs', $items);
```
tractorcow: what about CMSPagesController::Breadcrumbs()? It will construct parent breadcrumbs + extensions, then adds the current page to the end. You won't be able to decorate that.

You could use `$this->beforeExtending('updateBreadcrumbs', function () { });` in CMSPagesController.

robbieaverill: Strange that I haven't seen this happening while testing, but you're right - maybe I should shift this extension point into CMSPagesController instead

robbieaverill: image

Is this the breadcrumbs you're talking about from CMSPagesController?

tractorcow: it's in the listview when ParentID is set

robbieaverill: Ok thanks - I've put the CMSPagesController logic into a beforeExtending callback and pushed

PR #2102, opened 2018-02-19
BUG Fix behaviour towards versioned but unstagable records

tractorcow:
Parent issue: https://github.com/silverstripe/silverstripe-versioned/issues/115


dhensby:
@tractorcow these won't go green despite the other PRs being merged :(

tractorcow:
hm that test is broken even on 4.0 branch now.

tractorcow:
Ah, it's a regression from an earlier versioned change. :D don't worry the test was bad because it was relying on incorrect rollback behaviour.

tractorcow:
Green @dhensby

PR #2115, opened 2018-03-01
add controller_name config var to SiteTree for easier override

andrewandante:
Currently you can do this by overriding the `getControllerName()` method but this feels more idiomatic. Means that you can more easily do things like put your controllers in a `MySite\Control` namespace and have your pages in `Mysite\Model`. Additionally saves having to do the while loop if you don't have to - could be an aim to have this as required by 5.0 to reduce unnecessary cycles (and ✨ magic ✨ ).

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -185,6 +185,15 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      */
     private static $hide_ancestor = null;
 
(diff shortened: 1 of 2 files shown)


dhensby:
see #6549

chillu:
Awesome work Andrew! I'm asking to get this documented in a wider context as well: https://github.com/silverstripe/silverstripe-framework/issues/8279

PR #2121, opened 2018-03-12
NEW Add page GraphQL query HOC for history viewer component

robbieaverill:
Relies on silverstripe/versioned-admin being installed on SS 4.2 or above.

Parent issue: https://github.com/silverstripe/silverstripe-versioned/issues/37


client/src/state/history/readOnePageQuery.js
```diff
@@ -0,0 +1,95 @@
...
+    const versions = readOnePage || null;
+
+    const errors = error && error.graphQLErrors &&
+      error.graphQLErrors.map((graphQLError) => graphQLError.message);
+
+    return {
+      loading: networkLoading || !versions,
+      versions,
+      graphQLErrors: errors,
+      actions: Object.assign({}, actions, {
```
unclecheese: We're trying to deprecate this in favour of object spreads.

```js
actions: {
  ...actions,
  versions: {
    ...actions.versions
    goToPage(page)
  }
}
```

_config/graphql.yml
```diff
@@ -0,0 +1,15 @@
+---
+Name: cmsgraphql
+---
+SilverStripe\GraphQL\Controller:
+  schema:
+    scaffolding:
+      types:
+        SilverStripe\CMS\Model\SiteTree:
+          fields: '*'
```
unclecheese: Curious why this is wildcarded. Is it to support the "compare" view, where you'll show all the fields?

Any queries that depend on an all-inclusive field list will need to be generated dynamically, which is something we're still working on. Is that the idea?

client/src/state/history/readOnePageQuery.js
```diff
@@ -0,0 +1,95 @@
...
+          }
+        }),
+      }),
+    };
+  },
+};
+
+export { query, config };
+
+export default graphql(query, config);
```
unclecheese: Really nice job decoupling all of this. `HistoryViewer`, `withPagesHIstoryViewer`, and `readOnePageQuery` can all be consumed independently of each other.

_config/graphql.yml
robbieaverill: We only really need the versioning information at the moment, so wildcarding the fields is probably not required. I'll adjust.

client/src/state/history/readOnePageQuery.js
robbieaverill: Thanks :) I've updated the field list and spread operator syntax and rebased

client/src/state/history/readOnePageQuery.js
```diff
@@ -0,0 +1,97 @@
+import { graphql } from 'react-apollo';
+import gql from 'graphql-tag';
+
+// GraphQL query for retrieving the version history of a specific page. The
+// results of the query must be set to the "versions" prop on the component
+// that this HOC is applied to (see PageHistoryViewer.js) for binding
+// implementation.
+const query = gql`
+query readOnePage ($page_id: ID!, $limit: Int!, $offset: Int!) {
```
unclecheese: Minor point here, but these query names are arbitrary. They don't necessarily have to be consistent with their root query field name. For debugging purposes, it would probably help to give this a more distinct name like `query ReadHistoryViewerPage ($page_id: ID! ...)`

robbieaverill: Cool, will update

robbieaverill:
Todo:

 - [ ] Rename query to be less generic

robbieaverill:
Feedback addressed, self merging with @unclecheese's approval

client/src/components/PageHistoryViewer/PageHistoryViewer.js
```diff
@@ -0,0 +1,12 @@
+import { compose } from 'redux';
+import readOnePageQuery from 'state/history/readOnePageQuery';
+
+/**
+ * Transformer function which binds the readOnePageQuery GraphQL query to the HistoryViewer
+ * component.
+ */
+const withPagesHistoryViewer = (HistoryViewer) => compose(
+  readOnePageQuery
+)(HistoryViewer);
```
flamerohr: this reads really weird to me, isn't this the equivalent of just `const withPagesHistoryViewer = readOnePageQuery;`?

Don't get me wrong, great work :D I'm just curious why all the wrapping

unclecheese: Yup, same thing as with the [silverstripe/elemental PR](https://github.com/dnadesign/silverstripe-elemental/pull/207), the whole HOC can be eliminated.

```js
      updater.component(
        'HistoryViewer.pages-controller-cms-content',
        readOnePageQuery,
        'PageHistoryViewer'
      );
```

Thanks for catching this. Didn't realise there were two of them.

Issue #2125, opened 2018-03-15
SearchForm results() function exposed as controller action


rotassator:
Browsing to a page ending in `~/results` causes the following error:
```
[User Error] Uncaught ArgumentCountError: Too few arguments to function ContentControllerSearchExtension::results(), 1 passed in ~/framework/core/Object.php on line 749 and exactly 3 expected
```

This occurs when using `FulltextSearchable::enable()` (or related).

Using SilverStripe 3.6.5.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-cms/pull/2126

PR #2126, opened 2018-03-15
FIX: Remove SearchForm results() function from allowed_actions

rotassator:
Relates to https://github.com/silverstripe/silverstripe-cms/issues/2125


rotassator:
Also applies to SilverStripe 4. 
https://github.com/silverstripe/silverstripe-cms/blob/a02410b983678d1d38cfb9daabaa7d259cdea28c/code/Search/ContentControllerSearchExtension.php#L20

dhensby:
I've backported this to 3.5

dhensby:
I've also merged up into 4

Issue #1578, opened 2016-08-09
Better indicator of when you're viewing the draft site


JorisDebonnet:
WAS: "Unpublished pages still show when logged in"

This may be by design because it's been like this for as far as I can remember (3.1? - 3.4), but I think it's annoying, so I thought I'd raise it :)

If I go into the CMS and say "let's unpublish this page/post for now", it will still show up in the menu/list on the website because I'm still logged in. So apparently I see all "draft" pages as long as I'm logged in. That's quite confusing because I am not able to check what the site will look like to visitors unless I view it in a different browser, or log out each time. I would have expected unpublished pages to only show up when `?stage=Stage` is appended to the URL, or something like that. In fact, I see no difference between using `?stage=Stage` and not using it, at least as long as I am logged in (as Administrator). 

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7399
 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/45
 - [x] https://github.com/silverstripe/silverstripe-admin/pull/467
 - [x] https://github.com/silverstripe/silverstripe-cms/pull/2129

dhensby:
This issue has been raised repeatedly in many forms. In the long run I think it'd be better to use the `?stage=Stage` get var as the indicator of whether we're viewing the stage site (rather than a session as we currently do).

The user needs to switch their viewing mode to live by using the stage switcher in the bottom right of the CMS. Alternatively, using the https://github.com/jonom/silverstripe-betternavigator module might help for the short term?

JorisDebonnet:
That does look like an interesting module! 
I don't understand what you mean by selecting the view mode though: all I see in the bottom right is the choice between the modes Edit, Preview and Split. Those all sound Draft to me :)

dhensby:
@JorisDebonnet I'm talking about this:

![](https://dl.dropboxusercontent.com/u/3429338/Clippings/2016-08/2016-08-15_11-36-42-Qv5075exLB.png)

JorisDebonnet:
Ah yes, right, in the preview section in the CMS. I actually never use that, prefer using the pure edit more (I think the Split view constricts the view too much, and it also causes issues with pages that redirect in the page code). Instead, I make it look as close to the end result as possible right inside the editor by applying classes to TinyMCE's body element. 

[...]

PR #2129, opened 2018-03-21
Fix versioned regressions

tractorcow:
Requires https://github.com/silverstripe/silverstripe-versioned/pull/45

Parent story https://github.com/silverstripe/silverstripe-cms/issues/1578

Still some bugs to fix! Not done yet.


tractorcow:
Ah yes good spotting. Not using useDraftSite() is a good idea even if we don't deprecate it. :D

PR #2132, opened 2018-03-22
FIX Correct SilverStripeNavigator correctly in templates

### LINKED ISSUE #2130: $SilverStripeNavigator returning escaped HTML
Using a fairly bare bones SilverStripe 4.1.0 install. Added $SilverStripeNavigator to Page.ss and its returning escaped HTML

dhensby:
Fix #2130

`SilverStripeNavigator` needs to be cast as html otherwise it renders as escaped html


tractorcow:
We could have also just added a value for `private static $casting = ["SilverStripeNavigator" => "HTMLFragment"];`

dhensby:
Good point @tractorcow this hasn't been released yet so we can always revert...

PR #2138, opened 2018-04-03
MINOR Refactor implementation of history viewer to use only Injector.transform

robbieaverill:
Similarly to the feedback left on https://github.com/dnadesign/silverstripe-elemental/pull/207, all that is really needed at this stage for the CMS module is an `Injector.transform` call applying the `readOnePageQuery` HoC to the history viewer component.

This PR removes the unneeded PageHistoryViewer export from the components dir.

cc @unclecheese


chillu:
@robbieaverill Needs a rebase

robbieaverill:
My bad, I did this in https://github.com/silverstripe/silverstripe-cms/commit/e3237f9638d72d49e1e192738e00eacac2308c69

PR #2139, opened 2018-04-09
Migrate legacy SiteTree link tracking for all subsites

mikenz:
Currently this task only migrates the link tracking for the subsite it is called on when run from a browser.

### DIFF
--- a/code/Tasks/MigrateSiteTreeLinkingTask.php
+++ b/code/Tasks/MigrateSiteTreeLinkingTask.php
@@ -6,6 +6,7 @@ use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Dev\BuildTask;
 use SilverStripe\Dev\Debug;
 use SilverStripe\ORM\DB;
+use SilverStripe\Subsites\Model\Subsite;
 use SilverStripe\Versioned\Versioned;
 
 /**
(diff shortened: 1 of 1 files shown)


mikenz:
I couldn't think of an immediately obvious alternative solution.

mikenz:
Oh, maybe extend the task in Subsites and use the Injector to substitute it?

robbieaverill:
That could work!

I was thinking of something like:

```php
public function run()
{
    $this->extend('beforeRun');
    // run stuff
    $this->extend('afterRun');
}
```

And add a subsites extension like:

```php
class MigrateSiteTreeTaskExtension extends Extension
{
    public function beforeRun()
    {
        $this->origDisableSubsiteFilter = Subsite::$disable_subsite_filter;
        Subsite::disable_subsite_filter(true);
    }
    // ...
}
```

Noting that extensions are singletons by default now.

TBH I'd also expect that the BuildTask API will change in SS5 - it might allow for middlewares that could make this kind of thing easier

tractorcow:
Thanks for the report (it's completely valid) but please raise module-specific issues against the repo in question.

It's better that subsites implemented an alternative migration task than we try to bake too many assumptions into core. Requesting to add extension points is fine if you want to raise a PR, however.

maxime-rainville:
The issue is legit, but the proposed solutions has shortcomings. I've raised an issue against subsite so we don't loose site of it. https://github.com/silverstripe/silverstripe-subsites/issues/407

I'll close this for now, but if you feel like having another go at it, just let us know and I'll reopen it.

robbieaverill:
We have recently had a similar problem with fulltextsearch and file migration tasks. The solution needs to be that the tasks (like this one) have extension points added into them so that modules like subsites and fulltextsearch can turn their ORM filters on and off in some situations.

PR #2140, opened 2018-04-09
adjustment to show Page Type title on SiteTree mouseover

korthjp17:
fixes #2097

PR #2141, opened 2018-04-09
ENHANCEMENT Use i18n template for page tree title

tractorcow:
Allows positional localisation


robbieaverill:
Seeing as #2140 went into 4.0 as a patch change I think this PR is also safe to go into 4.0

tractorcow:
Yeah, that's my reasoning also.

On the other hand, had that change been included in a tagged release, I'd consider that api "set", and I would have made my change in the next minor release branch (4).

Also "enhancement" isn't strictly an "api" change, so it requires less semver consideration than an actual api change. :)

PR #2145, opened 2018-04-16
BUG Fix rollback button from not working

tractorcow:
Fixes regression introduced in #2089 

Required for https://github.com/silverstripe/silverstripe-versioned/issues/94


chillu:
@tractorcow Please don't forget to add a points estimate

PR #2153, opened 2018-04-24
ENHANCEMENT Backport versioned querystring fix

tractorcow:
Parent issue https://github.com/silverstripe/silverstripe-versioned/issues/145

### DIFF
--- a/tests/controller/ContentControllerTest.php
+++ b/tests/controller/ContentControllerTest.php
@@ -9,8 +9,7 @@ class ContentControllerTest extends FunctionalTest {
 
 	protected static $disable_themes = true;
 
-	public function setUp()
-	{
+	public function setUp() {
 		parent::setUp();
 		Config::inst()->update('Director', 'alternate_base_url', '/');
 		$this->useDraftSite(false);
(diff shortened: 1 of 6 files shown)


tractorcow:
Requires https://github.com/silverstripe/silverstripe-framework/pull/8026

tests/controller/ContentControllerTest.php
```diff
@@ -7,16 +7,22 @@ class ContentControllerTest extends FunctionalTest {
 
 	protected static $fixture_file = 'ContentControllerTest.yml';
 
-	protected static $use_draft_site = true;
-
 	protected static $disable_themes = true;
 
+	public function setUp()
+	{
+		parent::setUp();
```
unclecheese: Let's keep block formatting consistent with the rest of the class.

tests/model/RedirectorPageTest.php
```diff
@@ -2,47 +2,59 @@
...
 class RedirectorPageTest extends FunctionalTest {
 	protected static $fixture_file = 'RedirectorPageTest.yml';
-	protected static $use_draft_site = true;
+
 	protected $autoFollowRedirection = false;
-	
+
+	public function setUp()
+	{
+		parent::setUp();
```
unclecheese: Same here. Move to same line.

tractorcow: Ah right, reverse linting. ;)

unclecheese:
Have confirmed that tests are only failing because they depend on the [framework PR](https://github.com/silverstripe/silverstripe-framework/pull/8026). We should merge this ahead of green, as the PRs are interdependent.

PR #2157, opened 2018-05-01
Bugfix: Don't offer to restore to root if can_be_root is false

mikenz:
Remove the Restore button if there's no where viable to restore the page to.


robbieaverill:
I've merged up 4 into the master branch which should fix your PHPUnit tests, the Behat tests are still broken in master on framework and CMS so they'll fail. Would you mind rebasing this PR?

dhensby:
This merges cleanly and tests (apart from behat) pass, so rebase is not needed

PR #2159, opened 2018-05-01
Pulls/4.0/ssnaviagtor casting config

dhensby:
Alternative approach to #2132

A less significant API change for a patch release

### DIFF
--- a/code/Controllers/ContentController.php
+++ b/code/Controllers/ContentController.php
@@ -58,6 +58,10 @@ class ContentController extends Controller
         'LoginForm'
     );
 
+    private static $casting = array(
+        'SilverStripeNavigator' => 'HTMLFragment',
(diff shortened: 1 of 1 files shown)


code/Controllers/ContentController.php
```diff
@@ -58,6 +58,10 @@ class ContentController extends Controller
         'LoginForm'
     );
 
+    private static $casting = array(
```
robbieaverill: Super minor, but maybe we should use `[]` array syntax for SS4?

dhensby: I'm just using the same style as currently in the file - but I don't mind either way

dhensby: I've updated the code style for the file

PR #2161, opened 2018-05-07
Tests assume default config can fail

mikenz:
We've changed default config or unloaded extensions that are causing tests to fail. This fixes those tests, I guess other devs have their own instances of this too.


robbieaverill:
Behat failure is pre-existing

Issue #2166, opened 2018-05-16
SS4.1 - Installation files are not removed after remove via browser


odorisioe:
**Version:** 4.1

**Steps**
1. Install using /install.php on the browser.
2. On "Installation Successful!" page, click on the link _Click here to delete the install files_.
3. Log as an admin user if is required.

**Result**
The browser shows the url /home/deleteinstallfiles with the message _Installation files have been successfully deleted_ but install.php is not removed.

**Issue**
The code below is using BASE_PATH instead of PUBLIC_PATH to match with the new public webroot folder. I will create a pull request as it should only affect new v4 installations.
https://github.com/silverstripe/silverstripe-cms/blob/01b211e28abc1c38c906c16b645630470a33a31c/code/Controllers/ContentController.php#L522-L531

In other hand, I could not find all the files of the list to remove. Does ss4 use all of them?
https://github.com/silverstripe/silverstripe-cms/blob/01b211e28abc1c38c906c16b645630470a33a31c/code/Controllers/ContentController.php#L515-L520

# Pull requests
* [x] #2175

PR #2167, opened 2018-05-16
#2166 Update installfiles path to remove on installation

odorisioe:
Related to ticket #2166


maxime-rainville:
Sorry @odorisioe, I was blind and missed the fact you already had a PR open. I ended up duplicating your work in #2175. I'll pay more attention next time.

Thanks a lot for documenting the issue and researching a fix. It made it real easy to fix.

PR #2174, opened 2018-05-23
FIX Make RedirectorPage::Link compatible with SiteTree::Link

dhensby:
This prevents the following warning:

`Fatal error: Declaration of RedirectorPage::Link() must be compatible with SiteTree::Link($action = NULL) in /vagrant/project/cms/code/model/RedirectorPage.php on line 0` in PHP 7.2


kinglozzer:
`Fatal error`? 😮 I’m sure that’s only supposed to be a warning... I’m a bit confused as to why this has never raised any warnings/strict standards notices before 😕

Anyway, seems like a good change to me

dhensby:
Yeah, I'm not sure - but that's the error I get when running php 7.2 :/

dhensby:
behat failures seem to be pre-existing

PR #2175, opened 2018-05-23
Adjust `ContentController::deleteinstallfiles()` to use PUBLIC_PATH.

maxime-rainville:
`deleteinstallfiles` doesn't currently work because it expect the install files to be in the root.

I've also removes some old SS2.4 install legacy files from the list of files to check and added `install-frameworkmissing.html`

# Parent issue
* #2166


maxime-rainville:
Merging on green as discussed in stand up.

PR #2176, opened 2018-05-29
BUGFIX: Move AnchorSelectorField to admin

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-framework/issues/7996


unclecheese:
Behat test is failing because of dependency on the `silverstripe/admin` PR.

tractorcow:
You're only moving the js not the php backend AnchorSelectorField.php.

tractorcow:
See my comment at https://app.zenhub.com/workspace/o/silverstripe/silverstripe-framework/issues/7996#issuecomment-395295048. I dislike this approach.

tractorcow:
See https://github.com/silverstripe/silverstripe-cms/pull/2182 for alternative solution.

robbieaverill:
The original issue was closed with an alternative solution

PR #2181, opened 2018-06-01
BUG Ensure errorpage is built in live mode

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/8110

Also merge https://github.com/silverstripe/silverstripe-framework/pull/8122

### DIFF
--- a/code/model/ErrorPage.php
+++ b/code/model/ErrorPage.php
@@ -105,15 +105,31 @@ class ErrorPage extends Page {
 				);
 				$pageExists = ($page && $page->exists());
 				$pagePath = self::get_filepath_for_errorcode($code);
-				if(!($pageExists && file_exists($pagePath))) {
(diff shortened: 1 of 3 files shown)


code/model/ErrorPage.php
```diff
@@ -113,7 +113,12 @@ public function requireDefaultRecords() {
 					}
 
 					// Ensure a static error page is created from latest error page content
+					$origReadingMode = static::get_reading_mode();
+					Versioned::reading_stage(Versioned::LIVE);
```
dhensby: This placement won't work, it needs to go before `$page = DataObject::get_one(...)` because the `udapteLink` method looks at the query params on the object which are set at query time...

tractorcow: Oh yes you're right.

tractorcow: Fixed and added tests.

PR #2182, opened 2018-06-07
BUG Fix CMS components failing to register on other CMS sections

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7996

Alternative solution to moving CMS components to admin module (https://github.com/silverstripe/silverstripe-cms/pull/2176)

### DIFF
--- a/_config/scripts.yml
+++ b/_config/scripts.yml
@@ -1,0 +1,9 @@
+---
+Name: cmsscripts
+---
+SilverStripe\Admin\LeftAndMain:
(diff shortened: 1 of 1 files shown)


tractorcow:
@unclecheese LeftAndMain doesn't support that properly; If it's a map it assumes the right hand side is a config array rather than a nullable flag.

tractorcow:
Need to re-open because of branch change.

PR #2193, opened 2018-06-17
URLSegment field styling fixes 

### LINKED ISSUE #2137: URLSegmentField readonly state should look like a text field
image

**SS 4.2.x-dev**

The URL segment field should look like a standard text field in its readonly state.

lukereative:
Fixes #2137, now looks like other readonly fields:
screen shot 2018-06-18 at 11 06 17 am

In conjunction with improved field add-ons in https://github.com/silverstripe/silverstripe-admin/pull/546 I improved the look of the confirm and cancel buttons when editing the URLSegment:
From:
screen shot 2018-06-18 at 11 09 12 am
To:
screen shot 2018-06-18 at 11 06 34 am


maxime-rainville:
When I'm editing the URL segment in Split mode with a browser window under 2000 pixel wide, the URL Segment field is broken over two line.

image

templates/SilverStripe/CMS/Forms/SiteTreeURLSegmentField.ss
```diff
@@ -1,20 +1,31 @@
...
+        <div class="input-group-append">
+            <button role="button" data-icon="cancel" type="button" class="btn btn-outline-secondary btn-sm input-group-append cancel">
+                <%t SilverStripe\CMS\Forms\SiteTreeURLSegmentField.Cancel 'Cancel' %>
+            </button>
+        </div>
+    </div>
+    <% if $HelpText %><p class="form__field-description">$HelpText</p><% end_if %>
 </div>
+<% else %>
+<a class="readonly-url" href="$URL" target="_blank">
```
maxime-rainville: When I click this link I get taken to the live URL. The old form field would take me to the Draft URL.

templates/SilverStripe/CMS/Forms/SiteTreeURLSegmentField.ss
```diff
@@ -1,20 +1,31 @@
+<% if not $IsReadonly %>
```
maxime-rainville: You don't need an if statement here. You could just have `SiteTreeURLSegmentField_Readonly.ss` file to handle the readonly bit.

templates/SilverStripe/CMS/Forms/SiteTreeURLSegmentField.ss
maxime-rainville: The current field will post back a value when submitted in read only mode. This new one will not.

It's doesn't look like it's breaking anything on the CMS Form, but it would be safer to replicate the original behavior.

lukereative:
@maxime-rainville It appears to only happen when the 'Cancel' word is longer in another translation – so it wasn't appearing for me and only appearing because yours is in French. Fixed now.

templates/SilverStripe/CMS/Forms/SiteTreeURLSegmentField.ss
lukereative: I've reverted to the old behaviour of just hiding the Edit action, thus the field is still present.

code/Model/SiteTree.php
```diff
@@ -1926,6 +1926,7 @@ public function getCMSFields()
             $helpText .= _t('SilverStripe\\CMS\\Forms\\SiteTreeURLSegmentField.HelpChars', ' Special characters are automatically converted or removed.');
         }
         $urlsegment->setHelpText($helpText);
+        $urlsegment->setReadonly(true);
```
maxime-rainville: Don't think this is meant to be there.

Issue #2194, opened 2018-06-18
[SS 3.x] Defaults not applied on page type change. Wrong ClassName used when calling DataObject::newClassInstance from CMSMain::save


JoshuaCarter:
https://github.com/silverstripe/silverstripe-cms/blob/7052170e89799051344d45fd8b851bd004e0bd03/code/controllers/CMSMain.php#L976

Discovered as a result of investigating an issue with the ListingPage module here (for context):
https://github.com/nyeholt/silverstripe-listingpage/issues/23

#### Issue:

When changing page type and triggering a `write` from the CMS (saving to draft, etc), the page object's default's aren't applied to the DB.

#### Findings:

Initially was looking into why `populateDefaults`, `onBeforeWrite`, and `onAfterWrite` weren't being called on the `ListingPage` object after changing page type from `Page` to `ListingPage` and saving to draft in the CMS.

Instead I found that these were all being called on `Page` object instead. On looking into why, I found that `DataObject::newClassInstance` was being given the new/original `ClassName` the wrong way around.

So, when changing from `Page` to `ListingPage`, `newClassInstance` is called with a 'new' class of `Page` and an 'original' class of `ListingPage` (even though it's never been a ListingPage at this point). `newClassInstance` creates a new instance with the `ClassName` of 'Page', and then calls `setClassName` and `populateDefaults` on it. Something similar occurs regarding `on{Before/After}Write`, etc.

#### Cause:

Refer to top link (CMSMain#L976). The comments here make it clear that it's assumed that `ClassName` will have already been set to the new value via `$form->saveInto`, and thus can be retrieved from `$record` rather than `$data`. But you can see below, on line #985, that `saveInto` is called after the fact.

This can be juxtaposed with the SS4 equivalent, which solves the problem by effectively retrieving `ClassName` from `$record` and `$data` in the opposite order. See link:

https://github.com/silverstripe/silverstripe-cms/blob/7c3dba283c47a1db06da3571879d49e973045b95/code/Controllers/CMSMain.php#L1676

I presume this can be fixed in a similar manner for SS 3.7

Thanks.

dhensby:
We'll be releasing 3.7.1 very soon, if you want to get a fix into that, would you be able to get a PR to us?

JoshuaCarter:
Done, please find it here:
https://github.com/silverstripe/silverstripe-cms/pull/2195

PR #2195, opened 2018-06-19
[Fix] CMSMain::save - class change & publish perms

JoshuaCarter:
See issue:
https://github.com/silverstripe/silverstripe-cms/issues/2194

- Added missing check for publishing rights.
- Fixed class change as it was treating ClassName attrs in reverse order (going new->to->old, rather than old->to->new).

### DIFF
--- a/code/controllers/CMSMain.php
+++ b/code/controllers/CMSMain.php
@@ -963,7 +963,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 			if(!singleton($this->stat('tree_class'))->canCreate()) return Security::permissionFailure($this);
 			$record = $this->getNewItem($id, false);
 		}
-
+		
 		// TODO Coupling to SiteTree
 		$record->HasBrokenLink = 0;
 		$record->HasBrokenFile = 0;
@@ -972,12 +972,8 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
 		// Update the class instance if necessary
 		if(isset($data['ClassName']) && $data['ClassName'] != $record->ClassName) {
-			$newClassName = $record->ClassName;
-			// The records originally saved attribute was overwritten by $form->saveInto($record) before.
-			// This is necessary for newClassInstance() to work as expected, and trigger change detection
-			// on the ClassName attribute
-			$record->setClassName($data['ClassName']);
-			// Replace $record with a new instance
+			// Replace $record with a new instance of the new class
+			$newClassName = $data['ClassName'];
 			$record = $record->newClassInstance($newClassName);
 		}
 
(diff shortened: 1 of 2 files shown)


code/controllers/CMSMain.php
```diff
@@ -964,6 +964,12 @@ public function save($data, $form) {
 			$record = $this->getNewItem($id, false);
 		}
 
+		// Check publishing permissions
+		$doPublish = !empty($data['publish']);
+		if ($record && $doPublish && !$record->canPublish()) {
```
maxime-rainville: Would you mind removing this permission check? It's probably not going to cause any problem for any one, but this is going to be our last SS3 release and we would rather only do bug fixes and avoid introducing new behavior that doesn't directly fix a bug.

JoshuaCarter: Sure, I've removed the perms check.

JoshuaCarter:
Cheers @maxime-rainville,

I've removed the perms bit. Can I leave the unit test to you? Haven't looked into SS tests yet and not flush with time atm. Thanks.

code/controllers/CMSMain.php
```diff
@@ -972,13 +972,9 @@ public function save($data, $form) {
...
-			$newClassName = $record->ClassName;
-			// The records originally saved attribute was overwritten by $form->saveInto($record) before.
-			// This is necessary for newClassInstance() to work as expected, and trigger change detection
-			// on the ClassName attribute
-			$record->setClassName($data['ClassName']);
-			// Replace $record with a new instance
-			$record = $record->newClassInstance($newClassName);
+            // Replace $record with a new instance of the new class
+			$newClassName = $data['ClassName'];
+            $record = $record->newClassInstance($newClassName);
```
dhensby: consistent tabbing please

JoshuaCarter: Odd, indentation looked fine on my fork (and IDE), guessing there's a repo setting for that I don't know about. Anyway, a few odd commits later, and it should be fine now.

maxime-rainville:
Sure thing.

dhensby:
@JoshuaCarter can you squash your 5 commits into 1 please?

JoshuaCarter:
@dhensby sure, done.

tests/controller/CMSMainTest.php
```diff
@@ -529,20 +529,62 @@ public function testGetEditForm() {
...
+
+	public function testSave()
+	{
+		// Login is required prior to accessing a CMS form.
+		$this->loginWithPermission('ADMIN');
+
+		$page = $this->objFromFixture('Page', 'page1');
+		$page->doPublish();
+
+		$this->get('admin/pages/edit/show/' . $page->ID);
```
dhensby: `$page->CMSEditLink()` would be better than manually constructing the link

dhensby:
@maxime-rainville it looks like your test passes even without the fix :/

maxime-rainville:
@dhensby You sure? My test fails locally when I undo Joshua's change.
image

dhensby:
I'm not certain now! But Travis passed somehow on your test commit...

maxime-rainville:
Is this good to merge?

JoshuaCarter:
I think so ( ͡° ͜ʖ ͡°)

JoshuaCarter:
Thanks @maxime-rainville and @dhensby :)

PR #2206, opened 2018-07-02
MINOR: Show 'Inherit from site access settings' if there is no parent

lukereative:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/4766

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -2100,8 +2100,10 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             }
         }
 
+        $inheritMessage = $this->ParentID !== 0 ? _t(__CLASS__.'.INHERITPARENT', "Inherit from parent page") : _t(__CLASS__.'.INHERITSITECONFIG', "Inherit from site access settings");
+
         $viewersOptionsSource = [
-            InheritedPermissions::INHERIT => _t(__CLASS__.'.INHERIT', "Inherit from parent page"),
+            InheritedPermissions::INHERIT => $inheritMessage,
             InheritedPermissions::ANYONE => _t(__CLASS__.'.ACCESSANYONE', "Anyone"),
             InheritedPermissions::LOGGED_IN_USERS => _t(__CLASS__.'.ACCESSLOGGEDIN', "Logged-in users"),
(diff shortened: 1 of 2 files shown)


code/Model/SiteTree.php
```diff
@@ -2100,8 +2100,10 @@ public function getSettingsFields()
             }
         }
 
+        $inheritMessage = $this->ParentID !== 0 ? _t(__CLASS__.'.INHERITPARENT', "Inherit from parent page") : _t(__CLASS__.'.INHERITSITECONFIG', "Inherit from site access settings");
```
maxime-rainville: Just a pedantic comment about line breaks here, but this appeal to my sense of aesthetic more:
```php
        $inheritMessage = $this->ParentID !== 0 ? 
            _t(__CLASS__.'.INHERITPARENT', "Inherit from parent page") : 
            _t(__CLASS__.'.INHERITSITECONFIG', "Inherit from site access settings");
```

lang/en.yml
```diff
@@ -204,7 +204,8 @@ en:
     Editors: 'Editors Groups'
     HASBROKENLINKS: 'This page has broken links.'
     HTMLEDITORTITLE: Content
-    INHERIT: 'Inherit from parent page'
+    INHERITPARENT: 'Inherit from parent page'
```
maxime-rainville: I'm wondering, if we would be better to keep `INHERIT` here to avoid breaking every single translation we have for the existing key?

lukereative: You're right, I thought I'd done a search and there wasn't any other translations for that label but I must've searched wrong because there are, so I've reverted it.

code/Model/SiteTree.php
lukereative: Agree, I've updated to that format

dhensby: Is ParentID always an int? I thought it would normally be a string when pulled out of the DB...

code/Model/SiteTree.php
```diff
@@ -2100,8 +2100,12 @@ public function getSettingsFields()
             }
         }
 
+        $inheritMessage = $this->ParentID !== 0 ?
```
dhensby: I think strict type checking on DB fields is a bad idea as I'm pretty sure they are all loaded as strings

PR #2214, opened 2018-07-12
NEW add SiteTree::updateAnchorsOnPage() for defining content anchors

wilr:
The current AnchorSelectorField only takes into account a pages `Content` field. This causes issues with modules such as Elemental which don't use `$Content`.

Moves the anchor extraction functionality to `SiteTree::getAnchorsOnPage` function which then defines an extension hook to modify.


robbieaverill:
Looks good! This would need to go against the `4` branch as an enhancement though

wilr:
@robbieaverill changed base to `4`

PR #2215, opened 2018-07-15
NEW meta tag array

jonom:
Meta tags can now be individually accessed and modified by key, prior to being rendered and flattened in to a single string. This allows for much easier manipulation of meta tags when extending or subclassing SiteTree.

**Example:** override `MetaTagArray()` in Page.php to provide a dynamically generated meta title and description

```php
public function MetaTagArray()
{
    $tags = parent::MetaTagArray();
    $tags['title']['content'] = $this->obj('customMetaTitle')->forTemplate(); // needs to be html
    $tags['description'] = [
        'tag' => 'meta',
        'attributes' => [
            'name' => 'description',
            'content' => $this->customMetaDescription(), // raw text will be converted later
        ],
    ];
    return $tags;
}
```


code/Model/SiteTree.php
```diff
@@ -1338,61 +1338,102 @@ public function collateDescendants($condition, &$collator)
...
         if (Permission::check('CMS_ACCESS_CMSMain')
             && $this->ID > 0
         ) {
-            $tags[] = HTML::createTag('meta', array(
-                'name' => 'x-page-id',
-                'content' => $this->obj('ID')->forTemplate(),
-            ));
-            $tags[] = HTML::createTag('meta', array(
-                'name' => 'x-cms-edit-link',
-                'content' => $this->obj('CMSEditLink')->forTemplate(),
```
jonom: Side note: `HTML::createTag` calls `Convert::raw2att()` on each attribute so aren't all these calls to `->forTemplate()` (apart from the Title field) causing double escaping? Should I change them to call `->getValue()` instead, or skip the `->obj()` call to get the raw values?

jonom:
Most checks passed, is this a harmless fail?
> Script phpcs -s code/ tests/php/ tests/behat/src/ handling the lint event returned with error code 2

robbieaverill:
My only gripe is the naming of `MetaTagArray` - I don't like this, if it weren't for semver I'd say just rewrite the MetaTags method... I don't have a better suggestion though 😄

jonom:
I hear ya! I don't like it either, but I couldn't think of anything better since MetaTags is already taken as you point out. I guess `getMetaComponents` might be better?

jonom:
@robbieaverill could you sanity check this observation? https://github.com/silverstripe/silverstripe-cms/pull/2215/commits/619e7e7030b8dce3c8fc4db56973770e9febf343#r202553096

code/Model/SiteTree.php
robbieaverill: Currently SiteTree defines $casting for MetaTags to HTMLFragment, which doesn't escape when rendered in the template, so it shouldn't be double escaping unless people put `$MetaTags.XML` in their template rather than `$MetaTags`

It would probably be better to have a template include that loops through your new array method and renders the meta tags in the template instead of doing it all here - would you agree?

jonom: It's not the whole `$MetaTags` object that I'm talking about here, just some attributes on individual tags e.g. `$tags['cmsEditLink']['attributes']['content']` is set to `$this->obj('CMSEditLink')->forTemplate()` which I assume is an escaped value, which is then passed to `Convert::raw2att()`. It's actually only the `cmsEditLink` and `pageId` values that get this treatment, and I would guess no escaping is needed on those values which is why the double escaping hasn't surfaced as a problem, since escaping doesn't change the values. So I'd say it's probably harmless but may as well be corrected.

robbieaverill: Ah I see. Yeah - if we _were_ to move this into a template then the attribute escaping can be done explicitly in the template as well perhaps.

In the case of CMSEditLink, `forTemplate()` would cause whatever casting is defined for SiteTree which is Text, and that model does escape by default (`DBField::forTemplate`) so I think you're right, the Convert::raw2att here is causing double escaping and you're also right that it's not surfacing as a problem because of the contents of the existing data we're using =)

Nice to spitball this!

robbieaverill:
Yeah I suspect the behat failures are unrelated

jonom:
I got rid of that double escaping and changed `MetaTagsArray` to `MetaComponents`. Have also edited this PR's description to include a proposed doc entry to be added later.

code/Model/SiteTree.php
```diff
@@ -1338,51 +1338,103 @@ public function collateDescendants($condition, &$collator)
...
+                'attributes' => [
+                    'name' => 'x-cms-edit-link',
+                    'content' => $this->CMSEditLink(),
+                ],
+            ];
+        }
+
+        $this->extend('MetaComponents', $tags);
+
+        return $tags;
```
tractorcow: If $tags was an ArrayList instead of Array, we could use `<% loop $MetaComponents %>` in templates couldn't we?

You know, I just realised we really need a `casting` type for lists as well so we don't have to auto-wrap these in arraylists.

```
private static $casting = [
    'MetaComponents' => ArrayList::class,
];
```

Note: This doesn't work, I just wish it did. :P

jonom: > You know, I just realised we really need a casting type for lists as well so we don't have to auto-wrap these in arraylists.

That would be cool :)

> If $tags was an ArrayList instead of Array, we could use <% loop $MetaComponents %> in templates couldn't we?

The MetaTags are a bit of a mixed bag. We could use a template but the tag type is variable (due to inclusion of title tag 🙄), as are the attributes. So it would need to look something like:

```html
<$TagType <% if $Attributes %><% loop $Attributes %>$Name="$Value.ATT"<% end_loop %><% end_if %><% if $Contents %>>$Contents</$TagType><% else %> /><% end_if %>
```

Note that `$MetaTags` would need to render that template to retain backwards compatibility.

I don't think there's a big advantage in switching to a template at this point, unless it's with the view of removing `$MetaTags` in 5 and working only with `$MetaComponents` instead.

I'm not a big fan of changing  more than necessary where backwards compatibility is a priority. So far this PR isn't changing the way the data is rendered, I'm just exposing that data before it is flattened, so it should be nice and safe. Switching to templates probably wouldn't hurt but the more you change the more risk of side effects.

BTW I thought about splitting out the title tag from MetaComponents so it would only provide true `meta` tags, but the expectation is that `$MetaTags` can give you a title and I want `MetaComponents()` to be the source for everything in `$MetaTags`, so I decided to keep it all together. It makes the structure of the MetaComponents array more complicated than I'd like but I've tried to address that with clear documentation in the method and proposed addition to the Docs.

jonom:
Thanks @maxime-rainville! Your changes look good.

robbieaverill:
I've just fixed the phpcs violation and squashed it

jonom:
Not sure the best place for a docs entry about this. Their's a mention of meta tags here: https://docs.silverstripe.org/en/4/developer_guides/templates/common_variables/#meta-tags, should I add on to that or is it getting a bit off topic to add php code there?

jonom:
The docs file is in [framework](https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/02_Developer_Guides/01_Templates/02_Common_Variables.md) so I'll have to open a separate PR

jonom:
Thanks for the merge! Docs PR: https://github.com/silverstripe/silverstripe-framework/pull/8779

PR #2221, opened 2018-08-06
FIX Use LatestDraftVersion in GraphQL query to determine latest draft version

robbieaverill:
Parent issue: https://github.com/silverstripe/silverstripe-versioned-admin/issues/34


robbieaverill:
Retargeted at 4(.3) because this is really implemented in the new history viewer work which would be released with this line

PR #2226, opened 2018-08-10
FIX Separating ModelAsController catch-all route

### LINKED ISSUE #2225: Configured routes are being ignored
#2223 has introduced a critical bug where configured routes are now ignored as long as they are added after `#modelascontrollerroutes`.

A quick way to reproduce this is to install the `silverstripe/environmentcheck` module which configures [the route](https://github.com/silverstripe/silverstripe-environmentcheck/blob/master/_config/routes.yml) `health/check`. This is now 404-ing. This additionally means the GraphQL endpoint is 404-ing.

The catch all route is handling all requests after a certain point.

ScopeyNZ:
Fixes #2225 

By separating the catch all route off on its own and ensuring it applies after everything, we restore the original functionality while still resolving the issue the original change was trying to achieve.


ScopeyNZ:
I see. It's not possible to `after:` this config. However, with this PR you can still overload the `''` route which half resolves #2223? 🙃 

I figure the actual fix will require keeping some sort of registry of what `after: '*'` configs have been applied.

ScopeyNZ:
We'll have to either merge this or revert #2223 at some point though.

PR #2234, opened 2018-08-14
WIP: NEW Add Title, Last Edited and Last Editor as default columns in all CMS reports

robbieaverill:
Adds more default columns to all CMS reports.

image

image

The "Last Editor" column isn't able to be exported to CSV though, which should probably block this being merged. This is because there's no dot notation link available to get from the page to the Member who saved the latest version of it.

Resolves https://github.com/silverstripe/silverstripe-reports/issues/108

### DIFF
--- a/code/Reports/AbstractCMSReport.php
+++ b/code/Reports/AbstractCMSReport.php
@@ -1,0 +1,68 @@
+<?php
+
+namespace SilverStripe\CMS\Reports;
+
+use SilverStripe\CMS\Controllers\CMSPageEditController;
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Control\Controller;
+use SilverStripe\Reports\Report;
+use SilverStripe\Security\Member;
+use SilverStripe\Versioned\Versioned;
+
+/**
+ * Provides some common functionality for all CMS module reports
+ */
+abstract class AbstractCMSReport extends Report
+{
+    /**
+     * By default we provide the page title with a clickable link, the last edited date and the member's name
+     * who last edited the page.
+     *
+     * @return array[]
+     */
+    public function columns()
+    {
+        // @todo remove coupling to global state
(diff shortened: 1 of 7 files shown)


clarkepaul:
Nice one :)
Is the "Last editor" the same person as when it was last published, or could it be the person who last saved the item? I presume they match.

I think the column headers might make more sense if it was either:
1. Last edited & Author (could be saved or published)
2. Last published & Published by (only show last published)

With my preference being 1, and a shorter date format for smaller screens if it isn't a huge effort e.g. 2 August 2018, 11.06pm

code/Reports/AbstractCMSReport.php
```diff
@@ -0,0 +1,68 @@
...
+                    );
+                },
+            ],
+            'LastEdited' => [
+                'title' => $dateTitle,
+                'casting' => 'DBDatetime->Full',
+            ],
+            'AuthorID' => [
+                'title' =>  _t(__CLASS__ . '.LastEditor', 'Last Editor'),
+                'formatting' => function ($value, SiteTree $item) {
```
kinglozzer: If you added this as a getter to `SiteTree`, would that make it exportable?

robbieaverill: Yeah it would, but it's also couple it a bit more to versioned. I'm sure I can customise the export fields with callbacks, I just haven't looked at it enough yet 😆

robbieaverill:
I realise versioned is a dependency of the CMS and will always be, but it'd still be nice if I could achieve this without adding coupling to this module - will look at this again on hackday

brynwhyman:
Another idea for an additional column would be the 'stage' of the page, ie `published`, published (with changes)`, or `draft`. This is something that's provided in https://github.com/silverstripe/silverstripe-sitewidecontent-report

robbieaverill:
I'm probably not going to get this finished, so closing, but I'll leave the branch around if anyone wants to use it later

Issue #2241, opened 2018-08-23
Can't use CMS without reports module


baukezwaan:
SilverStripe 4.2.1

I would like to remove the reports section from the CMS. Now I could hide it by 

`CMSMenu::remove_menu_class(ReportAdmin::class);`

but I would like to completely remove it via compose.json, by:

```
“replace”: {
  “silverstripe/reports”: “*”
},
```

Problem now is that I get an PHP-error because there are 5 files in my project in /vendor/silverstripe/cms/code/Reports, like BrokenFilesReport.php and EmpyPagesReport.php.

Could these be moved to the reports repository?

# Related PRs
* [ ] https://github.com/silverstripe/silverstripe-cms/pull/2243

robbieaverill:
Yes, ideally the CMS reports would be in a separate module (not reports, but probably a new one). This could potentially be done as a minor semver change by deprecating the old classes, extending new ones in a new module and making them conditional on the parent class existing. Let’s see what the rest of the core team think before proceeding.

maxime-rainville:
Related to https://github.com/silverstripe/silverstripe-cms/issues/2242.

I'm all in favor of decoupling CMS from reports. Sounds a bit over the top create a brand new module for 2 reports however.

Could we get away with a simple `if class_exists` statement in those reports file?

On a related note, it could be interesting to see if we can create a _minimalist_ cms recipe for developers who want to keep things simple.

robbieaverill:
>Could we get away with a simple if class_exists statement in those reports file?

+1

baukezwaan:
You mean something like:

`if (!class_exists(ReportAdmin)) return false;`

That would probably work, but not so nice imho.

maxime-rainville:
Did a proof of concept PRs. https://github.com/silverstripe/silverstripe-cms/pull/2243

baukezwaan:
I've read the discussion in #2243 and it's not gonna happen in SS4.

Meanwhile the button can be hidden by:
```CMSMenu::remove_menu_class(ReportAdmin::class);```

Closing this issue for now

PR #2243, opened 2018-08-27
FIX Allow CMS to be used without Report

maxime-rainville:
Fixing https://github.com/silverstripe/silverstripe-cms/issues/2241

I've remove reports from the main composer dependencies and made it a dev dependency instead. I've also added it to the "suggests". Do let me know if this is not the recommended way of doing this.

### DIFF
--- a/code/Reports/BrokenFilesReport.php
+++ b/code/Reports/BrokenFilesReport.php
@@ -12,6 +12,10 @@ use SilverStripe\Reports\Report;
 use SilverStripe\ORM\DB;
 use SilverStripe\Versioned\Versioned;
 
+if (!class_exists(Report::class)) {
+    return;
+}
+
 class BrokenFilesReport extends Report
 {
 
--- a/code/Reports/BrokenFilesReport.php
+++ b/code/Reports/BrokenFilesReport.php
@@ -12,9 +12,11 @@ use SilverStripe\Reports\Report;
 use SilverStripe\ORM\DB;
 use SilverStripe\Versioned\Versioned;
 
+// phpcs:disable PSR1.Files.SideEffects
 if (!class_exists(Report::class)) {
     return;
 }
+// phpcs:enable
 
 class BrokenFilesReport extends Report
 {
--- a/code/Reports/BrokenLinksReport.php
+++ b/code/Reports/BrokenLinksReport.php
@@ -15,6 +15,10 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\Reports\Report;
 
+if (!class_exists(Report::class)) {
+    return;
+}
+
 /**
  * Content side-report listing pages with broken links
  */
--- a/code/Reports/BrokenLinksReport.php
+++ b/code/Reports/BrokenLinksReport.php
@@ -15,9 +15,11 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\Reports\Report;
 
+// phpcs:disable PSR1.Files.SideEffects
 if (!class_exists(Report::class)) {
     return;
 }
+// phpcs:enable
 
 /**
  * Content side-report listing pages with broken links
--- a/code/Reports/BrokenRedirectorPagesReport.php
+++ b/code/Reports/BrokenRedirectorPagesReport.php
@@ -11,9 +11,11 @@ use SilverStripe\ORM\DB;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\Reports\Report;
 
+// phpcs:disable PSR1.Files.SideEffects
 if (!class_exists(Report::class)) {
     return;
 }
+// phpcs:enable
 
 class BrokenRedirectorPagesReport extends Report
 {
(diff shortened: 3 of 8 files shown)


robbieaverill:
I'd suggest we should remove it from dependencies entirely and put it in suggests. We can still have it in the Travis builds by installing recipe-cms or explicitly requiring it

maxime-rainville:
@robbieaverill Agreed. We discussed yesterday we wanted to further decoupled our core modules. That's a nice contained issue to look at some of the potential problems with this.

maxime-rainville:
The build's failing because PHPCS doesn't lilke my `if (!class_exist(Report::class))` statements.

robbieaverill:
There's a phpcs rule you can disable (in phpcs.xml.dist) for that - it's one of the ones in framework's phpcs.xml.dist file

maxime-rainville:
I'll wait to see if we actually want to do this before I start disabling linting rules in framework.

robbieaverill:
I think we do want to do it. You could ask "how far do we go with granularity" at any stage, e.g. should GridField be it's own module? I think we should go ahead with this, @flamerohr's approved too

flamerohr:
do it :)
I mean, the other way you could do it is have a `silverstripe-cms-reports` module and pull these files in to that with cms and reports as dependencies, but I think that's undesirable

kinglozzer:
I’m fine with this change. My preference would be for these reports to live in the reports module with the reverse check - `if (!class_exists(SiteTree::class))` - but that could be done for SS5. Let’s keep it simple for now 😁

code/Reports/BrokenFilesReport.php
```diff
@@ -12,9 +12,11 @@
 use SilverStripe\ORM\DB;
 use SilverStripe\Versioned\Versioned;
 
+// phpcs:disable PSR1.Files.SideEffects
```
robbieaverill: I think it'd be better to put it into phpcs.xml.dist as an exclusion

robbieaverill:
> A downside of that is that suggest doesn't allow us to specify a version constraint. To take our current example, CMS use to require Report 4.1 or greater. I'm not sure what will happen if you try to install Report 4.0.

I guess if people are using a recipe then this won't be a problem. We can always start bumping constraints in the reports module too.

code/Reports/BrokenFilesReport.php
maxime-rainville: Just had a look at the [PSR1 standard](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md#23-side-effects).

Looks like this is the preferred way of achieving what we are aiming for.

```php
<?php
if (class_exists(Report::class)) {
    class RecentlyEditedReport extends Report
    {
        // ...
    }
}
```

ScopeyNZ: Interesting. Kind of goes against the "early return" stuff. But this stuff is all so magical anyway (class definitions are parsed before code is evaluated) - but PHP handles this somehow.

robbieaverill: I totally disagree with wrapping the entire definition in an if statement, hopefully that's not just me =)

maxime-rainville: It does look inelegant to my eye. However, I would prefer this to disabling `PSR1.Files.SideEffects` across the board.

flamerohr: I guess if it's already global in framework - keep it consistent firstly.

However, I'm keen for this change to be localised and "early return" practice :)

robbieaverill: I'd still really prefer to see these inline rules removed

wilr: Could we not have some way of only loading a class using the module api otherwise the manifest ignores it.

cms.yml
```
--
Name: cms-reports
Except:
  moduleexists: 'silverstripe/reports'
--
SilverStripe\Core\Manifest\ManifestFileFinder:
  dontAcceptDirs: 
     - 'src/Reports'
```

ScopeyNZ: Option 3 because having `if (class_exists()) return` is just so hacky in PHP. Function definitions prefer wrapping anyway... Having something in config to exclude directories would be a blocker for moving to PSR-4 autoloading.

robbieaverill: Option 1 is the current standard across all core modules

chillu:
As a general comment, some decoupling efforts are more worthwhile than others. If you don't want to expose reports to your authors, you can already remove it from the menu. without removing it from the code. Running less code (particularly less Config YAML and Extension) is always desireable of course. But I'd argue that 99% of use cases which need the `cms` module (not just the `admin` module) also benefit from the reports feature. Decoupling is a good goal, but the effort should always be sized in proportion to the payoff. Which seems to be the case here.

For comparison, decoupling the ORM is a *lot* more effort, but has potentially high payoffs (e.g. running SilverStripe core within a Laravel controller, or a Symfony-based GraphQL server).

Apart from that, I'm not too fussed on any of the three options.

maxime-rainville:
Doesn't look like we're going to reach a consensus on this. Let's maybe reconsider this for SS5.

Issue #2227, opened 2018-08-10
Link tracking adds extra tab to CMS edit form


kinglozzer:
4.2.0+

For DataObjects with a direct relation to a file, an extra tab is added called “LinkTracking”. `SiteTreeLinkTracking` should remove it with

```php
/**
 * @param FieldList $fields
 */
public function updateCMSFields(FieldList $fields)
{
    $fields->removeByName('LinkTracking');
}
```

### PR's
* [x] https://github.com/silverstripe/silverstripe-cms/pull/2247
* [x] https://github.com/silverstripe/silverstripe-framework/pull/8353

### Related Issues
* https://github.com/silverstripe/silverstripe-assets/issues/156

chillu:
I'm going to classify this as `impact/high` since we now have link tracking on things like "Member", without this object having *any* links to track by default. This is cluttering the UI and making it highly confusing.

chillu:
Talked with @lukereative @unclecheese and @maxime-rainville, we all agree that this tab should be opt-in to avoid that confusion. Files and images are an exception, but they also have a specialised "used by" view.

maxime-rainville:
1. I was telling you porkies ... You can define a config flag on the `DataExtension` and than override it on the specific class that the extension is applied to. Just make sure you fetch the config value through the `owner`. If the only thing this flag does is hide/show the link tracking tab, I might call `show_link_tracking`. `has` implied that it will alter how data will be handled in the background.
2. There's probably other objects that will need that flag set to true. Image/File, maybe Elemental data blocks. Remember that there's nothing stopping someone from sticking Files and Images in a GridField rather than managing them through asset-admin. I would expect them to behave c
3. Good
4. I just trying to think of the best place to document this. Maybe the [Relationship between Records](https://docs.silverstripe.org/en/4/developer_guides/model/relations/) or a brand new wider article about link tracking.

Don't forget to make sure the Used on field in asset admin still works afterwards.

chillu:
+1 on `show_link_tracking` approach

kinglozzer:
I like that approach, could we adopt it for file tracking too? https://github.com/silverstripe/silverstripe-assets/issues/156

chillu:
Good point, since you define this per `DataObject`, we should probably rename the implementation above to `show_sitetree_link_tracking` and add functionality for `show_file_link_tracking`? /cc @bergice

PR #2247, opened 2018-08-31
API: Add `show_link_tracking` config to `SiteTreeLinkTracking` extension

bergice:
This allows controlling visibility of the Link Tracking tab so not every DataObject will show it.


_config/linktracking.yml
```diff
@@ -4,3 +4,12 @@ Name: cmslinktracking
...
     - SilverStripe\CMS\Model\SiteTreeLinkTracking
+---
+Name: cmslinktrackingvisibility
+---
+SilverStripe\CMS\Model\SiteTree:
+  show_link_tracking: true
+SilverStripe\Assets\File:
+  show_link_tracking: true
+SilverStripe\Security\Member:
+  show_link_tracking: true
```
kinglozzer: What’s the use-case for this on `Member`? I’m drawing a blank 😬

dhensby: I suspect it's being set to true to mirror current behaviour?

dhensby:
This is a new config setting and feature, so it should be going into 4.x not 4.2.

kinglozzer:
> This is a new config setting and feature, so it should be going into 4.x not 4.2.

I’d argue it’s a bug fix for https://github.com/silverstripe/silverstripe-cms/issues/2227

_config/linktracking.yml
```diff
@@ -4,3 +4,8 @@ Name: cmslinktracking
 SilverStripe\ORM\DataObject:
   extensions:
     - SilverStripe\CMS\Model\SiteTreeLinkTracking
+---
+Name: cmslinktrackingvisibility
+---
+SilverStripe\Security\Member:
+  show_sitetree_link_tracking: true
```
lukereative: Super small thing, could you add a new line please (if you haven't already – you should get https://plugins.jetbrains.com/plugin/7294-editorconfig – that way it'll auto add one)

ScopeyNZ: New lines at the end of files is part of PHPStorm by default - just needs to be enabled 🙂 . Settings > Editor > General: 

image

PR #2252, opened 2018-09-05
fixed deprecated create_function() in PHP 7.2

zemiacsik:
fixed for "PHP Deprecated:  Function create_function() is deprecated in \cms\code\model\SiteTree.php on line 2116" while using PHP 7.2


robbieaverill:
Test failure is unrelated

zemiacsik:
Thanks, I was wondering about it.

robbieaverill:
I think we need to give Behat a bit more juice

PR #2253, opened 2018-09-05
Set locale on datefields from current user

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-admin/issues/631

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -922,11 +922,11 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         $dateFrom = DateField::create(
             'LastEditedFrom',
             _t('SilverStripe\\CMS\\Search\\SearchForm.FILTERDATEFROM', 'From')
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
Nice branch name. Is that the right base issue? 

I sort of feel like DateFields should use the current users locale by default.

unclecheese:
Yeah, I thought about that, but I think it's too much coupling. A better long-term fix would be to have the locale fallback on a setting, so that somewhere upstream we could just set `DateField::set_default_locale($currentUser->Locale)`, but that seems a bit too impactful given that this is for a story unrelated to DateField. Probably worth raising an issue.

robbieaverill:
Related: https://github.com/silverstripe/silverstripe-framework/issues/8339

unclecheese:
Actually on second look, this should be falling back to i18n::get_locale(), which is set in LeftAndMain::init(). This could be an issue isolated to our behat test.

robbieaverill:
I restarted Behat and it went green

PR #2258, opened 2018-09-12
Fix 'Insert links into a page' test

lukereative:
Fix tests for https://github.com/silverstripe/silverstripe-admin/pull/637 (they won't pass till this is merged in though)

### DIFF
--- a/tests/behat/features/insert-a-link.feature
+++ b/tests/behat/features/insert-a-link.feature
@@ -15,7 +15,7 @@ So that I can link to a external website or a page on my site
 
   Scenario: I can link to an internal page
     When I select "awesome" in the "Content" HTML field
-      And I press the "Insert Link" HTML field button
+      And I press the "Insert link" HTML field button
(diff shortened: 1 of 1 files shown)


robbieaverill:
The tests are failing =/

lukereative:
They won't pass until the mentioned PR is merged in (it's dependent on those changes)

ScopeyNZ:
> They won't pass until the mentioned PR is merged in (it's dependent on those changes)

Huh? It's a test on the PR branch so it runs behat with your updated changes? If this PR is supposed to fix behat I'm not sure that it's working correctly...

lukereative:
@robbieaverill These test changes nor the tests in the branch will pass until they are both merged in, as they are dependent on each other – this test looks for a different capitalisation which is only in the other branch.

robbieaverill:
Man I thought we'd gotten away from this kind of test coupling. Ok well I guess merge both then and make follow up PRs if they don't pass afterwards =/

PR #2262, opened 2018-09-13
FIX SiteTree Title field should have rounded corners before Update URL button is shown

### LINKED ISSUE #2261: Border radius is zero for top and bottom right on page's title field
**SS 4.3.x-dev**

image

Presumably this is because it turns into an input group when you modify the page title of a page that has already been saved:

image

It should still have a border radius before the Update URL button exists.

robbieaverill:
Fixes #2261


robbieaverill:
@ScopeyNZ fancy taking a look at this one at some point?

Issue #2250, opened 2018-09-04
/treeview endpoint is slow when a lot of nodes are in the tree


unclecheese:
Per https://github.com/silverstripe/silverstripe-admin/issues/571

Request admin/pages/edit/treeview

CMSMain::treeview() takes 12 to 18 seconds to generate response. This function just populates a template:

$this->renderWith($this->getTemplatesWithSuffix('_TreeView'))

Further investigation is needed to figure out which part of the template is slow. This issue is present on every CMS page which shows the site tree, but at least is triggered only once.

Summary:

Issue occurs once per every CMS page load.

## Pull Requests
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2248

caffeineinc:
I'm also finding issues with SiteTree performance, however i think the issues is largely deeper than _just_ rendering. I've got a SiteTree with ~6400 records, most of which is under a single node. Looking deeper, it feels like an issue with how `TreeDropdownField` populates its `MarkedSet`.

Do you think that it is related to this, or seperate?

caffeineinc:
Related: https://github.com/silverstripe/silverstripe-admin/issues/571

unclecheese:
The `TreeDropdownField` is definitely a bottleneck, and I'm looking into that, but are you saying it affects the tree view? It should only have an effect on the edit view.

caffeineinc:
So the tree drop down issue seems to be the fact it's calling `AllChildrenIncludingDeleted` twice, 
once for `markPartialTree` and once for `getChildrenAsArray` during render.  SilverStripe lumberjack can take care of some of this, but the whole SiteTree has to be returned and then filtered if you're searching within it. 

It makes more sense to me to do the filter on the initial query, because we're operating on less data being a subset that way.

The only other thing, is it calls the same function `AllChildrenIncludingDeleted` on every object in that partial tree, which for say SilverStripe Blog, is redundant. None of those items should have children due to lumberjack & flat structure

caffeineinc:
Hmm i think i see why now, `getSearchResults()` has to return all of the results, even if we're using a search query, as we might not know what permissions (e.g. `canView()`) the current user has. Sometimes the search query is vague or attributed to a large number of results, which then have to be iterated through twice for `markPartialTree` and once for `getChildrenAsArray`. 

Either way this looks like way too much work for these situations.

chrispenny:
We're at a point now where the `TreeDropdownField` is timing out.

[...]

PR #2266, opened 2018-09-18
FIX: Use Hierarchy::prepopulate_numchildren_cache in tree-generation

sminnee:
Only relevant if https://github.com/silverstripe/silverstripe-framework/pull/8380 is avialable,
however coded defensively so it can be merged before that PR if needs 
be.

See https://github.com/silverstripe/silverstripe-framework/issues/8379

# Parent issue
* silverstripe/silverstripe-cms#2250

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -54,6 +54,7 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 use SilverStripe\ORM\HiddenClass;
+use SilverStripe\ORM\Hierarchy\Hierarchy;
 use SilverStripe\ORM\Hierarchy\MarkedSet;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\ValidationResult;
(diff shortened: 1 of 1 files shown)


code/Controllers/CMSMain.php
```diff
@@ -491,13 +491,11 @@ public function LinkPreview()
...
     {
-        // Pre-cache sitetree version numbers for querying efficiency
-        Versioned::prepopulate_versionnumber_cache(SiteTree::class, Versioned::DRAFT);
-        Versioned::prepopulate_versionnumber_cache(SiteTree::class, Versioned::LIVE);
-
-        if (method_exists(Hierarchy::class, 'prepopulate_numchildren_cache')) {
-            Hierarchy::prepopulate_numchildren_cache(SiteTree::class, Versioned::DRAFT);
-        }
+        $filter = $this->getSearchFilter();
+        SiteTree::singleton()->prepopulateTreeDataCache(null, [
```
maxime-rainville: I think we should be consistent here and use `$this->config()->get('tree_class')` to decide what we are caching.

sminnee: Yes that would make sense. singleton($this->config()->get('tree_class'))`` most likely?

sminnee:
We can probably squish this into a single commit

PR #2278, opened 2018-09-26
PHP7.2 compatibility

RoyalPulp:
(affects v3)
each() and  create_function() are deprecated in PHP7.2 ,

-> I copied the code from the 4 branch - seems to work in my local environment

### DIFF
--- a/code/search/SearchForm.php
+++ b/code/search/SearchForm.php
@@ -180,14 +180,14 @@ class SearchForm extends Form {
         }
         // Add * to each keyword
         $splitWords = preg_split("/ +/", trim($keywords));
-        $newWords = [];
+        $newWords = array();
         for ($i = 0; $i < count($splitWords); $i++) {
             $word = $splitWords[$i];
-            if ($word[0] == '"') {
+            if ($word[0] === '"') {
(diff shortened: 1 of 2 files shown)


code/search/SearchForm.php
```diff
@@ -173,23 +173,31 @@ public function getResults($pageLength = null, $data = null){
...
-		return implode(" ", $newWords);
-	}
+	protected function addStarsToKeywords($keywords)
+    {
+        if (!trim($keywords)) {
+            return "";
+        }
+        // Add * to each keyword
+        $splitWords = preg_split("/ +/", trim($keywords));
+        $newWords = [];
```
robbieaverill: We still need to support PHP 5.3, could you please use `array()` here?

RoyalPulp:
hi robbie, if you know what has to be changed, could you you make the change? I am not really at my desk for 2 weeks - and I have then to look into, what has to be done to make it compatible...

robbieaverill:
Thanks for your efforts =)

robbieaverill:
Oops, we have a 3.7 patch branch for these kind of changes. I've cherry picked the commits back into 3.7

PR #2280, opened 2018-09-27
FIX Use correct subsites namespace in SiteTree and test classes

robbieaverill:
Also reduces the setUp() method visibility to what it should be


code/Model/SiteTree.php
```diff
@@ -1875,7 +1875,7 @@ public function getCMSFields()
                 'AbsoluteLink' => _t(__CLASS__.'.DependtPageColumnURL', 'URL'),
                 'DependentLinkType' => _t(__CLASS__.'.DependtPageColumnLinkType', 'Link type'),
             );
-            if (class_exists('Subsite')) {
+            if (class_exists(Subsite::class)) {
                 $dependentColumns['Subsite.Title'] = singleton('Subsite')->i18n_singular_name();
```
maxime-rainville: I think you've missed `singleton('Subsite')`

maxime-rainville: This crap itself when Subsite is installed.

robbieaverill: I obviously didn't test this. Thanks, updated.

PR #2281, opened 2018-09-28
NEW: Add CMSMain.enable_archive_warning_message config

sminnee:
This lets you disable this expensive message generation on larger sites
where that is a problem. Speeding it up is obviously a preferred
solution (or removing this feature until that is possible), but this
config option is a simpler solution to the problem in the short term.


ScopeyNZ:
We've had a client doing this via class extension but they opted for a generic message. I think it would be better for us to do the same?

`Warning: This page and all of its child pages will be unpublished before being sent to the archive.\n\nAre you sure you want to proceed?`

sminnee:
Good call.

sminnee:
Like this? I've changed the setting to "enable_dynamic_archive_warning_message" instead, since if it's set to false a static message is still shown.

@chillu or @robbieaverill might have a view on whether it should be true or false by default. it's true by default at the moment, so you need to make a yml config change to fix big sites.

PR #2282, opened 2018-10-02
FIX: Don’t hardcode ID#s into the tests.

sminnee:
This requires https://github.com/silverstripe/silverstripe-framework/pull/8434

And will enable https://github.com/silverstripe/silverstripe-framework/pull/8435

### DIFF
--- a/tests/behat/features/insert-a-link.feature
+++ b/tests/behat/features/insert-a-link.feature
@@ -1,4 +1,4 @@
-@assets 
+@assets
 Feature: Insert links into a page
 As a cms author
 I want to insert a link into my content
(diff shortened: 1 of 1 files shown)


tests/behat/features/insert-a-link.feature
```diff
@@ -33,7 +33,7 @@ So that I can link to a external website or a page on my site
     And I wait for 1 second
     And I select "youranchor" from "Form_EditorToolbarLinkForm_AnchorSelector"
     And I press the "Insert link" button
-    Then the "Content" HTML field should contain "<a href="[sitetree_link,id=3]#youranchor">awesome</a>"
+    Then the "Content" HTML field should contain /<a href="\[sitetree_link,id=[0-9]+\]#youranchor">awesome</a>/
```
robbieaverill: You know, I'm wondering whether we could split this up so it doesn't look at IDs at all. With a regex check like this we aren't concerned about "3" any more anyway... How about this:

```behat
Then the "Content" HTML field should contain "<a href="[file_link,id=""
And the "Content" HTML field should contain "#youranchor"
And the "Content" HTML field should contain "awesome"
```

This way we wouldn't need the framework regex change at all

robbieaverill:
Hey @sminnee, you closed the framework PR - should we close this one too?

maxime-rainville:
This doesn't look like a critical fix to me, so I don't think it should be merge in the `3` branch.

I'll close it now, but feel free to reopen if you disagree.

PR #2286, opened 2018-10-04
Typo

nxs47:


robbieaverill:
Hey @nxs47, thanks for taking the time to contribute!

We don't support SilverStripe 2.4 any more unfortunately - we actively support SilverStripe 4.x at the moment and SilverStripe 3.x in "limited support" (critical bug and security fixes).

If you'd like to contribute changes to the non-English translations, you can do so on Transifex.com.

For example, it looks like this word has been changed to "dass" since 2.4. If you'd like to change it to "daß" you can do so by signing up and modifying it here: https://www.transifex.com/silverstripe/silverstripe-admin/viewstrings/#de/master-js/112006106

Thanks again!

Issue #2276, opened 2018-09-25
SiteTree DependentPages performance issue


mfendeksilverstripe:
Version: `4.2.1` (`c611c981e45409c8ab318f709a5d94cdd7e0ce84`)

`SiteTree::DependentPages()` function creates an `ArrayList` of all dependant pages.

If you have a root page with 11 000 dependant pages the performance takes a big hit. It doesn't help that having this feature is mandatory and you can't remove this by configuration.

My recommendation is:

- do not use `ArrayList` unless there is no other choice
- make this feature be enabled/disabled via config
- make this feature show on demand, not during edit for the page load

# LazyLoading Pull Requests
* [ ] https://github.com/silverstripe/silverstripe-framework/pull/8460
* [ ] https://github.com/silverstripe/silverstripe-admin/pull/679
* [ ] https://github.com/silverstripe/silverstripe-cms/pull/2290

robbieaverill:
>make this feature show on demand, not during edit for the page load

This would be nice, but I'm not sure how feasible it would be to implement

chillu:
@maxime-rainville As discussed, please propose a longer term solution to this which doesn't put this auxiliary feature on the critical path of loading the edit form. I'd guess that less than 1% of edit form loads are done with the user goal of viewing dependant pages, so they shouldn't slow down every one of those loads. 

@clarkepaul Do you have ideas on how we could lazy load? The most obvious way would be to load them when opening the tab, right? Note that we'd need to make this work in both React and Entwine sections, since all kinds of records can have "dependant things" now.

maxime-rainville:
# Make Dependent Page render on its own controller
A somewhat left field idea could be to move it to it's own dedicated `CMSMain` controller, a bit like `CMSPageSettingsController`.

This would move it on equal footing with the `Content`, `Settings` and `History` tabs. You could make a straight face argument that this actually better UX: Dependent Pages are not part of the SiteTree content, so why should they be on the `Content` tab? It looks more like metadata about the page, that is not directly editable. (He said keeping a straight face)

# Make all form tabs lazy-loadable
We could try to update the basic form logic so form fields are only loaded/rendered when tabs are access. This could have across the board benefits for all tabbed forms, but could be moderately difficult to implement.

It would involved updating the form save logic to not assume that all form fields will always be present. Probably not reasonable to expect this to be done in time for the 4.3 release.

# Don't preload GridField's first page
[...]

PR #2290, opened 2018-10-09
MINOR Enable the lazy loading for DependentPages GridField

maxime-rainville:
Fix https://github.com/silverstripe/silverstripe-cms/issues/2276

Tweak the Dependent page GridField to use `GridFieldLazyLoader`.

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -1883,7 +1883,6 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         if ($dependentPagesCount) {
             $dependentColumns = array(
                 'Title' => $this->fieldLabel('Title'),
-                'AbsoluteLink' => _t(__CLASS__.'.DependtPageColumnURL', 'URL'),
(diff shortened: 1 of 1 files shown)


code/Model/SiteTree.php
```diff
@@ -1914,6 +1916,7 @@ public function getCMSFields()
                         );
                     }
                 ));
+            $dependentTable->getConfig()->addComponent(new GridFieldLazyLoader());
```
robbieaverill: Please use Injector to instantiate classes

maxime-rainville: None of the Gridfield components are injectable. Not sure if there's an underlying reason for this or if it's just an omission.

If we think it would be better to make them injectable, then we should probably have a separate job for this.

maxime-rainville:
Addressed all points. I've also remove the Absolute Link column following some discussion with Ingo.

I've created a follow up issue we could consider at a later date. https://github.com/silverstripe/silverstripe-cms/issues/2291

code/Model/SiteTree.php
robbieaverill: Yep, that's right. I'm happy to have a separate issue for it though

Issue #2277, opened 2018-09-26
SiteTree search form doesn't load in certain URLs


ScopeyNZ:
Using the new SiteTree search (`^4.3@dev`).

If I have versioned admin `1.x-dev` and I browse to the history section on a page the search form on the SiteTree never loads:

![image](https://user-images.githubusercontent.com/3260989/46049868-085e2980-c185-11e8-8cbf-add39e773f03.png)

It appears that the URL for the schema endpoint is not correct here. This means it miiight be a versioned-admin problem as it has a controller with its own "schema" endpoints, but versioned-admin does pass back to parent controllers given it doens't know about the form being requested.

maxime-rainville:
I set the impact to high because this needs to be resolved before we release SilverStripe 4.3.

maxime-rainville:
Found the problem. While the main request is handled by `CMSPageHistoryViewerController`, follow up requests (treeview and form schema) are handled by the original `CMSPageHistoryController`.

The reason for that is that `HistoryControllerFactory` can't find an ID attached to those follow up request. So `HistoryControllerFactory` will default to returning `CMSPageHistoryController`. 

Fixing `HistoryControllerFactory` is probably the best way of addressing this. We already have an issue opened for [defaulting to the new History controller](https://github.com/silverstripe/silverstripe-versioned-admin/issues/67). Those two issues are dealing with the exact same piece of code, so it's probably best to deal with them simultaneously and save ourselves are merge conflict.

On a related note, I'm getting an horizontal scrollbar in the site tree advanced filter when I force  `CMSPageHistoryViewerController`. We need to look at that as well.

![image.png](https://images.zenhubusercontent.com/5ac42fb74b5806bc2bcd8e13/c4645d24-70b7-4326-af63-3b9d7ff79e37)

PR #2300, opened 2018-10-24
MINOR Update CMS test to explicitetly register the CMSPageHistoryController

maxime-rainville:
Parent issues silverstripe/silverstripe-versioned-admin#67 and silverstripe/silverstripe-cms#2277

Once https://github.com/silverstripe/silverstripe-versioned-admin/pull/71 got merged in, the CMSPageHistoryController started failing because the new CMSPageHistoryViewerController was getting injected.


robbieaverill:
I guess this needs to be in the 4.3 branch, it may need to be cherry picked back into 4.3 from 4

PR #2306, opened 2018-10-31
FIX IE11+Edge17 Pages tree List View button

dnsl48:
Fixes: https://github.com/silverstripe/silverstripe-framework/issues/8551


client/src/legacy/CMSMain.js
```diff
@@ -73,7 +73,14 @@ $.entwine('ss', function ($) {
 
       localStorage.setItem('ss.pages-view-type', viewType);
       if(isContentViewInSidebar && viewType === VIEW_TYPE_LIST) {
-        window.location = $contentView.data('url-listviewroot');
+        const baseUrl = $('base').attr('href');
+
+        if (baseUrl) { // Edge17+IE11 fix
+            window.location.assign(baseUrl + $contentView.data('url-listviewroot'));
+        } else { // fall back to the original behaviour
+            window.location.assign($contentView.data('url-listviewroot'));
```
maxime-rainville: There's not much point having a fallback here. SilverStripe will always have a base tag on every page. So that if statement will always run.

Issue #2327, opened 2018-11-13
DependentPages should use a DataList, not an ArrayList


robbieaverill:
**Version: SS 4.3.x-dev**

The fact that `SiteTree::DependentPages()` generates an ArrayList means that it's executed whenever it runs. This is a major problem for high volume websites, particularly because this is called every time you call `getCMSFields()` on a page.

I really think we need to modify this functionality so it can return a lazily executed DataList instead.

robbieaverill:
I'm going to try out a POC

robbieaverill:
Another option is that we could add some config to disable back link tracking and dependent pages entirely. Large data sets make the CMS completely unusable by processing all of this data on every page load (in the CMS).

robbieaverill:
This seems to be working OK - we lose the identifier string for the type of relationship, but that could be added back in in the GridField that renders this list if we want to (by checking the class type)

ScopeyNZ:
Isn't this an issue because back links aren't necessarily pages? I think that's why it's using the `DependentLinkType` attribute right?

sminnee:
I think that that this would be _okay_ but really, the lesson here is that our public API should have been declared to be SS_List not ArrayList. We need to be much more careful about not being overly restrictive in how we declare our public API.

This also applies to, for example, the choice of private vs protected vs public members.

So the typing hinting should be changed to be SS_List, not DataList.

robbieaverill:
>Isn't this an issue because back links aren't necessarily pages? I think that's why it's using the DependentLinkType attribute right?

If that's the case then I guess we need to follow up on the todo in BackLinkTracking to implement a polymorphic many many list

maxime-rainville:
I've explored re-factoring DependentPage while looking at https://github.com/silverstripe/silverstripe-cms/issues/2276#issuecomment-426144450

That was my thought at the time. We decided to put that on the back burner because of time pressure, but I think all of my point are still relevant ... except maybe the "put your head in the sand" bit.

> # Problems with Dependent Pages logic
> 
> I've been digging on how `SiteTreeLink` and `SiteTree::DependentPage` have been implemented. 
> 
> `SiteTreeLink` has a Polymorphic `Parent` has-one relation and a `Linked` has-one relation to SiteTree` . `SiteTree::DependentPage` loops through all the `SiteTreeLink.Linked` for the current page, then all the VirtualPage, then all the RedirectorPage ... then it sticks everything in an ArrayList.
> 
[...]

PR #2330, opened 2018-11-14
MINOR Speed up DependentPages by shifting existence check to table join

maxime-rainville:
This PR updates `SiteTree::BackLinkTracking` to use some convoluted inner joins to avoid having to do individual query on each result to check for its existence.

It also alter queries to add the `DependentLinkType` alias field in the select statement which seems to give us a slight performance improvement over looping over all the results and adding it manually.

# Parent issue
* https://github.com/silverstripe/silverstripe-versioned/issues/189

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -1727,19 +1727,37 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     /**
      * Get the back-link tracking objects that link to this page
      *
-     * @retun ArrayList|DataObject[]
+     * @return ArrayList|DataObject[]
      */
     public function BackLinkTracking()
     {
         // @todo - Implement PolymorphicManyManyList to replace this
         $list = ArrayList::create();
-        foreach ($this->BackLinks() as $link) {
-            // Ensure parent record exists
-            $item = $link->Parent();
-            if ($item && $item->isInDB()) {
-                $list->push($item);
-            }
+
+        $joinClause = sprintf(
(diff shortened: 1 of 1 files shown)


robbieaverill:
Please consider #2327 when reviewing this. I don't think we should continue any longer using an ArrayList here, it's simply bad for performance

maxime-rainville:
Totally agree that this is just patching over the fundamental problem that `SiteTreeLink` is poorly thought.

I think this PR should be seen as a stopgap to allow us to ship key project. Rethinking Dependent Pages to work properly will probably require us to deprecated `SiteTreeLink` and build something else.

Looks like my change broke something. I'll have a look and try to fix it.

maxime-rainville:
Fixed it for real now. My change to `BackLinkTracking()` was returning a list of `SiteTreeLink` rather than returning the list of Parent Object like the original.

code/Model/SiteTree.php
```diff
@@ -1727,19 +1727,42 @@ public function getLiveURLSegment()
...
+
+        // Get the list of back links classes
+        $parentClasses = $this->BackLinks()->exclude(['ParentClass' => null])->columnUnique('ParentClass');
+
+        // Get list of sitreTreelink and join them to the their parent class to make sure we don't get orphan records.
+        foreach ($parentClasses as $parentClass) {
+            $joinClause = sprintf(
+                "\"%s\".\"ParentID\"=\"%s\".\"ID\"",
+                $siteTreelinkTable,
+                $linkTable,
```
robbieaverill:
```suggestion
```

PR #2333, opened 2018-11-16
BUGFIX reverting to publish still shows draft changes

### LINKED ISSUE #2305: "Cancel draft changes" leaves items in draft state & never disappears
Currently this button rolls back to the previously published version which is essentially "cancel(ing) draft changes". This actually creates a copy of the previously published version as the new saved state. This is an issue when you have relations that are owned by the page - they now show as draft. This is slightly confusing to the user because it is actually the same as the published version - so could be considered not actually in a draft state?

Additionally, with relations that are "owned" the "Cancel draft changes" option never disappears - even though the changed content has been discarded.

micmania1:
Fixed #2305 

The current version creates a new version of what is currently published. However, the version of the new draft doesn't match the version of what is published so the CMS things its unpublished.

This change publishes the new draft to ensure that the published/draft version is consistent throughout all systems (ie. CMS, changesets, campaigns etc.).

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -2081,6 +2081,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
             );
         } else {
             $record->doRevertToLive();
+            $record->publishRecursive();
             $message = _t(
                 __CLASS__ . '.ROLLEDBACKPUBv2',
                 "Rolled back to published version."


micmania1:
@ScopeyNZ yep, it does. It uses `rollbackRecursive()`

micmania1:
@ScopeyNZ changed branch to 4.3

ScopeyNZ:
Thanks. @silverstripe/core-team Any objections to this as a bugfix?

maxime-rainville:
Didn't test the fix, but it sounds sensible to me. 

Out of curiosity, was updating `Versioned::doRevertToLive()` to do a `publishRecursive` considered?

micmania1:
@maxime-rainville it was, but I thought putting it specifically where I wanted would be lower risk. I didn't want to cause any odd side-effects elsewhere.

PR #2352, opened 2018-12-08
MINOR: concept for a "add CMS Fields" shorthand

sunnysideup:
It usually takes a fair amount of time to add fields to the CMS for a page as they are not scaffolded. Doing so via shorthand method as suggested will speed this up a fair bit.  What is more, by using the scaffolding, you are less likely to make mistakes, etc...

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -2051,7 +2051,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
                 $tabName = _t(
                     __CLASS__ . '.EXRA_FIELDS',
                     "ExtraFields"
-                )
+                );
             }
             $fields->addFieldsToTab(
                 'Root.'.$tabName,
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -2025,6 +2025,41 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         return $fields;
     }
 
+    /**
+     * Quick way to add some fields to the CMSFields
+     *
+     * @param array $fieldsAsArray - list of fields to include, may include has_one, has_many, etc... e.g. `BackgroundImage`
+     * @param string $tabName - name of the tab to which they are to be adde, e.g. `MoreFields`
+     * @param null|FieldList - the existing CMSFields (or other FieldList) - OPTIONAL 
+     *
+     * @return FieldList
+     */
+    public function addFieldsToTab(array $fieldsAsArray, $tabName = '', $fieldList = null)
+    {
+            $fieldsToAdd = $this->scaffoldFormFields(
+                [
+                    'tabbed' => false,
+                    'includeRelations' => true,
+                    'restrictFields' => $fieldsAsArray,
+                    'ajaxSafe' => true
+                ]
+            );
+            if($fieldList === null) {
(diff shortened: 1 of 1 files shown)


maxime-rainville:
I agree that it's kind of weird that `SiteTree` doesn't use `scaffoldFormFields` like other DataObjects. That's probably something we should re-evaluate in SS5.

Looking at @sunnysideup's `addFieldsToTab`, I can remember coding variants of this method in a lot of projects over the years. I think we should seriously consider adding this method ... or some other variants of. This could be helpful on regular DataObjects as well.

maxime-rainville:
@silverstripe/core-team Does anyone else have views on this?

wilr:
SiteTree is a bit more sophisticated than most DataObjects (e.g getSettingsFields) etc. So I assume it would do something basic like dump everything into `Root.Main`. If we did do this I would want to see some shorter / easier API's for moving the fields to the correct tab and locations (even moving fields to a new tab is currently a pain - you have to remove then add). 

I'd imagine would need something like the following `$fields->moveFieldToTab()` `$fields->moveFieldAbove()`, $fields->moveFieldBelow()` at least, otherwise would end up with more code to move fields to the right places than you'd have without the magic of scaffolding.

kinglozzer:
Sounds like there’s potentially some crossover with https://github.com/silverstripe/silverstripe-framework/issues/8850?

code/Model/SiteTree.php
```diff
@@ -2025,6 +2025,41 @@ public function getCMSFields()
...
+            if($fieldList === null) {
+                $fieldList = $this->getCMSFields();   
+            }
+            if($tabName === '') {
+                $tabName = _t(
+                    __CLASS__ . '.EXRA_FIELDS',
+                    "ExtraFields"
+                );
+            }
+            $fields->addFieldsToTab(
```
tractorcow: `$fields` isn't defined.

tractorcow:
I kind of feel like this should be in the FormScaffolder API really, rather than SiteTree specifically. That whole API has aged quite poorly.

We've recently created FormFactory as a way of scaffolding a whole form at a time, rather than CMS fields alone. I wonder if we should focus on that instead of updating old APIs?

At the moment CMSMain doesn't use a FormFactory. Could there be a SiteTreeFormFactory?

chillu:
Having `SiteTree->addFieldsToTab()` behave differently from `FieldList->addFieldsToTab()` seems wrong. It has a different parameter signature, and the underlying intents are quite different. Even with a different name, I agree with Damian that this doesn't belong on `SiteTree`. 

But I think you're onto something there: Selective scaffolding of fields could be a big time saver. Way back when I built scaffolding, I decided not to implement it on `SiteTree->getCMSFields()`, both for backwards compat, and because it's quite an overhead. The same is true for `ModelAdmin` UIs of course: There's potentially a bunch of scaffolded fields which are thrown away and replaced with custom ones (needless processing). 

Maybe you could start this as a module, and then see if the API gets traction? You should be able to extend those methods onto `FieldList`, but would have to pass in the DataObject to scaffold on.

emteknetnz:
@sunnysideup  We’re currently going through pull-requests that haven’t had much activity for a while to see what is still active and what now makes sense to be closed.

It seems like there's going to be no further activity on this pull-request so we’ve closed this one.

Please feel free to open a new pull-request for anything else you think could be fixed or improved.

PR #2362, opened 2019-01-11
MINOR Do not assume SiteTree's table will always be call SiteTree.

maxime-rainville:
We assume to assuming SiteTree's table name as a constant.

That precludes some mad scientist from renaming that table in their YML config which I think should be a supported use case. e.g.:
```yml
SilverStripe\CMS\Model\SiteTree:
  table_name: SomethingOtherThanSiteTree
```

### DIFF
--- a/code/Tasks/MigrateSiteTreeLinkingTask.php
+++ b/code/Tasks/MigrateSiteTreeLinkingTask.php
@@ -5,6 +5,7 @@ namespace SilverStripe\CMS\Tasks;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Dev\BuildTask;
 use SilverStripe\Dev\Debug;
+use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DB;
 use SilverStripe\Versioned\Versioned;
 
@@ -35,11 +36,13 @@ class MigrateSiteTreeLinkingTask extends BuildTask
         Versioned::withVersionedMode(function () use (&$pages) {
             Versioned::set_stage(Versioned::DRAFT);
 
+            $sitetreeTbl = DataObject::singleton(SiteTree::class)->baseTable();
+
             /** @var SiteTree[] $linkedPages */
             $linkedPages = SiteTree::get()
                 ->innerJoin(
                     'SiteTree_LinkTracking',
-                    '"SiteTree_LinkTracking"."SiteTreeID" = "SiteTree"."ID"'
+                    "\"SiteTree_LinkTracking\".\"SiteTreeID\" = \"$sitetreeTbl\".\"ID\""
                 );
             foreach ($linkedPages as $page) {
                 // Command page to update symlink tracking
--- a/code/Tasks/SiteTreeMaintenanceTask.php
+++ b/code/Tasks/SiteTreeMaintenanceTask.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\CMS\Tasks;
 
+use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Controller;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\DataObject;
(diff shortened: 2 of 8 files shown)


code/Tasks/MigrateSiteTreeLinkingTask.php
```diff
@@ -35,11 +36,13 @@ public function run($request)
...
             Versioned::set_stage(Versioned::DRAFT);
 
+            $sitetreeTbl = DataObject::singleton(SiteTree::class)->baseTable();
+
             /** @var SiteTree[] $linkedPages */
             $linkedPages = SiteTree::get()
                 ->innerJoin(
                     'SiteTree_LinkTracking',
-                    '"SiteTree_LinkTracking"."SiteTreeID" = "SiteTree"."ID"'
+                    "\"SiteTree_LinkTracking\".\"SiteTreeID\" = \"$sitetreeTbl\".\"ID\""
```
robbieaverill: I assume you’ve thought about this already, but do you want to use the variable for the first part of this name? The pivot table will always start with the base table name

code/Tasks/SiteTreeMaintenanceTask.php
```diff
@@ -14,22 +15,23 @@ class SiteTreeMaintenanceTask extends Controller
...
+        $pages = DataObject::get(SiteTree::class, "\"$table\".\"URLSegment\" IN ($badURLs)");
 
         foreach ($pages as $page) {
             echo "<li>$page->Title: ";
             $urlSegment = $page->URLSegment;
             $page->write();
             if ($urlSegment != $page->URLSegment) {
                 echo _t(
-                    'SilverStripe\\CMS\\Model\\SiteTree.LINKSCHANGEDTO',
+                    SiteTree::class . '.LINKSCHANGEDTO',
```
robbieaverill: Unfortunately this doesn’t work on the text collector yet, it’ll need to stay the way it was

maxime-rainville: Change has been undone.

tractorcow:
I'm not sure I agree with this change; Allowing something to be variable means we need to support it as variable everywhere. We are adding a huge complexity / maintenance burden for no benefit.

Convention over configuration should rule here, and we shouldn't need to support core tables being renamed. 👎  from me I'm afraid.

robbieaverill:
I would’ve thought the convention is to do this, but I respect your point of view. It’s pretty low impact and am happy for you to close this without merge

maxime-rainville:
Not being able to have namespaced table name makes it more likely that we'll end up with collisions. That's not a big issue because in most cases people can just override the table name if they need to.

However that relies on us being consistent about how we get our table names.

Admittedly, you would need to be a bit crazy to want to rename the SiteTree table. But as a general principal, I think we should not assume table name when writting SQL queries.

maxime-rainville:
@silverstripe/core-team Does anyone else have views on whether hardcoding table names in SQL queries is kosher when working with core DataObject like SiteTree?

robbieaverill:
I agree with @tractorcow's thinking, but I still think we should be dogfooding our own stuff as well. If there's no significant performance impact from doing it I don't see why we shouldn't.

ScopeyNZ:
I don't want to say "We support renaming the SiteTree table" but I see no harm in accepting this code. I can't imagine there's any reasonable performance impact.

dhensby:
> @silverstripe/core-team Does anyone else have views on whether hardcoding table names in SQL queries is kosher when working with core DataObject like SiteTree?

It is definitely not kosher

sminnee:
Yeah I’m happy with this change, although I share Guy’s comment about it being a bit of an experimental feature.

chillu:
👍 on less hardcoding (what this PR does), 👎 on mentioning "rename sitetree table" as a supported feature anywhere (what the PR doesn't attempt)

sminnee:
Needs a rebase, conflict resolution, and a linting fix in SiteTree.php @maxime-rainville? Then we can merge.

You can probably squash the 2 commits into 1 as well.

maxime-rainville:
I've just rebased and fixed some linting issues.

maxime-rainville:
All green. I think the consensus is that we shouldn't officially support renaming the `SiteTree` table, while we should avoid hardcoding table names as a best practice.

Unless someone objects, I'll merge.

PR #2363, opened 2019-01-11
MINOR Clean up CMSMain::CanOrganiseSitetree

maxime-rainville:
This is a follow up from this PR. https://github.com/silverstripe/silverstripe-cms/pull/2033

Basically `CMSMain::CanOrganiseSitetree` is kind of coded in a silly way right now.

This makes it unsilly.

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -756,7 +756,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         if (!SecurityToken::inst()->checkRequest($request)) {
             return $this->httpError(400);
         }
-        if (!Permission::check('SITETREE_REORGANISE') && !Permission::check('ADMIN')) {
+        if (!$this->CanOrganiseSitetree()) {
             return $this->httpError(
                 403,
                 _t(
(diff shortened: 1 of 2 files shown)


code/Controllers/CMSMain.php
```diff
@@ -862,11 +862,16 @@ public function savetreenode($request)
             ->setBody(json_encode($statusUpdates));
     }
 
+    /**
+     * Whatever the current member has the permission to reorganise SiteTree objects.
```
robbieaverill:
```suggestion
     * Whether the current member has the permission to reorganise SiteTree objects.
```

code/Controllers/CMSMain.php
```diff
@@ -862,11 +862,16 @@ public function savetreenode($request)
...
+     * @return bool
+     */
     public function CanOrganiseSitetree()
     {
-        return !Permission::check('SITETREE_REORGANISE') && !Permission::check('ADMIN') ? false : true;
+        return Permission::check('SITETREE_REORGANISE');
```
robbieaverill: Is the admin check implicit?

robbieaverill: I see, this replaces #2033 and implements the suggested feedback from it

maxime-rainville: Yes and there's an entry in the unit tests to check for both SITETREE_REORGANISE and ADMIN to be extra sure.

code/Controllers/CMSMain.php
maxime-rainville: One day, I will spell `Whether` correctly.

code/Controllers/CMSMain.php
robbieaverill: Lovely!

tests/php/Controllers/CMSMainTest.php
```diff
@@ -660,4 +660,24 @@ public function testSearchField()
...
+
+        $this->logOut();
+        $this->logInWithPermission('SITETREE_REORGANISE');
+        $this->assertTrue($cms->CanOrganiseSitetree());
+
+        $this->logOut();
+        $this->logInWithPermission('ADMIN');
+        $this->assertTrue($cms->CanOrganiseSitetree());
+
+
```
robbieaverill: phpcs violation here - see https://travis-ci.org/silverstripe/silverstripe-cms/jobs/478282588#L786-L796

PR #2364, opened 2019-01-14
NEW Add font-icon support for site tree

sachajudd:
Resolves: https://github.com/silverstripe/silverstripe-admin/issues/574
Relies on: https://github.com/silverstripe/silverstripe-admin/pull/803

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -277,6 +277,13 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      */
     private static $icon = null;
 
+    /**
+     * Class attached to page icons in the CMS page tree. Also supports font-icon set.
+     * @config
+     * @var string
+     */
+    private static $icon_class = 'font-icon-page';
+
     private static $extensions = [
         Hierarchy::class,
         Versioned::class,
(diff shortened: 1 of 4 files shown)


code/Model/SiteTree.php
```diff
@@ -277,6 +277,13 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      */
     private static $icon = null;
 
+    /**
+     * Class attached to page icons in the CMS page tree. Also supports font-icon set.
+     * @config
+     * @var string
+     */
+    private static $icon_class = 'font-icon-page';
```
maxime-rainville: Can we document this somewhere?

code/Model/SiteTree.php
```diff
@@ -2747,7 +2754,8 @@ public function getTreeTitle()
         $children = $this->creatableChildren();
         $flags = $this->getStatusFlags();
         $treeTitle = sprintf(
-            "<span class=\"jstree-pageicon page-icon class-%s\"></span><span class=\"item\" data-allowedchildren=\"%s\">%s</span>",
+            "<span class=\"jstree-pageicon page-icon %s class-%s\"></span><span class=\"item\" data-allowedchildren=\"%s\">%s</span>",
```
maxime-rainville: Not a big fan of writing HTML in PHP. But it's no worst than it was.

code/Model/SiteTree.php
sachajudd: Yeah sure thing, I'll create a PR later today 🙂

code/Model/SiteTree.php
```diff
@@ -2747,7 +2754,8 @@ public function getTreeTitle()
...
+            $this->config()->get('icon') ? '' : $this->config()->get('icon_class'),
```
robbieaverill: I feel like this deserves to be in its own method, e.g. `getIconClass()` or something, but it's not a blocker

Issue #488, opened 2013-04-03
[2012-06-28] Folder class should use DataObjects to create & delete File records


silverstripe-issues:
created by: Jules
created at: 2012-06-28
original ticket: http://open.silverstripe.org/ticket/7582

---

It's not possible to effectively use the File object's onBeforeWrite & OnBeforeDelete hooks with Files & Images (/admin/assets).  AssetAdmin creates & deletes File records by calling methods of the Folder class which execute SQL, bypassing DataObjects.

e.g. in sapphire/filesystem/Folder.php, Folder::syncChildren has...
DB::query("DELETE FROM \"File\" WHERE \"ID\" = $unwantedDbChild->ID");
...where it should have...
$unwantedDbChild->delete();

Folder::constructChild also uses an SQL insert.

The same code is in all versions I checked from 2.4.4 to 3.0.0-rc3.

I came across this problem investigating why the Lucene search index is not updated when documents are added & removed.  If this change was made, new documents would be indexed immdiately & automatically.  The Lucene plugin documentation recommends scheduling a task to update the entire index, presumably nightly for large document archives.

There are many other good reasons to hook into file management e.g:  automatically optimise high res images uploaded from users' digital cameras, keep track of disk quotas, for security/moderation reasons.  None are possible because the hooks aren't called.

Please change the code in keeping with SilverStripe's principle of extendability.

Thanks,
Jules

wilr:
File system in 4.0 has been overhauled.

PR #2365, opened 2019-01-14
FIX Multibyte URL routing

chillu:
Was double url encoding (once in database value, then again in request)
Fixes https://github.com/silverstripe/silverstripe-framework/issues/8723

### DIFF
--- a/code/Controllers/ContentController.php
+++ b/code/Controllers/ContentController.php
@@ -25,6 +25,7 @@ use SilverStripe\Security\Security;
 use SilverStripe\SiteConfig\SiteConfig;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\View\ArrayData;
+use SilverStripe\View\Parsers\URLSegmentFilter;
 use SilverStripe\View\Requirements;
 use SilverStripe\View\SSViewer;
 use Translatable;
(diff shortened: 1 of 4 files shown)


code/Controllers/ContentController.php
```diff
@@ -195,11 +196,16 @@ public function handleRequest(HTTPRequest $request)
             if (class_exists('Translatable')) {
                 Translatable::disable_locale_filter();
             }
+
+            $filter = URLSegmentFilter::create();
```
maxime-rainville: I think we should be consistent with what `SiteTree` does. In the one place where it checks if multibyte character are allowed, it reads it from the config. https://github.com/silverstripe/silverstripe-cms/blob/097d2211507576dc618782642017d5bce2312e9e/code/Model/SiteTree.php#L1942

If someone ends up using using the injector to define their own `URLSegmentFitler`, you could end up in a situation where `SiteTree` and `ContentController` behave differently.

chillu: I've fixed that by correcting `SiteTree` to use `URLSegmentFilter::create()` instead. The instance is already used in a few other places where the logic is actually applied, this SiteTree line only adds some field descriptions.

chillu:
> There's a somewhat weird edge case where you create a URL Segment with utf8 characters and then disable it default_allow_multibyte

Yeah, then you're getting your site into an inconsistent state, I don't think we need to cater for that.

Pushed the `SiteTree` fix, ready for merge?

ScopeyNZ:
Noting that this PR has broken two tests in blog:

https://github.com/silverstripe/silverstripe-blog/blob/db18b5d31f49be7e5ef6295b530a51147c41c8fd/tests/BlogFunctionalTest.php#L31-L45

I haven't investigated why yet - perhaps @maxime-rainville or @chillu have some idea?

robbieaverill:
We probably have a workaround in the blog module to compensate for it. I'll take a look at it today.

robbieaverill:
Thoughts on backporting this to 4.0 so we don't need to have awkward workarounds in the blog module?

robbieaverill:
PR to backport it at #488 with linked justification

PR #2366, opened 2019-01-16
FIX: Fixes #2110 added default Title value for saved pages.

### LINKED ISSUE #2110: You shouldn't be able to save a page without page name and navigation label
It feels like a page should always have a name and navigation label, possibly it should be a required field—or at least get a warning. This would prevent empty items in the tree, and unintentional empty fields.

If they don't want the page to be displayed in the template devs should be able to accommodate that scenario.

Note an empty item in the tree:
image

# Related PR
* [ ] https://github.com/silverstripe/silverstripe-cms/pull/2366

ttunua:
Closes #2110

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -1514,9 +1514,9 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         // need to set the default values of a page e.g."Untitled [Page type]"
         if (empty($this->Title)) {
             $this->Title = _t(
-            'SilverStripe\\CMS\\Controllers\\CMSMain.NEWPAGE',
-            'New {pagetype}',
-            ['pagetype' => $this->i18n_singular_name()]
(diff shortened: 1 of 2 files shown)


ScopeyNZ:
Thanks for the PR @ttunua . It looks like this is still blocked a little by ongoing discussion in the attached issue. I'm not sure a consensus has been reached.

Additionally there's a failing test and some linting violations that will need to be corrected. I'll try to spur on some discussion on the related issue for you.

ttunua:
added some fixes, please review when you can

ScopeyNZ:
Thanks @ttunua it's looking good. There's still some debate (I think) about whether it should be "New [page type]" or "Untitled [page type]". We'll see what's decided before we merge anything.

Thanks for the contribution!

ScopeyNZ:
I think we can merge this if we change the default to "Untitled". That seems to be winning the vote in the attached issue.

PR #2370, opened 2019-01-17
BUG Don't convert datalist to arraylist when filtering pages

### LINKED ISSUE #2354: CMSSiteTreeFilter's run out of memory on large sites
When you have a site with 17,000 (number of pages on the site I am testing, not a bottom limit) then you try to filter all the pages by just the `CMSSiteTreeFilter` e.g. `CMSSiteTreeFilter_PublishedPages` it will exhaust your allocated memory limit (most likely)

This is because we get every page on the site (17,000) and then apply the below code:
```php
$pages = $pages->filterByCallback(function (SiteTree $page) {
    return $page->isPublished();
});
```

This means we end up with an array that slowly grows as we check if each page individually has been published.

tractorcow:
Fixes #2354


code/Controllers/CMSSiteTreeFilter_PublishedPages.php
```diff
@@ -41,11 +40,10 @@ public static function title()
...
-        $pages = $this->applyDefaultFilters($pages);
-        $pages = $pages->filterByCallback(function (SiteTree $page) {
-            return $page->isPublished();
-        });
-        return $pages;
+        return SiteTree::get()
+            ->innerJoin(
+                'SiteTree_Live',
+                '"SiteTree"."ID" = "SiteTree_Live"."ID"'
+            );
```
ScopeyNZ: Yeah that's a lot simpler than a `WHERE EXISTS` 😄

tractorcow:
Failing on live, but not on draft, which we don't really handle anymore.

I guess we can support it but we probably should try to deprecate live-only pages.

robbieaverill:
I agree in terms of actual records - they shouldn’t be in live but not in draft. Relationships could be though, but I don’t think that’s what you’re referring to so I agree overall

tractorcow:
Updated. :)

robbieaverill:
@maxime-rainville wanna take a look before merge?

PR #2374, opened 2019-01-29
Enforce canCreate() in "add page" controller

chillu:
It's already enforced when processing the form,
so there's no reason to list those page types in the first place.
This is a common way to make the CMS less cluttered,
e.g. by creating a DataExtension to remove VirtualPage
from the page types unless you're an admin.

Technically, this changes the CMSMain->PageTypes() API,
but in practice it's only used in CMSPageAddController.
It already respects need_permission, which is described as follows:
> List of permission codes a user can have to allow a user to create a page of this type.

Hence I'm considering the lack of canCreate() checks a bug in this context.

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -1186,11 +1186,18 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
         foreach ($classes as $class) {
             $instance = SiteTree::singleton($class);
+
+            // skip if implements HiddenClass
             if ($instance instanceof HiddenClass) {
                 continue;
             }
 
-            // skip this type if it is restricted
+            // skip if access is restricted via custom permission checks
+            if (!$instance->canCreate()) {
+                continue;
+            }
+
+            // skip this type if access is restricted via config
             $needPermissions = $instance->config()->get('need_permission');
             if ($needPermissions && !$this->can($needPermissions)) {
                 continue;
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -484,7 +484,8 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     }
 
     /**
-     * Return a subclass map of SiteTree that shouldn't be hidden through {@link SiteTree::$hide_ancestor}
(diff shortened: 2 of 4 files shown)


robbieaverill:
Note to self - this might break the subsites test suite, need to test

chillu:
I actually think this will break adding pages which return can create false on top level pages because of JS logic. Needs some more testing

maxime-rainville:
Just played with it a bit. It seems to be working OK for pages that's can't be added under the current node.

image

The unavailable page type just get greyed out as usual.

robbieaverill:
Should be ok, subsite filtering is done in the backend

chillu:
More research:

 * `SiteTree->canCreate()` supports an optional `$context['Parent']`, so we have to assume that subclasses of this also use it
 * `LeftAndMain->save()` uses `singleton($this->config()->get('tree_class'))->canCreate())` without a `Parent` context. That might be considered a bug
 * `CMSMain->childfilter()` is used in `CMSPageAddController`, and does pass in the context (`$obj->canCreate(null, array('Parent' => $parent)`)
 * If you change the parent page in the "add page" dialog, it'll refresh based on the new context

Just to clarify the motivations for this PR a bit more: I want to reduce the noise for authors in the "add page" dialog. There's some page types where creation are allowed/restricted in certain hierarchies (e.g. toplevel only, or below "blog holder" only). This is already handled through `can_be_root` and `CMSMain->childfilter()`. No changes there.

I'm targeting page types which have custom `canCreate()` implementations (e.g. `ADMIN` only). If a non-admin is never allowed to create them, regardless of the parent, we shouldn't show them.

But as I feared, this breaks pages which return `canCreate=false` on toplevel pages, but `canCreate=true` otherwise. And that's not a simple fix: `CMSMain->childfilter()` returns a list of "disallowed children", not a new full list of all page types. 

You can reproduce this issue with this test class:

```php
<?php
class ToplevelOnlyPage extends Page
{
    public function canCreate($member = null, $context = [])
    {
        if ($context && $context['Parent']) {
            return ($context['Parent']->ID != 0);
        }

        return false;
    }
}
```

I think we'll have to wait with fixing this until we can refactor the "add page" dialog to be a proper API driven UI. Closing.

PR #2388, opened 2019-02-12
Fix linking anchor on the same page

wilr:
To reproduce this issue open a page, add an anchor, save the page, attempt to link to newly created anchor on the same page.

Previous incorrect result will be the new anchor does not appear in the list as the list is not refetched off the server. This patch ensures that the page anchors are reloaded.


robbieaverill:
The test failure is valid

wilr:
@robbieaverill  Is the test valid though? Why would you load the anchors without  pageId?

wilr:
Ah sorry yep I see what the purpose was for that test. I've updated the code to achieve what the test meant.

ScopeyNZ:
Thanks @wilr. Can you make sure you're committing the bundle as well?

wilr:
@ScopeyNZ I get this while trying to build (not watch) ```ReferenceError: Unknown plugin "transform-object-rest-spread" specified in "silverstripe/admin/package.json" at 0, attempted to resolve relative to "silverstripe/admin"```

ScopeyNZ:
You will probably need to `yarn` in admin too. Also it looks like a rebase is needed.

I'll fix it up for you. As this is a bugfix did you want to target the next patch release instead of the next minor?

ScopeyNZ:
There you go - all fixed up and targetted at the next patch 🙂

ScopeyNZ:
Closing/re-opening to get travis working correctly

Issue #2369, opened 2019-01-16
[4.4] SiteTree font icons aren't displaying in the "add new page" menu


robbieaverill:
**4.x-dev**

Introduced with silverstripe/silverstripe-admin#574

* Right click on a page in the site tree
* Hover over "Add new page here"

Expected: page types should have icons
Actual: they don't

SS 4.3:
![image](https://user-images.githubusercontent.com/5170590/51241212-a288ee00-197d-11e9-9811-8b4de76f33e5.png)

SS 4.4:
![image](https://user-images.githubusercontent.com/5170590/51241232-a87ecf00-197d-11e9-804b-fa7b3bf64a84.png)

robbieaverill:
So this used to be based on the class name of the page. I think those styles were removed. I think we'll need to adjust the `allowedChildren` values in CMSMain.Tree.js a little so they can contain the value of the icon class, filename or file contents to get this to work

maxime-rainville:
@sachajudd ^

sachajudd:
Thanks for all these catches! Really appreciate all the testing going on 🙂hopefully I can look into these end of next week 🙂

PR #2391, opened 2019-02-25
FIX Deprecate creatableChildren and add new function to support font-icon classes for allowedChildren

sachajudd:
Resolves: #2369
Admin PR for hover colour: https://github.com/silverstripe/silverstripe-admin/pull/835


code/Model/SiteTree.php
```diff
@@ -2781,20 +2818,34 @@ public function getStatusFlags($cached = true)
...
+    /**
+     * Returns the CSS class used for the page icon in the site tree.
+     *
+     * @return string
+     */
+    public function getIconClass()
+    {
+        if ($this->config()->get('icon')) {
+            return '';
+        } else {
```
ScopeyNZ:
```suggestion
        }
```

code/Model/SiteTree.php
```diff
@@ -2781,20 +2818,34 @@ public function getStatusFlags($cached = true)
...
+            return $this->config()->get('icon_class');
+        }
```
ScopeyNZ:
```suggestion
```

code/Model/SiteTree.php
```diff
@@ -2781,20 +2818,34 @@ public function getStatusFlags($cached = true)
...
+     * Returns the CSS class used for the page icon in the site tree.
+     *
+     * @return string
+     */
+    public function getIconClass()
+    {
+        if ($this->config()->get('icon')) {
+            return '';
+        } else {
+            return $this->config()->get('icon_class');
```
ScopeyNZ:
```suggestion
        return $this->config()->get('icon_class');
```

Issue #2406, opened 2019-04-06
publishall needs overhaul


sunnysideup:
see: https://github.com/silverstripe/silverstripe-cms/blob/4/code/Controllers/CMSMain.php#L2147-L2210

There are a few issues:

1. limit should be $pages->limit(30, $start);
(the current setup probably works, but it is not confusing as it is the exact opposite in ORM

2. it chokes when you run it due to excessive use of memory AFAIK

3. you can not run it from the command line

sunnysideup:
Here is what I have written for our project:

```
<?php

namespace BSA\App\Tasks;

use SilverStripe\Control\Director;
use SilverStripe\Control\Controller;
use SilverStripe\Core\Environment;
use SilverStripe\Dev\BuildTask;
use SilverStripe\Security\SecurityToken;
use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\Forms\FieldList;

class PublishAll extends BuildTask
{

    protected $title = 'Publish All Pages';

    protected $description = 'Publish All Pages on the Entire Site (copy content from draft to live)';

    protected $step = 10;

    public function run($request)
    {
        Environment::increaseTimeLimitTo();
        Environment::increaseMemoryLimitTo();
        if ($request->requestVar('confirm') || Director::is_cli()) {
            // Protect against CSRF on destructive action
            if (Director::is_cli() || SecurityToken::inst()->checkRequest($request)) {
                $start = 0;
                $pages = SiteTree::get()->limit($this->step, $start);
                $this->flushNow('<ol>');
                while ($pages && $pages->count()) {
                    foreach ($pages as $page) {
                        $this->flushNow('<li>publishing: '.$page->Title.'</li>');
                        $page->publishRecursive();
                        $page->destroy();
                        unset($page);
                    }
                    $start += 30;
                    $pages = SiteTree::get()->limit($this->step, $start);
                }
                $this->flushNow('</ol>');
                $this->flushNow(
                    _t(__CLASS__ . '.PUBPAGES', 'Done: Published {count} pages', array('count' => $count))
                );
            } else {
                return Controller::curr()->httpError(400);
            }
        } else {
            $response = '';
            $token = SecurityToken::inst();
            $fields = FieldList::create();
            $token->updateFieldSet($fields);
            $tokenField = $fields->first();
            $tokenHtml = ($tokenField) ? $tokenField->FieldHolder() : '';
            $publishAllDescription = _t(
[...]

PR #2407, opened 2019-04-09
API Deprecated CMSMain->publishall()

chillu:
The current implementation doesn't scale,
and due to the proliferation of versioned objects
no longer fully works (e.g. doesn't publish all files).

Fixes https://github.com/silverstripe/silverstripe-cms/issues/2406


kinglozzer:
IIRC, this page has absolutely no styling - I’d always assumed it was legacy code that someone forgot to remove 😝

PR #2409, opened 2019-04-12
FIX: VirtualPage missing methods from target page (fixes #2408)

### LINKED ISSUE #2408: VirtualPage + relations are partially broken
## The problem

If you point a virtual page to a page that has, for example, a `has_many` relationship, the related objects will be fetched/rendered as expected when you visit the virtual page directly. However, if the virtual page is rendered in another context (e.g. in a menu), those relations are lost as `__call()` is never invoked.

For example, given a tree containing just two top-level pages - a source page and a virtual page pointing to it:
```
<% loop $Menu(1) %>
    {$Title} - {$CarouselSlides.Count}<br />
<% end_loop %>

will render...

Source Page - 3
Virtual Page - 
```

However, visiting the virtual page will correctly show the related objects.

## The cause

The reason everything works as expected when you visit the virtual page is that the `VirtualPage` instance is wrapped in a `PageController`. Top-to-bottom, the call stack goes:

- `ContentController::__construct($virtualPage)`
- [`ViewableData::setFailover($virtualPage)`](https://github.com/silverstripe/silverstripe-cms/blob/d4571d927dbe60afce96ef89a20d5bd8d6993521/code/Controllers/ContentController.php#L91)
- [`ViewableData::defineMethods()`](https://github.com/silverstripe/silverstripe-framework/blob/2c971eea832da926028a3944d2cee1f13aab2312/src/View/ViewableData.php#L175)
- [`CustomMethods::addMethodsFrom('failover')`](https://github.com/silverstripe/silverstripe-framework/blob/2c971eea832da926028a3944d2cee1f13aab2312/src/View/ViewableData.php#L238)
- [`CustomMethods::findMethodsFromExtension($virtualPage)`](https://github.com/silverstripe/silverstripe-framework/blob/2c971eea832da926028a3944d2cee1f13aab2312/src/Core/CustomMethods.php#L236)
- [`$virtualPage->allMethodNames(true)`](https://github.com/silverstripe/silverstripe-framework/blob/2c971eea832da926028a3944d2cee1f13aab2312/src/Core/CustomMethods.php#L205)

When the `VirtualPage` is rendered in a menu however, nothing ever calls `VirtualPage::allMethodNames()`, so it never gets a chance to define the “source” page’s extra methods.

It seems to me that `allMethodNames()` is designed to (or had its design changed to) only apply to _extensions_: it’s not called on object construction as part of `defineMethods()` like (I think) it used to be.

## The fix

There are two possible fixes that I can see:

1) Add a `VirtualPage::hasMethod()` and hope this is the only place the `allMethodNames()` shortcoming crops up
2) Add a `VirtualPage::defineMethods()` which explicitly defines the extra methods when the page is instantiated

I’m going to roll with number 2 and post a PR containing the solution I came up with when debugging this.

kinglozzer:
Needs test coverage, just submitting for feedback

### DIFF
--- a/code/Model/VirtualPage.php
+++ b/code/Model/VirtualPage.php
@@ -505,18 +505,43 @@ class VirtualPage extends Page
         return parent::castingHelper($field);
     }
 
-    /**
-     * {@inheritdoc}
-     */
+    public function defineMethods()
+    {
+        parent::defineMethods();
+
+        $methods = array_diff($this->getVirtualMethodNames(true), parent::allMethodNames(true));
+        $methodInfo = [
+            'property' => 'CopyContentFrom',
+            'index' => null,
+            'callSetOwnerFirst' => false
+        ];
+
+        $class = static::class;
+        $newMethods = array_fill_keys($methods, $methodInfo);
+        if (isset(self::$extra_methods[$class])) {
+            self::$extra_methods[$class] = array_merge(self::$extra_methods[$class], $newMethods);
+        } else {
+            self::$extra_methods[$class] = $newMethods;
+        }
+    }
+
     public function allMethodNames($custom = false)
     {
-        $methods = parent::allMethodNames($custom);
+        return array_merge(parent::allMethodNames($custom), $this->getVirtualMethodNames($custom));
+    }
 
-        if ($copy = $this->CopyContentFrom()) {
-            $methods = array_merge($methods, $copy->allMethodNames($custom));
+    /**
+     * @param bool $custom
+     * @return array
+     */
+    protected function getVirtualMethodNames($custom = false)
+    {
+        $copy = $this->CopyContentFrom();
+        if ($copy && $copy->exists()) {
+            return $copy->allMethodNames($custom);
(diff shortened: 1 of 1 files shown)


unclecheese:
@kinglozzer Are you able to change this to a draft PR? Would help keep our pipeline clean as we head toward feature freeze on 4.4. If not, that's ok.

ScopeyNZ:
Looks like a good change to me. Add some test coverage and I think it's ready to go.

@unclecheese this is a patch PR - I don't think it's bound by the 4.4 feature freeze?

unclecheese:
No, it's not. But if it's just about collecting feedback it should be a draft PR.

kinglozzer:
Weird, you can’t switch it to draft after creating it... I’ll re-open once I’ve added some tests 😉

kinglozzer:
Found a simpler solution than messing around with trying to override the behaviour of the `CustomMethods` trait: https://github.com/silverstripe/silverstripe-cms/pull/2411

PR #2421, opened 2019-05-08
FIX Only add page font icon classes if there's no icon image defined

ScopeyNZ:
Fixes silverstripe/silverstripe-admin#872


robbieaverill:
Tested with custom page types that use font icons, and with page types that use font image resource paths, both work nicely.

PR #2427, opened 2019-05-09
FIX Allow all types of content to be linked in dependent pages

scott1702:
This causes issues as content blocks can also be 'Dependent pages'. So it currently links to a page with the ID of the content block.


robbieaverill:
This also fixes the incompatibility with configurable "admin" route. Thanks.

PR #2431, opened 2019-05-16
BUGFIX Page Type listed in Page Type dropdown when $can_be_root is false

### LINKED ISSUE #2430: BUG Page Type listed in Page Type dropdown when $can_be_root is false
If a page is on the root level and page type has `private static $can_be_root = false`, it shouldn't be included as a dropdown option.

muskie9:
resolves #2430

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -2567,6 +2567,11 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         foreach ($classes as $class) {
             $instance = singleton($class);
 
+            //if the current page is root and the instance can't be, exclude
+            if (!$instance->config()->get('can_be_root') && $this->ParentID == 0) {
+                continue;
+            }
+
             // if the current page type is this the same as the class type always show the page type in the list
             if ($this->ClassName != $instance->ClassName) {
                 if ($instance instanceof HiddenClass) {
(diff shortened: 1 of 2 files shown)


code/Model/SiteTree.php
```diff
@@ -2567,6 +2567,11 @@ protected function getClassDropdown()
         foreach ($classes as $class) {
             $instance = singleton($class);
 
+            //if the current page is root and the instance can't be, exclude
+            if ($instance->config()->get('can_be_root') == false && $this->ParentID == 0) {
```
ScopeyNZ:
```suggestion
            if (!$instance->config()->get('can_be_root') && $this->ParentID === 0) {
```

muskie9: I'll try that. I couldn't seem to get tests running locally for whatever reason so I've been having to leverage travis, which isn't ideal for tracking down the issue.

ScopeyNZ: This suggestion is the same functionally, just making strict comparisons where the should be 🙂

muskie9: Still no luck with local test, so debugging the failure is proving to be difficult. At a glance does it look like my methodology of testing that the class is in the source array as a key?

ScopeyNZ: Seems right - I don't have time to dive into it though. Sometimes with travis you'll have to reproduce the install steps to get the same failure. This'll mean `git clone silverstripe/silverstripe-cms`, checking out your PR branch, and then running the steps in `.travis.yml`.

muskie9: No worries, I'll give that a try and see what I can come up with.

muskie9: @ScopeyNZ that's gotten me a lot closer on local, thnx for the pointer 👏

muskie9:
The failure was ultimately permissions for the user rather than the new config check. I should've noticed the other `$this->loginWithPermission('ADMIN');` calls.

michalkleiner:
Merge this in? :)

PR #2441, opened 2019-05-28
FIX: Make test comparison less brittle in MigrateSiteTreeLinkingTaskTest

sminnee:
Comparing every single field is unnecessary and brittle, only the IDs
need to be compared.

Notably this tripped over a potential bug fix in
https://github.com/silverstripe/silverstripe-framework/pull/8591
but the change should be incorporated regardless.


sminnee:
@robbieaverill @ScopeyNZ pinging you two because you're always quick-fire on the PR review :P

PR #2442, opened 2019-05-30
NEW Add updateHintsCacheKey extension point to fix invalid caching

Cheddam:
If a module augments the allowed pagetypes based on external conditions, the Hints Cache Key will not cover these conditions and may cause incorrect output. An example of this is Subsites, which allows each Subsite to have a different set of allowed pagetypes.

Since this is technically new API, it'd normally require a new minor release, however this extension point is being introduced purely in pursuit of fixing a bug, so I'd like to get it in as a patch update for the last few minor releases.

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -2272,7 +2272,11 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
      */
     protected function generateHintsCacheKey($memberID)
     {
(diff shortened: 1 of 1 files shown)


robbieaverill:
>Since this is technically new API, it'd normally require a new minor release, however this extension point is being introduced purely in pursuit of fixing a bug, so I'd like to get it in as a patch update for the last few minor releases.

That's fine, you can introduce new APIs to fix bugs in patch changes with semantic versioning

PR #2444, opened 2019-05-31
FIX Keep New Page type selection if still allowed when changing parent

Cheddam:
A UX wonk that has annoyed me since I started using SilverStripe:

When creating a new page, the most prominent part of the UI is the pagetype selection field, so I frequently act on it first. Almost as often, I realise that I need to change the parent page to 'Top level', since I've come from editing another page and that's the parent by default. Now ready to create the page, I hit submit - and lo-and-behold, I've ended up with a page of the wrong type because changing the parent needlessly reset the pagetype selection.

This PR changes the behaviour of the parent change logic so that if the selected pagetype is still available under the newly selected parent, it _stays_ selected.

(This might be worth backporting to SS3 - I'm fairly confident the code won't have changed.)


Cheddam:
In that case, the previous behaviour is kept - either it switches to the default pagetype, or to the first allowed pagetype.

Issue #2318, opened 2018-11-08
Add  new page here appears off-screen


NightJar:
[silverstripe add new page here.zip](https://github.com/silverstripe/silverstripe-cms/files/2563808/silverstripe.add.new.page.here.zip)
 - contains a `.mov` of issue

When editing in the CMS Pages section, right clicking (opening a context menu) on a page in the site tree widget brings brings up an option to "Add new page here". Hovering on this menu item open a sub-menu that then renders mostly off-screen, making it impossible to effectively select a page type to add.

Affected versions (at least, may be more):
`4.2.1`
`4.3.0-rc1`

![image](https://user-images.githubusercontent.com/778003/48231332-9da23e00-e412-11e8-8c06-daee3a17ef3b.png)

adrexia:
Also, This is happening (again) when there are a lot of pagetypes: 

![sitetree](https://user-images.githubusercontent.com/984753/48234147-363dbb80-e41d-11e8-8e9d-0bb68461a6d5.png)

This used to be an issue in Silverstripe 3 until we spread the pagetypes over 3 columns.  I'm not sure if this should be a seperate issue?

NightJar:
See https://github.com/silverstripe/silverstripe-cms/issues/811 for original context on the return of the issue @adrexia reports.

adrexia:
And this was the original fix: https://github.com/silverstripe/silverstripe-framework/commit/e6b06cade45a13d3c26ab9178037b3b791ec2020

adrexia:
Ok, this is because of the addition of bootstrap, where col-3 is used as a column in a grid. Whereas here is means 'span 3 columns'. This should be fixable by changing the class name used here to something less likely to conflict with bootstrap's naming.

robbieaverill:
Nice find! Fancy a PR? ;-)

NightJar:
Still an issue, testing for 4.4.0
![image](https://user-images.githubusercontent.com/778003/57202299-e15bb600-6ff7-11e9-8784-0e305114b60b.png)

lexakami:
Still an issue in `4.4.1` as well

bergice:
Unable to replicate on Chrome/Firefox for:

- `4.2.1`
- `4.3.0-rc1`
- `4.4.0`
- `4.4.1`

Can someone else confirm?

NightJar:
Checking now.

[...]

PR #2449, opened 2019-06-19
FIX: Change sitetree new page column class naming to avoid conflicts with bootstrap

lexakami:
Fix for #2318. 

Reliant on https://github.com/silverstripe/silverstripe-admin/pull/899

Currently when you add a page via the SiteTree widget, if there are too many pages it will overflow outside the viewport. 

image


ScopeyNZ:
Hmmn. Was this col class actually used anywhere?

ScopeyNZ:
Oh I missed the dependant PR 🤦‍♂

PR #2458, opened 2019-07-16
FIX display recognisable names for dependent content with no title

### LINKED ISSUE #2428: Dependent pages can have empty titles
Following on from https://github.com/silverstripe/silverstripe-cms/pull/2427. It's a fairly common pattern for CMS users to not add titles to things like content blocks.
Because of this, any blocks which link to a page and don't have titles appear as empty rows on the dependent pages table.
image

One way to fix this would be to use the singular_name and ID where a title doesn't exist:
image

But there are other formats that this could be in e.g. always having the type of object but using ID where a title doesn't exist:
```
Home | Home Page | Click here to edit
#123 | Content block | Click here to edit
```

Keen to get feedback on a preferred format. /cc @silverstripeux

### Solution
- Use the singular name, i.e 'Untitled block'

NightJar:
Content blocks and other DataObjects that contain HTMLText fields are
also used in the 'Dependent pages' report for each page in the CMS,
however these objects may have neither a Title field, nor a name to
describe them. Instead of showing an empty field in the Title column for
this table, we can instead show "Untitled " with the localised singular
name appended.

Resolves #2428

### DIFF
--- a/tests/php/Model/SiteTreeBacklinksTest.php
+++ b/tests/php/Model/SiteTreeBacklinksTest.php
@@ -6,6 +6,7 @@ use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTP;
 use SilverStripe\Dev\SapphireTest;
+use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Versioned\Versioned;
 
 /**
@@ -21,6 +22,10 @@ class SiteTreeBacklinksTest extends SapphireTest
         ],
     ];
 
+    protected static $extra_dataobjects = [
+        SiteTreeBacklinksTestContentObject::class,
+    ];
+
     public function setUp()
     {
         parent::setUp();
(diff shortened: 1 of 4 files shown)


ScopeyNZ:
Hey @NightJar . Failing test is due to IDs in the fixture. Fixtures shouldn't specify IDs so you should just remove this line:

https://github.com/silverstripe/silverstripe-cms/blob/6cb20778786e92ba8fa606ab866cffc19175c064/tests/php/Model/SiteTreeBacklinksTest.yml#L3

NightJar:
Updated, thanks @ScopeyNZ :)

tests/php/Model/SiteTreeBacklinksTest.php
```diff
@@ -251,4 +264,24 @@ public function testLinkTrackingOnExtraContentFields()
         // assert backlink to page 2 no longer exists
         $this->assertNotContains($page2->ID, $page1->BackLinkTracking()->column('ID'), 'Assert backlink to page 2 has been removed');
     }
+
+    public function testLinkTrackingWithUntitledObjectsDisplaysAReadableIdentifier()
+    {
+        $referencingObject = new SiteTreeBacklinksTestContentObject();
+        $referencingObject->Content = '<p><a href="[sitetree_link,id=2]">Page 2</a></p>';
```
ScopeyNZ:
```suggestion
        $page2ID = $this->objFromFixture('Page', 'page2')->ID;
        $referencingObject->Content = '<p><a href="[sitetree_link,id=' . $page2ID . ']">Page 2</a></p>';
```

NightJar: oh duh, of course! Same fix, different context - completely glossed over that! :roll_eyes:

NightJar: I've pushed the fix using a variable to avoid fetching the fixture twice, plus it allowed a squash :)
Thanks!

PR #2463, opened 2019-08-02
Changing to protected to match the rest of the changes

ninty9notout:
The `onBeforeWrite` and `onAfterWrite` methods have been changed to `protected` in `DataObject` and `SiteTree` but have been missed out from `RedirectorPage` so it errors out.

### DIFF
--- a/code/Model/RedirectorPage.php
+++ b/code/Model/RedirectorPage.php
@@ -136,7 +136,7 @@ class RedirectorPage extends Page
         }
     }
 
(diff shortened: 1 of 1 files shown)


ninty9notout:
Agree with you as well. I suppose it's just keeping consistent with base classes.

robbieaverill:
I think it's likely to be shown in dev/build which would prevent it from being an issue in production during an upgrade

Issue #2154, opened 2018-04-24
Prevent deletion of homepage


chillu:
# Overview

It's far too easy to delete the homepage of a website, rendering it effectively useless to visitors. This is a common feedback from our operations team.

# Acceptance Criteria

- [x] As a CMS author, I get an extra confirmation dialog when unpublishing or archiving the homepage
- [x] The system prevents me from renaming the URL segment of the homepage
- [x] The system prevents me from moving the homepage into another parent
- [x] I can use the logic with subsites and fluent (multiple homepages)

# Notes

 * Sometimes there's good reasons to delete the homepage, e.g. in development mode where you might want to reset its settings, or change a page type. Ideally we're not preventing this from happening completely.
 * Should use `RootURLController::$default_homepage_link`
 * Related to [Add RootURLController::is_homepage()](https://github.com/silverstripe/silverstripe-cms/issues/2051) RFC

# Review
- [ ] https://github.com/silverstripe/silverstripe-cms/pull/2470
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/937
- [x] Publish https://www.npmjs.com/package/@silverstripe/reactstrap-confirm

robbieaverill:
Related: https://github.com/silverstripe/cwp/issues/69

chillu:
Discussed this with @lukereative @maxime-rainville @newleeland and we think that it's best to keep this minimal and deny the ability to delete/unpublish/move the homepage. So if you hit one of these actions as an author or CMS admin, you'll get a nicely styled error response and the actions aren't triggered. This ensures that it also applies to batch actions.

Ideally we'd have a confirmation warning dialog, but that's too much investment for this feature. @clarkepaul Are you OK with this approach?

clarkepaul:
Yup all good.
I would think that ensuring the Homepage remains on the top-level of the tree hierarchy would be a sufficient check rather than ensuring it isn't moved. Otherwise you might want to check if a page other than the Homepage is moved above the Homepage as well (if you are going to restrict moving the Homepage itself).

frankmullenger:
One of our large projects had this same problem, it wasn't only the home page but several top level pages (of different types) that needed to remain published. We ended up implementing an "Un-unpublishable" checkbox in settings which users with certain permissions could check to remove UI elements that would normally allow a content author to archive/unpublish/change URL segment/move etc. As well as adding URL environment checks for those pages so that we could be alerted if they happen to disappear.

[...]

Issue #2051, opened 2017-12-07
RFC: Add "RootURLController::is_home_page(SiteTree $page)" function.


silbinarywolf:
**Why?**
Currently to support Multisites, I need to write code like this to ensure a page is actually a home page:
```
protected function isHomePage(\SiteTree $page)
{
    $parent = $page->Parent();
    return $page->URLSegment === 'home' && ((class_exists('Site') && $parent instanceof \Site) || !$parent->exists());
}
```

source:
https://github.com/silbinarywolf/silverstripe-oldman/blob/master/src/Cloudflare.php#L276

**Proposed solution:**
- Create a function with the signature: `RootURLController::is_home_page(SiteTree $page)`
- Uses existing `RootURLController::get_homepage_link` function and compares to the $page-Link(), which will return a boolean $result value.
- This $result value will be fed through an extension point `$this->extend('updateIsHomePage', &$result);` wherein modules like Multisites can modify whether it's actually on the home page or not.
- Write unit test to cover this function, it'll just use the extension point and set `$result = false`, meaning that no page will fit the criteria of home page.

tractorcow:
It seems like extensions modify the home page link, and that isHomePage() is done by comparing this urlsegment to the one being tested.

With that in mind, should we put the extension into `get_homepage_link()` instead? It could be implemented like this:


```php
if (!isset(self::$cached_homepage_link)) {
	$homepage = Config::inst()->get(self::class, 'default_homepage_link');
	static::singleton()->extend('updateHomepageLink', $homepage);
	self::$cached_homepage_link = $homepage;
}
return self::$cached_homepage_link;
```

silbinarywolf:
@tractorcow The conditions for whether something is a home page or not is currently as I understand it, is:
- ParentID == 0
- URLSegment == RootURLController::get_homepage_link()

This doesn't work out for Multisites because a HomePage will have:
- ParentID != 0 (ie. could be anything, but definitely not 0 as "Site" DataObject's are at the top of the SiteTree)
- URLSegment == RootURLController::get_homepage_link()

This is why I'm suggesting a function that checks if a SiteTree instance is a home page.

EDIT: In SS4, I found a function that does what I want already. It just needs an extension point seemingly:
https://github.com/silverstripe/silverstripe-cms/blob/master/code/Controllers/RootURLController.php#L82

silbinarywolf:
[...]

PR #2470, opened 2019-08-16
Add warnings when trying to remove the homepage

bergice:
Warnings are triggered when:
* Changing the URL Segment
* Unpublishing
* Unpublishing & Archiving

Also add `SiteTree->isHomePage()`

Resolves https://github.com/silverstripe/silverstripe-cms/issues/2154
Resolves https://github.com/silverstripe/silverstripe-cms/issues/2051


client/src/legacy/CMSMain.EditForm.js
```diff
@@ -476,4 +477,33 @@ $.entwine('ss', function($){
...
+  $('.cms-edit-form .btn-toolbar #Form_EditForm_action_unpublish.homepage-warning,' +
+	'.cms-edit-form .btn-toolbar #Form_EditForm_action_archive.homepage-warning,' +
+	'#Form_EditForm_URLSegment_Holder.homepage-warning .btn.update').entwine({
+    onclick: async function (e) {
+      if (confirmed) {
+        return this._super(e);
+      }
+      e.stopPropagation();
+
+      if (await reactConfirm(['Warning: This page has the URL of ',<em>home</em>,'. ' +
```
robbieaverill: You'll need to make this text i18n compatible

code/Model/SiteTree.php
```diff
@@ -2009,7 +2009,8 @@ public function getCMSFields()
                 'SilverStripe\\CMS\\Controllers\\CMSMain.NEWPAGE',
                 'New {pagetype}',
                 array('pagetype' => $this->i18n_singular_name())
-            )));
+            )))
+            ->addExtraClass(($this->isHomePage()?'homepage-warning':''));
```
robbieaverill: Can you run a linter over this PR please?

code/Model/SiteTree.php
```diff
@@ -3259,4 +3260,9 @@ public function getAnchorsOnPage()
 
         return $anchors;
     }
+
+    public function isHomePage()
```
robbieaverill: This could use a doc block I reckon

robbieaverill: You can also use PHP 7 return types in 4.5.x =)

client/src/legacy/CMSMain.Tree.js
```diff
@@ -140,6 +141,46 @@ $.entwine('ss.tree', function($) {
...
+          'CMS.RemoveHomePageWarningMessage',
+          'Warning: This page has the URL of %s. ' +
+          'This is required by your website to  identify the home page. ' +
+          'Without this, visitors will not be able to view your home page.'
+        ),
+        "<em>home</em>"
+      );
+
+      const regex = /(.*)(<em>home<\/em>)(.*)/;
+      var match = regex.exec(message);
```
ScopeyNZ: Looks like you're injecting `home` into the translated string? You can do this using `i18n.inject`. Here's an example: https://github.com/dnadesign/silverstripe-elemental/blob/4/client/src/components/ElementActions/UnpublishAction.js#L32

code/Model/SiteTree.php
ScopeyNZ: This was resolved but it looks like the linter config might be wrong? There should be spaces around the ternary delimiters here. Additionally there are some extra brackets in there.

client/src/legacy/CMSMain.EditForm.js
```diff
@@ -476,4 +477,50 @@ $.entwine('ss', function($){
...
+      e.stopPropagation();
+
+      var message = i18n.sprintf(
+        i18n._t(
+          'CMS.RemoveHomePageWarningMessage',
+          'Warning: This page has the URL of %s. ' +
+          'This is required by your website to  identify the home page. ' +
+          'Without this, visitors will not be able to view your home page.'
+        ),
+        "<em>home</em>"
```
robbieaverill: You can use `i18n.inject()` to inject placeholders, instead of sprintf. I'm not sure why you need it here though, since the injection is hardcoded as home

client/src/legacy/CMSMain.Tree.js
```diff
@@ -140,6 +141,46 @@ $.entwine('ss.tree', function($) {
         }
       };
       return config;
+    },
+
+    canMove: async function(data) {
```
robbieaverill: Can you add a doc block please?

client/src/legacy/CMSMain.Tree.js
```diff
@@ -140,6 +141,46 @@ $.entwine('ss.tree', function($) {
...
+      }
+
+      var message = i18n.sprintf(
+        i18n._t(
+          'CMS.RemoveHomePageWarningMessage',
+          'Warning: This page has the URL of %s. ' +
+          'This is required by your website to  identify the home page. ' +
+          'Without this, visitors will not be able to view your home page.'
+        ),
+        "<em>home</em>"
```
robbieaverill: Same comment as the previous file - as Guy suggested you can use `inject` if the injection is dynamic, but here it looks static, so you may as well just make it part of the base translation string

code/Model/SiteTree.php
```diff
@@ -2862,9 +2863,10 @@ public function getTreeTitle()
         $children = $this->creatableChildPages();
         $flags = $this->getStatusFlags();
         $treeTitle = sprintf(
-            "<span class=\"jstree-pageicon page-icon %s class-%s\"></span><span class=\"item\" data-allowedchildren=\"%s\">%s</span>",
+            "<span class=\"jstree-pageicon page-icon %s class-%s%s\"></span><span class=\"item\" data-allowedchildren=\"%s\">%s</span>",
             $this->getIconClass(),
             Convert::raw2htmlid(static::class),
+            $this->isHomePage() ? " homepage" : "",
```
robbieaverill: Can you use single quotes here, since you've used them in the rest of your code?

code/Model/SiteTree.php
```diff
@@ -3259,4 +3261,14 @@ public function getAnchorsOnPage()
 
         return $anchors;
     }
+
+    /**
+     * Returns whether this is the home page or not
+     *
+     * @return bool
+     */
+    public function isHomePage() : bool
```
robbieaverill:
```suggestion
    public function isHomePage(): bool
```

package.json
```diff
@@ -32,6 +32,7 @@
   },
   "homepage": "https://github.com/silverstripe/silverstripe-cms#readme",
   "dependencies": {
+    "@silverstripe/reactstrap-confirm": "0.0.4",
```
robbieaverill: This can be updated to `~0.0.5` now

robbieaverill:
You can also add a `NEW` commit prefix, so Cow picks it up as an enhancement in the next release changelog =)

client/src/legacy/CMSMain.Tree.js
bergice: Here I am finding the `home` tag and splitting by it, then replacing it with a tag instead of a string and passing that as an array so that React allows the HTML formatting to stay intact.

I'll add a remark to make this more obvious.

client/src/legacy/CMSMain.Tree.js
```diff
@@ -140,6 +141,52 @@ $.entwine('ss.tree', function($) {
...
+          'CMS.RemoveHomePageWarningMessage',
+          'Warning: This page has the URL of {home}. ' +
+          'This is required by your website to  identify the home page. ' +
+          'Without this, visitors will not be able to view your home page.'
+        ),
+        { home: "<em>home</em>" }
+      );
+
+      // split by the home tag so we can inject it to React without escaping the HTML
+      const regex = /(.*)(<em>home<\/em>)(.*)/;
```
robbieaverill: It seems like overly complex logic just for some italics, to be honest. I'd rather see if @silverstripeux would compromise and say `This page has a URL of "home".` instead. Also, should we be saying "URL segment" rather than "URL", since it's both not a URL and doesn't match the UI label for the field that contains that value?

robbieaverill: You really need this PR: https://github.com/silverstripe/silverstripe-admin/pull/895

ScopeyNZ: 😰

newleeland: ^^ Yeah using quotations are fine but it should be single quote marks (') not double ("), as used in other parts of the CMS. 

We can use 'URL segment' over 'URL'

client/src/legacy/CMSMain.EditForm.js
```diff
@@ -476,4 +477,44 @@ $.entwine('ss', function($){
...
+        return this._super(e);
+      }
+      e.stopPropagation();
+
+      var message = i18n._t(
+		  'CMS.RemoveHomePageWarningMessage',
+		  'Warning: This page has the URL segment of \'home\'. ' +
+		  'This is required by your website to  identify the home page. ' +
+		  'Without this, visitors will not be able to view your home page.'
+	  );
```
robbieaverill: Indentation issues here

client/src/legacy/CMSMain.EditForm.js
```diff
@@ -476,4 +477,44 @@ $.entwine('ss', function($){
...
+    onclick: async function (e) {
+      if (confirmed) {
+        return this._super(e);
+      }
+      e.stopPropagation();
+
+      var message = i18n._t(
+		  'CMS.RemoveHomePageWarningMessage',
+		  'Warning: This page has the URL segment of \'home\'. ' +
+		  'This is required by your website to  identify the home page. ' +
```
robbieaverill:
```suggestion
		  'This is required by your website to identify the home page. ' +
```

client/src/legacy/CMSMain.EditForm.js
```diff
@@ -476,4 +477,44 @@ $.entwine('ss', function($){
...
+      	title: i18n._t(
+          'CMS.RemoveHomePageWarningTitle',
+          'Remove your home page?'
+        ),
+		confirmLabel: i18n._t(
+          'CMS.RemoveHomePageWarningLabel',
+          'Remove'
+        ),
+        confirmColor: 'danger'
+	  })) {
```
robbieaverill: Indentation issues here

client/src/legacy/CMSMain.Tree.js
```diff
@@ -140,6 +141,45 @@ $.entwine('ss.tree', function($) {
...
+      var oldParentId = data.rslt.op.data('id');
+      var newParentId = data.rslt.np.data('id');
+      if (oldParentId === newParentId) {
+        return true;
+      }
+
+      var message = i18n._t(
+        'CMS.RemoveHomePageWarningMessage',
+        'Warning: This page has the URL segment of \'home\'. ' +
+        'This is required by your website to  identify the home page. ' +
```
robbieaverill:
```suggestion
        'This is required by your website to identify the home page. ' +
```

code/Model/SiteTree.php
```diff
@@ -2374,7 +2375,7 @@ public function getCMSActions()
             $moreOptions->push(
                 FormAction::create('unpublish', _t(__CLASS__.'.BUTTONUNPUBLISH', 'Unpublish'), 'delete')
                     ->setDescription(_t(__CLASS__.'.BUTTONUNPUBLISHDESC', 'Remove this page from the published site'))
-                    ->addExtraClass('btn-secondary')
+                    ->addExtraClass('btn-secondary'.($this->isHomePage()?' homepage-warning':''))
```
robbieaverill:
```suggestion
                    ->addExtraClass('btn-secondary' . ($this->isHomePage() ? ' homepage-warning' : ''))
```

code/Model/SiteTree.php
```diff
@@ -2428,7 +2429,7 @@ public function getCMSActions()
                 : _t('SilverStripe\\CMS\\Controllers\\CMSMain.ARCHIVE', 'Archive');
             $moreOptions->push(
                 FormAction::create('archive', $title)
-                    ->addExtraClass('delete btn btn-secondary')
+                    ->addExtraClass('delete btn btn-secondary'.($this->isHomePage()?' homepage-warning':''))
```
robbieaverill:
```suggestion
                    ->addExtraClass('delete btn btn-secondary' . ($this->isHomePage() ? ' homepage-warning' : ''))
```

code/Model/SiteTree.php
```diff
@@ -2875,9 +2876,10 @@ public function getTreeTitle()
         $children = $this->creatableChildPages();
         $flags = $this->getStatusFlags();
         $treeTitle = sprintf(
-            "<span class=\"jstree-pageicon page-icon %s class-%s\"></span><span class=\"item\" data-allowedchildren=\"%s\">%s</span>",
+            '<span class="jstree-pageicon page-icon %s class-%s%s"></span><span class="item" data-allowedchildren="%s">%s</span>',
             $this->getIconClass(),
             Convert::raw2htmlid(static::class),
+            $this->isHomePage() ? " homepage" : "",
```
robbieaverill:
```suggestion
            $this->isHomePage() ? ' homepage' : '',
```

client/src/legacy/CMSMain.Tree.js
```diff
@@ -140,6 +141,45 @@ $.entwine('ss.tree', function($) {
...
+      // only display warning if we're moving to a new parent
+      var oldParentId = data.rslt.op.data('id');
+      var newParentId = data.rslt.np.data('id');
+      if (oldParentId === newParentId) {
+        return true;
+      }
+
+      var message = i18n._t(
+        'CMS.RemoveHomePageWarningMessage',
+        'Warning: This page has the URL segment of \'home\'. ' +
```
robbieaverill: Ok, so `home` is configurable via `RootURLController::get_homepage_link()` - for that reason we'll need to get this from PHP somehow and pass it as an argument to `i18n.inject()`.

Without adding irrelevant schema data to something in the page for this, or adding an XHR request just to check whether it's the home page, I'd probably suggest removing this part of the text as well so it just says "This page is the home page. Are you sure you want to delete it?" - I realise you've been through copy changes a bunch of times already so sorry 😆  but we can't hard code it in one place and make it dynamic in another

bergice:
> Looking good! There's some weird stuff in your JavaScript indentation, and there's some missing spaces around operators in the SiteTree changes. Our linting rules might be a bit more relaxed about things like `.($foo?'bar':'baz')` at the moment but we should still aim to space them correctly e.g. `. ($foo ? 'bar' : 'baz')`

Okay @robbieaverill , should be all fixed up now. Hopefully we can get this merged sometime this millenia. :P

client/src/legacy/CMSMain.EditForm.js
```diff
@@ -476,4 +477,43 @@ $.entwine('ss', function($){
...
+        title: i18n._t(
+          'CMS.RemoveHomePageWarningTitle',
+          'Remove your home page?'
+        ),
+	    confirmLabel: i18n._t(
+          'CMS.RemoveHomePageWarningLabel',
+          'Remove'
+        ),
+        confirmColor: 'danger'
+	  })) {
```
robbieaverill: Your indentation is still pretty whack - have you done a thorough self review on this PR?

bergice: Should be fixed now. :)

bergice:
@robbieaverill 

> * You can modify the URL segment and save the page without pressing "OK" in the segment field, and there is no warning but the segment is saved

I'll change this.

> * You don't actually have to change the URL segment for it to warn you - simply click Edit then OK (you can check for a `changed` class on the field to see if it's changed)

I'll change this.

> * If you change the URL segment to something like `homeyhomehome` and save, then change it back to `home` you get the warning

I'll change this so we don't show warnings when changing the URL segment to the root URL segment.

> * If you change the URL segment from `home` to `homeyhomehome` and publish the page, then access the root of your site, you still have a home page, it redirects to the new URL segment
> * I can put my home page underneath the About Us page (as a child) and it still resolves correctly from the base URL
> It seems to me that the checks on changing the URL segment aren't required, and that the check should only be to prevent the page from being archived. Thoughts?

It does, until you do a flush, which then creates a new home page (and presumably clears some cache for the URL segment of the homepage).

robbieaverill:
>It does, until you do a flush, which then creates a new home page (and presumably clears some cache for the URL segment of the homepage).

Oh you mean on dev/build? `requireDefaultRecords()` may create one... we can use `isHomePage()` which you've added to check in that function to prevent this from happening.

bergice:
> > It does, until you do a flush, which then creates a new home page (and presumably clears some cache for the URL segment of the homepage).
> 
> Oh you mean on dev/build? `requireDefaultRecords()` may create one... we can use `isHomePage()` which you've added to check in that function to prevent this from happening.

It already does.

robbieaverill:
We probably need to tighten up that API then. Non-blocking for this PR though

PR #2471, opened 2019-08-22
FIX Disable caching on page previews (CMSPreviews or stage=Stage)

Cheddam:
A replacement PR for the [closed one on framework](https://github.com/silverstripe/silverstripe-framework/pull/8805), which explicitly disables caching through headers when a browser requests a page via the CMS Preview or in draft state.

Resolves [silverstripe/silverstripe-framework#8804](https://github.com/silverstripe/silverstripe-framework/issues/8804).

### DIFF
--- a/code/Controllers/ContentController.php
+++ b/code/Controllers/ContentController.php
@@ -8,6 +8,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
+use SilverStripe\Control\Middleware\HTTPCacheControlMiddleware;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injector;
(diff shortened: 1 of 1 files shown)


code/Controllers/ContentController.php
```diff
@@ -139,6 +140,14 @@ protected function init()
     {
         parent::init();
 
+        // In the CMS Preview or draft contexts, we never want to cache page output.
+        if (
+            $this->getRequest()->getVar('CMSPreview') === "1"
+            || $this->getRequest()->getVar("stage") === "Stage"
```
robbieaverill: Any reason for the mixture of quote types? We usually stick to single unless we need double.

You can also use `Versioned::DRAFT` instead of `"Stage"` if you'd like to use a constant.

Cheddam: Tidied, sorry! Hopefully nabbed the PSR-2 failure too (FWIW I was following PSR-12, didn't expect it to be different).

PR #2475, opened 2019-09-04
DOCS PHP5 no longer supported - change to PHP

christopherdarling:


patricknelson:
This didn't make it into `master` yet, which is usually bleeding edge. I was wondering why `4` was updated but `master` still called out ancient PHP 5:

image

robbieaverill:
Time for a merge up!

PR #2476, opened 2019-09-06
Keep a more shared understanding of Archived

adrhumphreys:
I feel as though archived should be the same in this context as others rather than having different definitions of what archived means throughout the code


ScopeyNZ:
Do you want to re-target this against `4.3` or `4.4` as a bugfix?

adrhumphreys:
@ScopeyNZ yeah you're right, have rebased 👍

tractorcow:
Looks good to me. Not sure what the old logic was trying to do.

Issue #2438, opened 2019-05-23
OldPageRedirector should use 'static::' for better extensibility.


patricknelson:
OldPageRedirector should use `static::` not `self::` for better extensibility. Currently if you attempt to override this class via `Injector` for the purpose of modifying the static `::find_old_page()` method, you cannot actually change anything since the method `->onBeforeHTTPError404()` calls it via `self::find_old_page()`, meaning the injected class is still bypassed. This then requires you to duplicate code in BOTH methods instead of just the one method.

See for reference: https://github.com/silverstripe/silverstripe-cms/blob/master/code/Controllers/OldPageRedirector.php#L32

robbieaverill:
Make sense, would you like to make a pull request to update it?

robbieaverill:
Fixed in 4.4.x with $2478 - thanks @elabuwa!

patricknelson:
Awesome @elabuwa, thanks! Sorry @robbieaverill for not addressing this myself.

robbieaverill:
Don’t be sorry, it’s open source! 🙂

mikey-harveycameron:
This fix has been lost in #2523 and needs to be reinstated

patricknelson:
Yikes, thanks for catching this @mikey-harveycameron. A [quick blame](https://github.com/silverstripe/silverstripe-cms/blame/4/code/Controllers/OldPageRedirector.php) on that file (at least in `4`) shows this regression happening in https://github.com/silverstripe/silverstripe-cms/commit/ed41f157996162366d218eb2208f05bc4c77efd1 via PR #2548 (issue #2522).

Good example of why they're now requiring unit tests to validate changes now (like the ones that were made to address _this_ ticket in #2478). To catch regressions like these. 😄  

@michalkleiner, I'm not sure if you're the main one triaging tickets now, but:  I propose reopening this ticket again. Hopefully the new/fixing PR will include the appropriate validating unit test.

patricknelson:
cc @maxime-rainville as well. Hopefully this can at least be fixed in 4 since it appears to break functionality that was previously fixed in 4. 😬

PR #2478, opened 2019-09-10
Change self:: to static:: for better extensibility

elabuwa:
Addresses :: https://github.com/silverstripe/silverstripe-cms/issues/2438

### DIFF
--- a/code/Controllers/OldPageRedirector.php
+++ b/code/Controllers/OldPageRedirector.php
@@ -29,7 +29,7 @@ class OldPageRedirector extends Extension
         $getvars = $request->getVars();
         unset($getvars['url']);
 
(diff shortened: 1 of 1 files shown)


code/Controllers/OldPageRedirector.php
```diff
@@ -29,7 +29,7 @@ public function onBeforeHTTPError404($request)
         $getvars = $request->getVars();
         unset($getvars['url']);
 
-        $page = self::find_old_page($params);
+        $page = static::find_old_page($params);
```
robbieaverill: Can you explain how this will help? I can't see why this would make any difference

robbieaverill: Nevermind, I'm being dumb

PR #2488, opened 2019-10-15
FEATURE add config var for namespace mapping page->control

andrewandante:
Creates a config variable where you can define a mapping between your Pages namespace and your Controllers namespace.

A common desired pattern is to keep things properly organised by having `App\Pages\MyPageType` and `App\Control\MyPageTypeController` - but this isn't currently caught by the ✨ "magic" ✨ . This allows for a global setting, rather than the one-off afforded by `$controller_name`.

Opens the door for adding a default map in 5 - maybe auto-substituting Page => Control, but that's another debate to have.

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -215,6 +215,22 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      */
     private static $controller_name = null;
 
+    /**
+     * You can define the a map of Page namespaces to Controller namespaces here
+     * This will apply after the magic of appending Controller, and in order
+     * Must be applied to SiteTree config e.g.
+     *
+     * SilverStripe\CMS\Model\SiteTree:
+     *   namespace_map:
+     *     "App\Pages": "App\Control"
+     *
+     * Will map App\Pages\MyPage to App\Control\MyPageController
+     *
+     * @config
+     * @var string
+     */
+    private static $namespace_map = null;
+
     private static $db = [
         "URLSegment" => "Varchar(255)",
         "Title" => "Varchar(255)",
@@ -2989,6 +3005,8 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             return $controller;
         }
 
+        $namespaceMap = Config::inst()->get(SiteTree::class, 'namespace_mapping');
+
         //default controller for SiteTree objects
         $controller = ContentController::class;
 
@@ -3002,8 +3020,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             }
             // If we have a class of "{$ClassName}Controller" then we found our controller
             if (class_exists($candidate = sprintf('%sController', $class))) {
-                $controller = $candidate;
-                break;
+                return $candidate;
             } elseif (class_exists($candidate = sprintf('%s_Controller', $class))) {
                 // Support the legacy underscored filename, but raise a deprecation notice
                 Deprecation::notice(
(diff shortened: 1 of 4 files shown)


code/Model/SiteTree.php
```diff
@@ -2959,17 +2977,28 @@ public function getControllerName()
...
                 Deprecation::notice(
                     '5.0',
                     'Underscored controller class names are deprecated. Use "MyController" instead of "My_Controller".',
                     Deprecation::SCOPE_GLOBAL
                 );
-                $controller = $candidate;
-                break;
+                return $candidate;
+            } elseif ($namespaceMap !== null) {
+                foreach ($namespaceMap as $pageNamespace => $controllerNamespace) {
```
michalkleiner: Add a check `$namespaceMap` is an associative array?

andrewandante: Could do - it's likely to just fail on the `strpos` check though, which I think is acceptable. The config variable is called `namespace_mapping` after all, it does indicate an asssociative array should be set

michalkleiner: It wouldn't get to the `strpos` check if it wasn't an array, failing with `Invalid argument supplied for foreach()`

andrewandante: Ah well if it's not an `array` at all, yeah - I thought you meant it if it was a sequential array.

So basically you are suggesting:

```diff
- } elseif ($namespaceMap !== null) {
+ } elseif (is_array($namespaceMap) {
```

is that right?

michalkleiner: Initially, I meant to check for an actual associative array as it's a map, but checking for an array as a minimum is a good-enough middle ground.

code/Model/SiteTree.php
```diff
@@ -2959,17 +2977,28 @@ public function getControllerName()
...
                     '5.0',
                     'Underscored controller class names are deprecated. Use "MyController" instead of "My_Controller".',
                     Deprecation::SCOPE_GLOBAL
                 );
-                $controller = $candidate;
-                break;
+                return $candidate;
+            } elseif (is_array($namespaceMap)) {
+                foreach ($namespaceMap as $pageNamespace => $controllerNamespace) {
+                    if (strpos($class, $pageNamespace) === false) {
```
GuySartorelli, line 2991:
```suggestion
                    if (strpos($class, $pageNamespace) !== 0) {
```
Bit of an edge case but we should probably make sure the mapping namespace isn't actually somewhere in the middle of the class name.

tests/php/Model/SiteTreeTest.php
```diff
@@ -1622,6 +1624,13 @@ public function testGetControllerNameFromConfig()
         Config::inst()->update(Page::class, 'controller_name', 'This\\Is\\A\\New\\Controller');
         $class = new Page;
         $this->assertSame('This\\Is\\A\\New\\Controller', $class->getControllerName());
+
+        Config::inst()->update(SiteTree::class, 'namespace_mapping', [
+            'SilverStripe\\CMS\\Tests\\Page' => 'SilverStripe\\CMS\\Tests\\Controllers',
+        ]);
+
+        $namespacedSiteTree = new SiteTreeTest_NamespaceMapTest();
+        $this->assertSame(SiteTreeTest_NamespaceMapTestController::class, $namespacedSiteTree->getControllerName());
```
GuySartorelli, line 1633: This test should really be in a new method.

GuySartorelli (changes requested):
Couple of small changes to make, and some conflicts to resolve.
@andrewandante I know this has been sitting here for a long time without any action but if you're willing to make these changes I'd be happy to help get this across the line.

andrewandante:
@GuySartorelli changes made and rebased - I seem to have a couple of fixes in there for some other tests, I don't remember what that is about but hopefully they are useful changes?? I can remove them if needed. I honestly don't remember why they are there 😅

GuySartorelli:
@andrewandante I had intended to give those a proper look and see what is going on with them but I don't think I'll have time to.😅If you could remove those unrelated test changes that would be great, then I can merge on the merits of the new functionality and its related tests.

andrewandante:
@GuySartorelli consider them removed - if the tests break I'll add them back 😉

GuySartorelli (approved):
Nice work!
Keeps the default behaviour, but gives some very flexible new configuration which allow developers to use naming conventions that may be more appropriate than the defaults.

PR #2492, opened 2019-10-29
Negotiator amends: Make it possible to disable Preview links in the CMS

oilee80:
Existing functionality is limited to `RedirectorPage`

Another option would be to use the HiddenClass methodology

### DIFF
--- a/code/Controllers/SilverStripeNavigatorItem_LiveLink.php
+++ b/code/Controllers/SilverStripeNavigatorItem_LiveLink.php
@@ -58,9 +58,18 @@ class SilverStripeNavigatorItem_LiveLink extends SilverStripeNavigatorItem
             $record->hasExtension(Versioned::class)
             && $record->hasStages()
             && $this->getLivePage()
-            // Don't follow redirects in preview, they break the CMS editing form
-            && !($this->record instanceof RedirectorPage)
+            && $this->showLiveLink()
         );
+    }
+
+    public function showLiveLink()
+    {
+        try {
+            return $this->record->config()->get('show_live_link');
+        } catch (\BadMethodCallException $e) {
+            // Not using `config()` or similar
+            return false;
+        }
     }
 
     public function isActive()
--- a/code/Controllers/SilverStripeNavigatorItem_LiveLink.php
+++ b/code/Controllers/SilverStripeNavigatorItem_LiveLink.php
@@ -3,6 +3,7 @@ namespace SilverStripe\CMS\Controllers;
 
 use SilverStripe\CMS\Model\RedirectorPage;
 use SilverStripe\Control\Controller;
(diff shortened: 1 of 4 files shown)


code/Controllers/SilverStripeNavigatorItem_LiveLink.php
```diff
@@ -58,11 +58,20 @@ public function canView($member = null)
...
-            // Don't follow redirects in preview, they break the CMS editing form
-            && !($this->record instanceof RedirectorPage)
+            && $this->showLiveLink()
         );
     }
 
+    public function showLiveLink()
+    {
+        try {
+            return $this->record->config()->get('show_live_link');
```
andrewandante: Maybe cast this to `(bool)`, in case it's `null` or something odd?

code/Model/SiteTree.php
```diff
@@ -137,6 +137,10 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      */
     protected static $_allowedChildren = array();
 
+    private static $show_stage_link = true;
+
+    private static $show_live_link = true;
```
andrewandante: In this base class, could you add a comment that explains what these do in the practical sense? Something as simple as "set to false to disable the CMS preview for this class"

code/Controllers/SilverStripeNavigatorItem_LiveLink.php
ScopeyNZ: For more flexibility with configuration you could just use `Config::inst()->get(get_class($this->record), 'show_live_link')` instead. Then you can remove the `try...catch`.

code/Controllers/SilverStripeNavigatorItem_LiveLink.php
```diff
@@ -58,11 +58,20 @@ public function canView($member = null)
             $record->hasExtension(Versioned::class)
             && $record->hasStages()
             && $this->getLivePage()
-            // Don't follow redirects in preview, they break the CMS editing form
-            && !($this->record instanceof RedirectorPage)
+            && $this->showLiveLink()
```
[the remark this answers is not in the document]
oilee80: Moved 👍

code/Model/SiteTree.php
oilee80: Commented 👍

code/Controllers/SilverStripeNavigatorItem_LiveLink.php
oilee80: Updated 👍

andrewandante:
Lovely, very nice!

oilee80:
Heya @andrewandante @ScopeyNZ
Are either of you able to merge this?

andrewandante:
Unfortunately not mate or I would have already

PR #2493, opened 2019-11-03
Cache page icons

adrhumphreys:
We might be outliers but we've found that a good chunk of time (7ish%) on a request to the admin was being taken up by getting the page icons. These should only change when adding a new page which from my understanding would require a dev/build anyways so the cache wouldn't be any big issue

### DIFF
--- a/code/Controllers/LeftAndMainPageIconsExtension.php
+++ b/code/Controllers/LeftAndMainPageIconsExtension.php
@@ -2,32 +2,59 @@
 
 namespace SilverStripe\CMS\Controllers;
 
+use Psr\SimpleCache\CacheInterface;
+use Psr\SimpleCache\InvalidArgumentException;
+use ReflectionException;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Extension;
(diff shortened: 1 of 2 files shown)


maxime-rainville:
Looks cool. Sounds a tiny bit overkill to have a dedicated cache that will store only one key, but off the top of my head I can't see an obviously better way of doing it.

I'll test it locally to confirm it works, but at first glance it looks good.

adrhumphreys:
@maxime-rainville Thanks for the quick review :)
Have retargeted and added the flushable call

code/Controllers/LeftAndMainPageIconsExtension.php
```diff
@@ -2,32 +2,58 @@
 
 namespace SilverStripe\CMS\Controllers;
 
+use Psr\SimpleCache\CacheInterface;
+use Psr\SimpleCache\InvalidArgumentException;
+use ReflectionException;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Core\ClassInfo;
-use SilverStripe\Core\Config\Config;
```
maxime-rainville: Removing this use statements seems to be breaking the build.

adrhumphreys: Weird, was showing as unused in my IDE so I blindly removed it, have blindly added it back in

maxime-rainville:
This will be available in SS4.5 which should be release some time before the end of the year.

PR #2495, opened 2019-11-05
Upgrade dependencies

maxime-rainville:
# Parent issue
* https://github.com/silverstripeltd/product-issues/issues/48


package.json
```diff
@@ -55,13 +55,10 @@
   },
   "devDependencies": {
     "@silverstripe/eslint-config": "0.0.5",
-    "@silverstripe/webpack-config": "^1.0.0",
-    "babel-core": "^6.26.3",
+    "@silverstripe/webpack-config": "^1.3.0",
     "babel-jest": "^23.6.0",
-    "babel-loader": "^7.0.0",
     "babel-polyfill": "^6.26.0",
```
bergice: If you're removing these dependencies because they're unused, there might be more you can remove:

image

maxime-rainville: When in doubt, I'd rather keep them there otherwise, I'll have to spend time to make sure removing them doesn't actually break anything.

Because of the way our projects are set up it's not always clear what we can ditch.

PR #2500, opened 2019-11-18
Fixed 5 links in Contributing doc

rawbitt:
Updated Contributing link in intro. Fixed 4 broken (404) links.


maxime-rainville:
Thanks for taking the time to correct those links. That's awesome!

PR #2503, opened 2019-12-02
FIX: Prevent Fluent non-default locale 'Page not found' redirect loop

taoceanz:
Get url segments from $request->remaining() as ->getURL gets locale prefix which causes issues when attempting to 'get_old_page'.


taoceanz:
This actually doesn't work.

Issue #2522, opened 2020-02-05
OldPageRedirector doesn't work properly with multiple pages having same URLSegment


t3hn0:
We have multiple news list pages with same URLSegment and we noticed if we changed URLSegments of News (children of news listing pages) redirects didn't work.

I found a problem in OldPageRedirector::find_old_page. As it starts to look for old URL it doesn't necessarily start with ParentID=0 (root) as it's not requested and then it just picks first/random page (order by _Sort_ column) with the same URLSegment.

This lines:
```php
        if ($parent) {
            $pages = $pages->filter(array(
                'ParentID' => $parent->ID,
            ));
        }
```

should have a fallback to look on root level if no $parent is set:
```php
        if ($parent) {
            $pages = $pages->filter(array(
                'ParentID' => $parent->ID,
            ));
        } else {
            $pages = $pages->filter(array(
                'ParentID' => 0,
            ));
        }
```

I'm using SilverStripe v4.4.4, with Fluent 4.4. 

**Test case:**
1. Create page on root, Title "Test page" (URLSegment: "test-page") and publish it. In my case gets ID #19354, Sort=50
    _/test-page_
2. Create subpage on previously created page, Title "Goals" (URLSegment: "goals") and publish it. In my case gets ID #19355, Sort=1
    _/test-page/goals_
3. Create page on root, Title "Goals" (URLSegment: "goals") and publish it. In my case gets ID #19356, Sort=51
    _/goals_
4. Update URLSegment of "Goals" page (ID #19356) from previous line to "goals-new" and publish it
5. if you open _/goals_ you get redirected to _/test-page/goals/_ but it should redirect you to _/goals-new_

t3hn0:
Created pull request #2523  if bug is confirmed.

t3hn0:
May I ask will this be released in 4.5.x or branch?
I tried upgrading SS (with default installer) and looks like Recipe CMS is using different branch - I guess branch 4.5.

dhensby:
No, it will be in the next minor release (4.6) as the PR was opened against that branch (4) and merged into it

t3hn0:
I see. Do you have any info when 4.6 would be released?

PR #2523, opened 2020-02-05
Fixing issue #2522

t3hn0:
Pull request fixin issue #2522

### DIFF
--- a/code/Controllers/OldPageRedirector.php
+++ b/code/Controllers/OldPageRedirector.php
@@ -29,7 +29,10 @@ class OldPageRedirector extends Extension
         $getvars = $request->getVars();
         unset($getvars['url']);
 
-        $page = static::find_old_page($params);
+        $page = self::find_old_page($params, 0);
+        if (!$page) {
+            $page = self::find_old_page($params);
+        }
         $cleanPage = trim(Director::makeRelative($page), '/');
         if (!$cleanPage) {
             $cleanPage = Director::makeRelative(RootURLController::get_homepage_link());
@@ -59,7 +62,7 @@ class OldPageRedirector extends Extension
      */
     public static function find_old_page($params, $parent = null, $redirect = false)
     {
-        $parent = is_numeric($parent) ? SiteTree::get()->byID($parent) : $parent;
+        $parent = is_numeric($parent) && $parent > 0 ? SiteTree::get()->byID($parent) : $parent;
         $params = (array)$params;
         $URL = rawurlencode(array_shift($params));
         if (empty($URL)) {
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
Thanks for the PR. I think this would read cleaner if you combined this filter with the prior filter:

```php
$pages = SiteTree::get()->filter([
    'URLSegment' => $URL,
    'Parent' => $parent ? $parent->ID : 0,
]);
```

Then you can remove that whole `if...else` block.

In saying that, looks like you've got a valid test failure caused by this PR. I can't provide any insight on that sorry.

t3hn0:
Thanks I can fix that but somebody should check what the error is since it's the same logic and would still cause same error :)

ScopeyNZ:
You can still optimise further but combining those two `->filter` calls together.

The builds on `4` are fine, so this logic change _is_ causing a test failure. I'm sorry, but I don't have time to look at it now. Perhaps look at the assertion it's trying to make (if you haven't already) - it could give you clues as to why this patch might have affected it.

t3hn0:
Sorry just added quick fix...I made it propper now.

I reviewed the test and the problem is in this test
ModelAsControllerTest::testRedirectionForPreNestedurlsBookmarks

This test tests if URL` /newlevel3` should be found at `/newlevel1/newlevel2/newlevel3/`. My PR surely disables this with condition of ParentID=0 if no $parent is set.

I'm not sure how and if this is important to work..you'll decide. But other way of fixing reported error would be if $parent would somehow be transfered to OldPageRedirector::onBeforeHTTPError404() function in RequestHandler::httpError.

Cheddam:
Right, so the issue is that the existing logic was actually supporting cases where a page changed nesting level (e.g. went from being a top-level page to being a child page).

Perhaps a good trade-off would be to prioritise the new behaviour (searching with ParentID = 0), and then fall back to the old behaviour in order to find any matches that would have occurred before. It's two queries instead of one (although you could optimise the second query away), but seems worthwhile to get the best of both worlds.

Thoughts @ScopeyNZ?

ScopeyNZ:
I trust your judgement. Seems like the second query would be uncommon anyway.

t3hn0:
Hi @Cheddam 

I've reverted all the chages I made so far and updated code as you suggested and now all test pass.

Cheddam:
@t3hn0 Awesome! Before we give it a final tick and merge, could you squash the commits together and give the combined commit a formatted message like `BUG Prioritise same-level pages in OldPageRedirector`? See [the commit message docs](https://docs.silverstripe.org/en/4/contributing/code/#commit-messages) for more detail.

t3hn0:
@Cheddam great. Should I create new PR with single commit than and close this one?

Cheddam:
Up to you - you could rebase the current branch and force push it, though, to save creating a separate PR.

I've noticed that you've created this pull request from one of the "core branches" (`4`) rather than a branch specifically intended for this pull request. This would cause problems if we asked you to target this PR at a different branch. No action is needed, this is for your information.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request).

t3hn0:
@Cheddam thanks for the tip & for links to docs. I managed to squash commits and I updated commit message.

PR #2529, opened 2020-02-27
Update Travis config to include PHP 7.4 build (3.x)

Cheddam:


Cheddam:
I've raised an issue to cover getting this PR green and merged: silverstripe/silverstripe-framework#9421

dhensby:
Looks like there is some deprecated syntax to fix

dhensby:
fixed in framework 3.x branch

PR #2535, opened 2020-04-04
[ENHANCEMENT] Allow menu_title to be updated

mattclegg:
Currently, any class that extends from CMSMain has `Title` in the Breadcrumbs function set statically from; CMSPagesController::menu_title().

Using the function `$this->menu_title(CMSPagesController::class)` preserves the same functionality but also allows classes that extend from CMSMain to overwrite the returned value (if required).

Any feedback welcome.

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -1053,7 +1053,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
         if ($this->TreeIsFiltered()) {
             $items->push(new ArrayData([
-                'Title' => CMSPagesController::menu_title(),
+                'Title' => $this->menu_title(CMSPagesController::class),
                 'Link' => ($unlinked) ? false : $this->LinkPages()
             ]));
             $items->push(new ArrayData([
@@ -1071,7 +1071,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         $record = $this->currentPage();
         if (!$record) {
             $items->push(new ArrayData(array(
-                'Title' => CMSPagesController::menu_title(),
+                'Title' => $this->menu_title(CMSPagesController::class),
                 'Link' => ($unlinked) ? false : $this->LinkPages()
             )));
 


code/Controllers/CMSMain.php
```diff
@@ -1071,7 +1071,7 @@ public function Breadcrumbs($unlinked = false)
         $record = $this->currentPage();
         if (!$record) {
             $items->push(new ArrayData(array(
-                'Title' => CMSPagesController::menu_title(),
+                'Title' => $this->menu_title(CMSPagesController::class),
```
dhensby, line 1074: doesn't this involve turning the static method into a non-static method?

mattclegg: I think the parent method `menu_title` doesn't need to be a static method as it expects the $class parameter; https://github.com/silverstripe/silverstripe-admin/blob/1/code/LeftAndMain.php#L896-L899

Or are you referring to something else?

mattclegg:
> This seems like a slightly strange way to solve the problem - isn't this shoe-horning a fix into a static method?

Yeah, it's a strange way to fix it but I think it's also quite strange that CMSMain uses a static function from CMSPagesController when CMSPagesController extends CMSMain.

code/Controllers/CMSMain.php
dhensby: Whether it needs to be static or not is a bit irrelevant because the function is static and that can't change until SS5.

As the docs say the first param is for use only when calling directly from `LeftAndMain` (`@param string $class Optional class name if called on LeftAndMain directly`)

dhensby:
I don't see that this change is needed at all. This should work as is.

```php
class MyCMSAdmin extends LeftAndMain
{
    private static $menu_title = 'My Menu Title';
}
```

Otherwise you can use `CMSMenu` to add menu items with custom names too.

mattclegg:
The issue doesn't apply when you extend from `LeftAndMain` it's only when you extend from `CMSMain` (or any of its subclasses). I do appreciate it's an edge case. The issue I'm having is I want to customize the menu_title when it appears in the Breadcrumbs - without changing too much or the whole function from CMSMain.

My class looks like this;
```php
class MyCMSAdmin extends CMSPagesController
{
    private static $menu_title = 'My Menu Title';

    public static function menu_title($class = null, $localise = true)
    {
        if ($class==CMSPagesController::class) {
            return 'My Breadcrumb title';
        } else {
            return parent::menu_title($class, $localise);
        }
    }
}
```

With the patch in this MR applied the Breadcrumb is "My Breadcrumb title" & when it's not applied then the Breadcrumb is "Pages" because of the hardcoding to `CMSPagesController::menu_title()`.

Maybe the current wording is misleading as the `menu_title` function is used differently to a variable with the same name? Would it make more sense to introduce a static ```private static $breadcrumb_title``` that can default to `$menu_title` when not set?

dhensby:
If you're trying to change the title of `CMSPagesController` then change the config for that class. This isn't about changing the title for subclasses, it seems(?)

mattclegg:
I'd like to keep the title of `CMSPagesController` as "Pages" and only update it accordingly for MyCMSAdminCars, MyCMSAdminMotobikes, MyCMSAdminTrucks etc.

dhensby:
Sure - and your example controller won't do that, it's only updating the return value for `CMSPagesController`.

I also wouldn't have thought your CMS controllers should extend `CMSPagesController` - are they pages?

mattclegg:
Yeah, I'm aware the first example is only for `CMSPagesController`, I was just trying to show the initial problem with 1 controller.

Your right, they are pages. I just want to try to make them appear as the object they relate to and try to avoid using the word `Pages` in the UI.

Thanks for your help.

Issue #2358, opened 2019-01-05
CMS tree reloads after a Page is loaded


JorisDebonnet:
This issue is new in 4.3.0. 

Whenever I click a page in the CMS tree, the whole CMS is covered with a loading screen, which is okay I guess (although only the content part really needs to reload), but when it finishes - and this is new - the CMS tree on the left starts reloading on it own, too. No idea why; it shouldn't, and it doesn't in 4.2.3. 

![image](https://user-images.githubusercontent.com/3286264/50718727-f3056f00-1092-11e9-8461-1f76df1b1490.png)

Not sure what else to say, because I don't know what causes it. I tested it in two separate websites I maintain. In case it matters, I am also using silverstripe/blog on both. 

# Pull requests
* [x] https://github.com/silverstripe/silverstripe-cms/pull/2543

kinglozzer:
This is still an issue 😞

maxime-rainville:
This should fix it https://github.com/silverstripe/silverstripe-cms/pull/2543

PR #2543, opened 2020-04-24
BUG Prevent Treeview from always reloading

maxime-rainville:
`false` is save to local storage as a string which always triggers a treeview reload

# Parent issue
* https://github.com/silverstripe/silverstripe-cms/issues/2358


kinglozzer:
Tested and confirmed this fixes the issue. Should this be merged into `4.5` instead of `4` @maxime-rainville?

maxime-rainville:
I've retargeted to 4.5.

kinglozzer:
Tests are failing, but they’re clearly unrelated to this change, so let’s merge. Thanks @maxime-rainville

PR #2548, opened 2020-05-04
BUG Prioritise same-level pages in OldPageRedirector

### LINKED ISSUE #2522: OldPageRedirector doesn't work properly with multiple pages having same URLSegment

t3hn0:
Issue #2522 was merged into wrong branch (4) and we'd like to use this bugfix as soon as possible. 
I'm requesting new PR to make it available sooner for us.

Original description:
_Added option to pass integer ParentID=0 to OldPageRedirector::find_old_page to ensure search through nested pages from the root of SiteTree structure. Added new call of function and still offering fallback to old behaviour if first call returns false (fixed #2522)_

### DIFF
--- a/code/Controllers/OldPageRedirector.php
+++ b/code/Controllers/OldPageRedirector.php
@@ -29,7 +29,10 @@ class OldPageRedirector extends Extension
         $getvars = $request->getVars();
         unset($getvars['url']);
 
-        $page = static::find_old_page($params);
+        $page = self::find_old_page($params, 0);
+        if (!$page) {
+            $page = self::find_old_page($params);
+        }
         $cleanPage = trim(Director::makeRelative($page), '/');
         if (!$cleanPage) {
             $cleanPage = Director::makeRelative(RootURLController::get_homepage_link());
@@ -59,7 +62,7 @@ class OldPageRedirector extends Extension
      */
     public static function find_old_page($params, $parent = null, $redirect = false)
     {
-        $parent = is_numeric($parent) ? SiteTree::get()->byID($parent) : $parent;
+        $parent = is_numeric($parent) && $parent > 0 ? SiteTree::get()->byID($parent) : $parent;
         $params = (array)$params;
         $URL = rawurlencode(array_shift($params));
         if (empty($URL)) {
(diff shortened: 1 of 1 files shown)


robbieaverill (approved):

robbieaverill:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request)

code/Controllers/OldPageRedirector.php
```diff
@@ -29,7 +29,10 @@ public function onBeforeHTTPError404($request)
         $getvars = $request->getVars();
         unset($getvars['url']);
 
-        $page = static::find_old_page($params);
+        $page = self::find_old_page($params, 0);
```
patricknelson, line 32: Just FYI: It looks like this use of `self::` here instead of `static::` introduced a regression.  that was supposed to be addressed in #2478 (which itself was fixing #2438).

PR #2562, opened 2020-06-24
FIX: Incorrect method call in savetreenode (fixes #2561)

### LINKED ISSUE #2561: VirtualPage causes errors in savetreenode() when hierarchy is changed
> BadMethodCallException Core/CustomMethods.php in __call
Object->__call(): the method 'TreeTitle' does not exist on 'Page'

It comes from here: https://github.com/silverstripe/silverstripe-cms/blob/c58144ad6d624aabe10c5b7ce3ca5c487e8921ad/code/Controllers/CMSMain.php#L821

The problem is that the method is actually named `getTreeTitle()`: https://github.com/silverstripe/silverstripe-cms/blob/c58144ad6d624aabe10c5b7ce3ca5c487e8921ad/code/Model/SiteTree.php#L2879

Probably a regression from the last VirtualPage refactor

kinglozzer:
Fixes #2561

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -818,7 +818,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
             $virtualPages = VirtualPage::get()->filter("CopyContentFromID", $node->ID);
             foreach ($virtualPages as $virtualPage) {
                 $statusUpdates['modified'][$virtualPage->ID] = array(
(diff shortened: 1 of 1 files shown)


emteknetnz:
@kinglozzer thanks for the pull-request, I should get a chance to look closer at this in the next week or so

ScopeyNZ:
@kinglozzer Do you want to add a test as a proof?

emteknetnz:
@kinglozzer are you able to add a test to this PR?

Issue #2448, opened 2019-06-13
Unpublishing a Page does not show in history.


wilr:
Steps to reproduce in 4.4

1. Create a page
2. Publish page
3. Unpublish the page

When viewing the History tab you can only see 'Published' and 'Saved' no where is it clear that a page has been 'Unpublished' and when.

Before I submit a PR - is this intended behaviour? Was there a reason that this is not available?

![Screen Shot 2019-06-13 at 4 32 52 PM](https://user-images.githubusercontent.com/101629/59403974-f8f93c00-8df8-11e9-80e5-61c48d8a8941.png)

# Pull Requests
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2563
- [x] https://github.com/silverstripe/silverstripe-versioned-admin/pull/160
- [x] https://github.com/silverstripe/silverstripe-versioned/pull/279
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2575

wilr:
Right so I've found a few reasons why. It looks like this was a toggle in 3 which is no longer available. In my mind unpublished records (not archived) should still appear. Or should we have this toggle ala 4?

1. First step is `Versioned::augmentSQLVersioned` does not include unpublished records so would need to be updated. Line 611. 

```
-$query->addWhere(["\"{$baseTable}_Versions\".\"WasDeleted\"" => 0]);

+$query->addWhere(
            "\"{$baseTable}_Versions\".\"WasDeleted\" = 0 OR (".
                "\"{$baseTable}_Versions\".\"WasDeleted\" = 1 AND ".
                "\"{$baseTable}_Versions\".\"WasDraft\" = 0".
            ")"
        );
```

2. `DataObjectScaffolderExtension` needs to pass the WasDraft option.

3. Then all the UI needs to be updated (HistoryViewerVersionState) from `versioned-admin`

newleeland:
It would definitely be good to have Unpublished records in history. I cannot think of any UX reason why we wouldn't want it in History.

chillu:
Yeah I'd say that's unintended behaviour. Unpublishing should create a new version to record this activity, and hence should show as a new entry there.

[...]

PR #2563, opened 2020-07-03
Show the type of save that was performed for the record instead of just showing either Saved/Published.

bergice:
See https://github.com/silverstripe/silverstripe-cms/issues/2448


package.json
```diff
@@ -9,7 +9,7 @@
     "node": "^10.x"
   },
   "scripts": {
-    "build": "yarn && yarn test && NODE_ENV=production webpack -p --bail --progress",
+    "build": "yarn && NODE_ENV=production webpack -p --bail --progress",
```
robbieaverill, line 12: Can you revert this change? It doesn't seem relevant to this PR

Cheddam:
It should be safe to bump the requirements to `4.7@dev`, regardless of whether the branches exist yet - Composer will resolve it to `4`.

robbieaverill:
I don't know if that's true, given that we've removed all the branch aliases

Cheddam:
Looks like the bump worked - @robbieaverill are you happy to merge?

PR #2566, opened 2020-07-07
PATCH: change "with No Content" to "without content" in reports

sunnysideup:


sunnysideup:
I checked the dutch translation.

The dutch translation said: "zonder" - which is translated to "without". 

Also, "with no" is basically bad English for "without" so I don't really see your point - i.e. if the translations were bad in the first place then they will not get worse by changing the English to better English?

sunnysideup:
i  am happy to check Spanish / German / French the the reo if the helps?

robbieaverill:
Thanks for checking. In this case it’s probably not an issue, but usually we don’t change translation source (English) definitions, because people have translated “with no” rather than “without” when may then invalidate the translated versions. I presume it’ll be fine in this case because it’s a minor adjustment.

sunnysideup:
Totally agree.

robbieaverill:
Would you mind updating the failed behat test copy too?

sunnysideup:
Done

robbieaverill:
cc @silverstripeux - if you’re happy here feel free to merge

PR #2570, opened 2020-07-17
NEW: Add getPageByLink GraphQL operation

unclecheese:
Usage:
```
query {
  getPageByLink(Link: "about-us") {
    ID
    Link
  }
}
```

### DIFF
--- a/_config/graphql.yml
+++ b/_config/graphql.yml
@@ -16,3 +16,12 @@ SilverStripe\GraphQL\Manager:
             fields: [ID, FirstName, Surname]
             operations:
               readOne: true
+    default:
+      scaffolding:
+        types:
+          SilverStripe\CMS\Model\SiteTree:
+            fields:
+              ID: true
+              Link: true
+      queries:
+        getPageByLink: SilverStripe\CMS\GraphQL\GetByLinkCreator
--- a/code/GraphQL/GetByLinkCreator.php
+++ b/code/GraphQL/GetByLinkCreator.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace SilverStripe\CMS\GraphQL;
+
+use GraphQL\Type\Definition\Type;
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\GraphQL\QueryCreator;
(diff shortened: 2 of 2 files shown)


_config/graphql.yml
```diff
@@ -16,3 +16,12 @@ SilverStripe\GraphQL\Manager:
             fields: [ID, FirstName, Surname]
             operations:
               readOne: true
+    default:
```
chillu, line 19: Hmmm, this is introducing a GraphQL API in to every project using the CMS now. What are the implications of that? Is there auth that only applies to the `admin` API?

code/GraphQL/GetByLinkCreator.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * @param $obj
+     * @param array $args
+     * @return SiteTree|null
+     */
+    public function resolve($obj, array $args = [])
+    {
+        $link = $args['Link'] ?? null;
```
chillu, line 52: A bit nitpicky but `Type::nonNull()` makes this `null` handling obsolete here, right? It could just throw if you pass in null.

unclecheese: oh yeah, forgot i did that! :)

_config/graphql.yml
unclecheese: No.. the `default` schema was always pre-configured for you, but it doesn't actually activate until you route it. All this configuration doesn't turn into anything executable until this happens:

```
SilverStripe\Control\Director:
  rules:
    'graphql': '%$SilverStripe\GraphQL\Controller.default'
```

michalkleiner:
Firstly I typed "Should the name of the method somehow reflect that's a relative link?" but then I looked at the resolver and it's doing `SiteTree::get_by_link` which actually supports URLs including the protocol, hostname and port, through calling `Director::makeRelative` on the provided path — does the query parameter description need to reflect that or do you want to keep it as a 'hidden' feature?

unclecheese:
Superceded by GraphQL 4 compat: https://github.com/silverstripe/silverstripe-cms/pull/2577/commits/43fa0b67db4c2aebf6d0c8423fd2890c92555263

Issue #1945, opened 2017-08-31
Redirect page preview toolbar not fixed to bottom of viewport


clarkepaul:
You can't close the preview as you can't get to the actions. 
<img width="577" alt="pasted_image_31_08_17__3_40_pm" src="https://user-images.githubusercontent.com/555033/29905433-d5e51096-8e62-11e7-9ad1-3e58335e7b21.png">

### PRs

- [x] https://github.com/silverstripe/silverstripe-cms/pull/2572 << - merge this first, then rerun tests on admin
- [x] https://github.com/silverstripe/silverstripe-admin/pull/1086

Blocked by https://github.com/silverstripeltd/product-issues/issues/450

chillu:
Removing milestone, it's a fringe use case (previewing redirector pages isn't common I could imagine)

robbieaverill:
Still a bug

clarkepaul:
A few options:
1. Close the preview panel for this page type (disable the preview mode toggle)
2. Have a no no "preview available message - show URL"
3. Show the page you are directing the person too

I think the opt. 1 would probably be the easiest and most practical approach, and opt. 3 is the nicest for the user (might not be easy to do though?).

phptek:
I just came across this yesterday myself in 4.4.3. Doesn't look too polished does it!?

clarkepaul:
Left a message in the PR about it passing my criteria, looks good. 

I've opened another issue to take care of some other related stuff... Contrast of messaging when no preview shows, and centering of message when viewing the message at different device sizes captured here https://github.com/silverstripe/silverstripe-cms/issues/2671

maxime-rainville:
Merged the first PR and rerunning the build on the second PR.

maxime-rainville:
Build on second PR is failing becasue of https://github.com/silverstripeltd/product-issues/issues/450

maxime-rainville:
All merge

PR #2572, opened 2020-07-30
FIX Show 'No preview available' message for pages with no state

Cheddam:
Issue https://github.com/silverstripe/silverstripe-cms/issues/1945#issuecomment-869322412

This accompanies a [logic patch in the admin module](https://github.com/silverstripe/silverstripe-admin/pull/1086) to ensure that when no preview state is available for a pagetype, it renders a message stating this. Previously a broken layout would appear instead.

### DIFF
--- a/templates/SilverStripe/CMS/Controllers/Includes/CMSMain_PreviewPanel.ss
+++ b/templates/SilverStripe/CMS/Controllers/Includes/CMSMain_PreviewPanel.ss
@@ -1,6 +1,9 @@
 <div class="cms-preview fill-height flexbox-area-grow" data-layout-type="border">
 	<div class="panel flexbox-area-grow fill-height">
-		<div class="preview-note"><span><!-- --></span><%t SilverStripe\CMS\Controllers\CMSPageHistoryController.PREVIEW 'Website preview' %></div>
+		<div class="preview-note">
+            <div class="icon font-icon-monitor display-1"></div>
+            <%t SilverStripe\CMS\Controllers\CMSPageHistoryController.NO_PREVIEW 'No preview available' %>
+        </div>
 		<div class="preview__device">
 			<div class="preview-device-outer">
 				<div class="preview-device-inner">
@@ -10,5 +13,4 @@
 		</div>
 	</div>
 	<div class="toolbar toolbar--south cms-content-controls cms-preview-controls"></div>
-	<div class="cms-preview-overlay ui-widget-overlay-light"></div>
 </div>
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "silverstripe/admin": "^1.7@dev",
+        "silverstripe/admin": "^1.9@dev",
         "silverstripe/campaign-admin": "^1.7@dev",
         "silverstripe/framework": "^4.7@dev",
         "silverstripe/reports": "^4.7@dev",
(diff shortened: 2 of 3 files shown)


templates/SilverStripe/CMS/Controllers/Includes/CMSMain_PreviewPanel.ss
```diff
@@ -1,6 +1,9 @@
 <div class="cms-preview fill-height flexbox-area-grow" data-layout-type="border">
 	<div class="panel flexbox-area-grow fill-height">
-		<div class="preview-note"><span><!-- --></span><%t SilverStripe\CMS\Controllers\CMSPageHistoryController.PREVIEW 'Website preview' %></div>
+		<div class="preview-note">
+            <div class="icon font-icon-monitor display-1"></div>
```
sachajudd, line 4: What's the extra icon class for?

emteknetnz: @Cheddam ?

bergice:
Looks good!

When toggling the view type, there is some blank margin being added to the right side:
image

Small suggestion, maybe the modes should just be greyed out (with tooltip explanations) as there is no reason for using them: 
image

brynwhyman:
> When toggling the view type, there is some blank margin being added to the right side:

I think we should get this resolved before merging. 

@clarkepaul what are your thoughts on disabling the 'split' and 'preview' modes in this scenario?

clarkepaul:
@brynwhyman If the issue is anything to do with the preview width we can get rid of that just for this view. 

Is there maybe a case for a person being stuck in Preview mode and not being able to navigate out into edit/split mode? If the redirect is working properly they should be redirected away but just thinking of edge cases or errors.

The mode is also set for all pages you view afterwards (unless changed manually). Not such a biggy to disable it but it would also stop people from changing between Edit and Split modes if they wished.

I would think that the contents of the dropdown just has a line of CSS conflicting or positioned outside the area. Could be as simple as applying overflow: hidden to fix the issue? I would prefer an attempt to fix the CSS first before disabling the dropdown.

Cheddam:
I can actually reproduce the weird overflow bug without these PRs present, so perhaps we can raise a separate issue to track resolving that?

Cheddam:
Have validated that the CI failure on this one is an [existing issue](https://travis-ci.com/github/silverstripe/silverstripe-cms/jobs/488334077). This PR needs to be merged before the admin one for that to go green.

emteknetnz:
@Cheddam could you reply to this [old feedback](https://github.com/silverstripe/silverstripe-cms/pull/2572/files#r464154272) before getting merged

emteknetnz (approved):

emteknetnz (changes requested):
We'll need to add a constraint of admin ^1.8.1 (or whatever the corresponding admin tag is) otherwise this will happen when you click a regular page, then the redirector page in split view

image

maxime-rainville:
@clarkepaul What's our view on the contrast of these colours? Is this an accessibility problem?

image

maxime-rainville:
As @bergice mentioned, there's some funkiness going on when you start using the preview width feature. I also don't think this is worth fixing. Here's what it looks like for reference: https://youtu.be/VxajIPsVx7M

clarkepaul:
Happy with the fix (just got a walk through from @emteknetnz), toolbar is now fixed to the bottom, user can still switch between view modes which is a good result.

The issue of the messaging not being in the centre for the varying screen sizes which are not "Auto" can be raised as a separate issue. Will log soon.

@maxime-rainville in regards to contrast of messaging, it doesn't meet the standards so I can raise an issue for that. The icon should be at least 3:1 and the text 4.5:1. The icon could be swapped out for a SVG now and lucky the text is actually text so easy to improve.

PR #2573, opened 2020-08-05
Making 3.7 build green

maxime-rainville:
The 3.7 build is broken because of the 5.4/PostgreSQL matrix. First, I'll try to see if the problem is with 5.4 or PostgreSQL.


maxime-rainville:
Humm ... looks like everything is green now. I'll blame goblins.

PR #2575, opened 2020-08-10
Read from `ALL_VERSIONS` when looking up page history

bergice:


client/src/state/history/readOnePageQuery.js
```diff
@@ -9,7 +9,7 @@ const query = gql`
 query ReadHistoryViewerPage ($page_id: ID!, $limit: Int!, $offset: Int!) {
   readOnePage(
```
emteknetnz, line 10: Is there an existing corresponding unit test for `ReadHistoryViewerPage readOnePage` ?

bergice: There is `\SilverStripe\CMS\Tests\Controllers\CMSPageHistoryControllerTest`, we might be able to add something to check for archived entries there.

But it should already be covered by the `versioned` module: https://github.com/silverstripe/silverstripe-versioned-admin/pull/160/files.

emteknetnz: OK, I guess given it's just altering a graphql query it's probably all good

PR #2578, opened 2020-08-21
NEW Set VersionedProvider labels for recipe-core and recipe-cms

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-installer/issues/276

Related PR (merged) to add functionality https://github.com/silverstripe/silverstripe-framework/pull/9628

Reason that recipes are added here rather than within silverstripe/recipe-core/cms is because recipes are included in the projects vendor folder like normal modules

If we were to add the config in the recipes, then we would need to have a pair of seperate yml files copied into the project app/_config folder which would be pretty annoying for a relatively minor feature

### DIFF
--- a/_config/version.yml
+++ b/_config/version.yml
@@ -6,3 +6,4 @@ SilverStripe\Core\Manifest\VersionProvider:
   modules:
     silverstripe/framework: ''
     silverstripe/cms: CMS
+    silverstripe/recipe-cms: CMS Recipe


_config/version.yml
```diff
@@ -5,4 +5,6 @@ After: 'framework/*'
 SilverStripe\Core\Manifest\VersionProvider:
   modules:
     silverstripe/framework: ''
+    silverstripe/recipe-core: Core Recipe
```
Cheddam, line 8: Would it make more sense for this to be preferred over silverstripe/cms?  The recipe version paints a broader picture of what you're "running", though the version of the CMS module could vary 🤔

emteknetnz: Order based of revision Max recommended https://github.com/silverstripe/silverstripe-framework/pull/9628#issuecomment-669629976

Also recipe-core lacks CMS, so it probably makes no difference if we swapped them.

Cheddam: Okay, in that case maybe the `recipe-core` reference belongs in framework, since it'll never be used here 🤔

emteknetnz: Framework doesn't have this config cos the feature lives in CMS.

No much point having a CMS version indicator if you have a CMSless install =)

michalkleiner: CMS brings SiteTree, mainly, but you can still have an admin back-end and manage things through model admins e.g. for an API provider sort of installation. Does silverstripe-admin have this feature to report on framework it's using?

Cheddam: Had a bit of a dig, and it appears this functionality lives in the _admin_ module, not the cms module. I wasn't suggesting moving the CMS entry to framework, just the `recipe-core` one, since it will always be overridden by the CMS entry if it's listed here anyway.

My final recommendation: Remove the `recipe-core` reference here since it's inert. If we want it to actually be a viable choice for the `VersionProvider`, add it back either in framework or admin.

emteknetnz: @Cheddam have removed the mention of recipe-core.  We probably not need a PR in framework or admin since VersionProvider is only ever used within the CMS.  We probably should never even mentioned it in the ACs

Cheddam: I think you're conflating "the CMS" with the Admin UI. Here's a project that uses `recipe-core` and `admin` - the `cms` module is not present:

Screen Shot 2020-08-26 at 14 15 57

So I think another PR to framework or admin makes sense. Regardless, the part pertaining to this PR is resolved 👍

Cheddam (approved):

Issue #2579, opened 2020-08-26
Unpublish permission inconsistency


mfendeksilverstripe:
# Unpublish permission inconsistency

Tested on version `4.5.1`

It seems that the Unpublish permission is inconsistent between UI and action execution.

## UI permission check

* requires both `canPublish` and `canUnpublish`

```
// "unpublish"
if ($isPublished && $canPublish && $isOnDraft && $canUnpublish) {
    $moreOptions->push(
        FormAction::create('unpublish', _t(__CLASS__.'.BUTTONUNPUBLISH', 'Unpublish'), 'delete')
            ->setDescription(_t(__CLASS__.'.BUTTONUNPUBLISHDESC', 'Remove this page from the published site'))
            ->addExtraClass('btn-secondary' . ($this->isHomePage() ? ' homepage-warning' : ''))
    );
}
```

## Action execution

* requires only `canUnpublish`

```
if ($record && !$record->canUnpublish()) {
    return Security::permissionFailure($this);
}
```

## Example use case

* publishing requires audit (workflow)
* unpublishing doesn't require audit

In such case it doesn't make sense to tie publish and unpublish permission together. This is only one example but I think we want to have these permissions separated in general.

## Suggested change

Remove the `canPublish` condition from the UI check.

## Solution

https://github.com/silverstripe/silverstripe-cms/pull/2581

PR #2581, opened 2020-08-31
BUG: Unpublish permission decoupled from publish permission.

mfendeksilverstripe:
# Unpublish permission decoupled from publish permission

* unpublish action no longer requires explicit publish permission

## Backwards compatibility

* projects with out of the box setup will not be impacted as unpublish permission falls back to publish permission
* the change only impacts projects which use explicit and separate permissions for publish / unpublish actions
* note that separating these permissions makes sense in cases where publishing requires an audit and is considered risky action while unpublishing not so much

## Related issues

https://github.com/silverstripe/silverstripe-cms/issues/2579

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -2371,7 +2371,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         }
 
         // "unpublish"
-        if ($isPublished && $canPublish && $isOnDraft && $canUnpublish) {
+        if ($isPublished && $isOnDraft && $canUnpublish) {
             $moreOptions->push(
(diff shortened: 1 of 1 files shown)


emteknetnz:
Seems like a change of behaviour so it may be an enhancement instead?  Or would you say that that current behaviour is unintentional and you're just fixing it?

mfendeksilverstripe:
@emteknetnz Current behaviour is incorrect and this is fixing it.

emteknetnz:
OK cool, I'll look to get this merged in the next couple of weeks

emteknetnz:
Cool looks good, as mentioned in the original description, out of the box there is no unpublish permission and it just fallsback to the public permissions

```php
public function canUnpublish($member = null)
...
        // Default to relying on canPublish
        return $owner->canPublish($member);
```

Issue #1350, opened 2015-12-22
CMSBatchAction::getParameterFields() never seems to get called/used


micschk:
`CMSMain::BatchActionParameters()` calls `getParameterFields()` for each batch action. This is supposed to add additional parameter fields to the batch actions in CMS, but it seems `CMSMain::BatchActionParameters()` never gets called and thus no parameter dropdown gets added.

### PRs
- [x] ~~https://github.com/silverstripe/silverstripe-cms/pull/2590~~
- [x] ~~https://github.com/silverstripe/silverstripe-admin/pull/1124~~
- [ ] https://github.com/silverstripe/silverstripe-cms/pull/2629
- [ ] https://github.com/silverstripe/silverstripe-cms/pull/2628
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/1167
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/1168

tractorcow:
I guess we should remove it in 4.x? do you agree?

GuySartorelli:
Batch actions parameters are now used per https://github.com/silverstripe/silverstripe-cms/pull/2746

PR #2590, opened 2020-09-25
Removing BatchActionParameters as it's not used anymore

bergice:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/1350


emteknetnz (changes requested):
Even though they're unused, removing public functions is a semver breaking change so this needs to target SilverStripe 5.

dhensby:
We should probably add deprecation notices to v4, though

chillu:
@bergice Can you please rewrite this PR to deprecation notices, or close it?

Issue #1511, opened 2016-05-27
Site tree scroll to selected page doesn't work (previously was overshooting/undershooting with large trees)


clarkepaul:
With large trees, if you select an item and the page reloads to go that page for editing the site tree also reloads but the selected item is not in current view anymore and overshoots it.

The desired behaviour is that the currently selected CMS page is visible in the pages list when initially loaded.

### PR
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2591

hoKRa1:
Is this still an issue?

clarkepaul:
This is no longer the issue but a new one, in v4 the sitetree does not scroll down to the selected page anymore at all from my test. Will update title of issue.

maxime-rainville:
Fixed by https://github.com/silverstripe/silverstripe-cms/pull/2591

PR #2591, opened 2020-09-25
Fixes pages not being scrolled to on treeview

bergice:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/1511


bergice:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/1511

emteknetnz (changes requested):
This is definitely a big improvement from the current behaviour

One thing we could improve is to add "one-row" of padding at the top so that the selected page is not at the very top.  With this PR it will scroll to here:

image

I think it might be nicer if it scrolled to here instead

image

Alternatively, maybe it would make sense for to be "one from the bottom"
image

@sachajudd what's your view?

sachajudd:
Near the top makes more sense than at the bottom e.g if you're opening up a page with children pages it'd be a hassle to keep scrolling after the first initial scroll. Not fussed either way where the selected page lands. Looks good either way nice work @bergice

emteknetnz:
@sachajudd great ta.  So at the top with or without "one row" of padding?

bergice:
@emteknetnz it should select the previous list item now.

maxime-rainville (approved):
Works well tested in IE11, Firefox and Chrome.

Issue #2359, opened 2019-01-08
SiteTree search for '+' in a URLSegment has unexpected behaviour


jakxnz:
## Version

silverstripe/cms 4.3.x-dev 309dc97

## Description

When `URLSegment` is available in the CMS SiteTree search (`pages/`), if searching for a URL with a `+` in it, including `+` in the search string yields no result.

**Please Note: SilverStripe does not preserve `+` symbols in URLs by default**

## Prerequisites

1. SilverStripe configured so that `+` symbols can be used in `URLSegment` 
2. SilverStripe CMS, with `URLSegment` available in `SiteTree`'s searchable fields, and in the CMS search form.
3. An existing Page with a URLSegment that contains a `+` e.g `url+segment`

## Steps

1. Log into the CMS
2. Browse to the `pages/` section
3. Click the Search icon
4. Expand the form to input a value for a specific search field
3. Type `+segment` into the "URL Segment" input
4. Submit

### Expected result

The existing Page with `URLSegment` `url+segment` is visible amongst the search results

### Actual result

The existing Page with `URLSegment` `url+segment` does not appear

## Notes

I notice that any search value that is submitted this way has `+` replaced with ` ` (space). I also notice that if the replacement is bypassed, when `%2Bsegment` or `+segment` is submitted via get params the value is parsed as `+segment`, but the `URLSegment` value is stored with URL encoding - so this is a non-result as well. The value needs to be parsed as a url encoded `+` (`%2B` in order to yield a result).

maxime-rainville:
How did you manage to a put a `+` sign in your URL segment? When I try to do that it automatically gets swapped out for a dash, which I think is the intended behaviour.

Looks like special characters are not being URL encoded on the front end when submitting the site tree search, which should be a simple fix.

jakxnz:
Ah, yes, sorry @maxime-rainville, our project has allowed `+` symbols in `URLSegment`. It would be fantastic if the posterity of submitted search parameters could be preserved in one way or another.

maxime-rainville:
Cool ... can you post the code snippet that would allow us to enable + signs in URL Segment for reference?

mfendeksilverstripe:
This issue can be closed as it was fixed via https://github.com/silverstripe/silverstripe-cms/pull/2593

PR #2593, opened 2020-10-05
BUG: URL segment filter correction for CMS page search.

mfendeksilverstripe:
# URL segment filter correction for CMS page search.

Affected version:

`4.5.1`

## Description

URL segment search lacks encoding which causes the search results to omit some pages.

## Example

Consider a page with the following URL segment:

`lake-wanaka+adventure`

Searching for this page will yield no results because the search matches raw value which doesn't work as the DB value is stored in encoded format which looks like this:

`lake-wanaka%2Badventure`

## Solution

Apply encoding to all places where URL segment value is passed to DB lookup.

## Notes

* we can't rely on `URLSegmentFilter` to sanitise the values in general because the sanitisation rules are configurable via the `default_replacements` configuration API
* we need to make sure any encoded values can be used for search

## Related issues

https://github.com/silverstripe/silverstripe-cms/issues/2359

### DIFF
--- a/code/Controllers/CMSSiteTreeFilter.php
+++ b/code/Controllers/CMSSiteTreeFilter.php
@@ -5,6 +5,7 @@ namespace SilverStripe\CMS\Controllers;
 use SilverStripe\Admin\LeftAndMain_SearchFilter;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Forms\DateField;
 use SilverStripe\ORM\DataList;
@@ -203,11 +204,17 @@ abstract class CMSSiteTreeFilter implements LeftAndMain_SearchFilter
             switch ($name) {
                 case 'Term':
                     $query = $query->filterAny(array(
-                        'URLSegment:PartialMatch' => $val,
+                        'URLSegment:PartialMatch' => Convert::raw2url($val),
                         'Title:PartialMatch' => $val,
                         'MenuTitle:PartialMatch' => $val,
                         'Content:PartialMatch' => $val
                     ));
+                    break;
+
+                case 'URLSegment':
+                    $query = $query->filter([
+                        'URLSegment:PartialMatch' => Convert::raw2url($val),
+                    ]);
                     break;
 
                 case 'LastEditedFrom':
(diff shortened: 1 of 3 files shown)


code/Controllers/CMSSiteTreeFilter.php
```diff
@@ -203,13 +204,19 @@ protected function applyDefaultFilters($query)
             switch ($name) {
                 case 'Term':
                     $query = $query->filterAny(array(
-                        'URLSegment:PartialMatch' => $val,
+                        'URLSegment:PartialMatch' => Convert::raw2url($val),
```
dhensby, line 207: Perhaps we should use `rawurlencode()` as that's what is used here: https://github.com/silverstripe/silverstripe-cms/blob/4.6.0/code/Model/SiteTree.php#L425

dhensby (approved):

code/Controllers/CMSSiteTreeFilter.php
mfendeksilverstripe: I tried your suggestion but this breaks the newly added unit test which indicates that this doesn't cover all cases.

dhensby: Interesting - that implies `SiteTree::get_by_link()` will be broken in some instances too?

dhensby:
> * we can't rely on `URLSegmentFilter` to sanitise the values in general because the sanitisation rules are configurable via the `default_replacements` configuration API

I note that `Convert::raw2url()` uses the `URLSegmentFilter` to do the sanitisation - if anything we *have* to use `URLSegmentFilter` because that is what `SiteTree` uses when writing to the DB. I'm not saying anything in the PR needs to change, just this note is confusing.

code/Controllers/CMSSiteTreeFilter.php
dhensby: OK - I think that `rawurlencode()` is used in `get_by_link` because the assumption is that the provided url would have all the conversions done already (they'd be in the actual URL that a request has been made to). If `URLSegmentFilter` was used, for example, then a search for `you&me` would return a page `you-and-me` even though you wouldn't expect that to be the case.

mfendeksilverstripe:
@dhensby Sorry for the confusing note, I'll update it :)

Cheddam:
Noting that this change failed in the [CWP Kitchen Sink build](https://travis-ci.org/github/silverstripe/cwp-recipe-kitchen-sink/jobs/734110038). Looking into the cause now.

mfendeksilverstripe:
@Cheddam I recommend reviewing the `default_replacements` in `URLSegmentFilter` which is used to run the test. If needed you can override this config via [withConfig](https://github.com/silverstripe/silverstripe-framework/blob/4/src/Core/Config/Config.php#L129)

Cheddam:
@mfendeksilverstripe I ended up replicating the change in Elemental's version of this logic - see https://github.com/dnadesign/silverstripe-elemental/pull/841.

Issue #1391, opened 2016-02-06
Custom Meta Tags : Closing tag gets stripped


MischaKr:
Larger explanation: 
http://www.silverstripe.org/community/forums/content-editor-discussions/show/99962

Tested on: 
IIS 8.5 
PHP 5.6 

Silverstripe version: 
3.1.x 
3.2.1

tractorcow:
In HTMLText::whitelistContent, there is a bit of code that parses then re-generates all allowed tags safely.

The only allowed tags for this field is link and meta, by default.

It uses the HTMLValue service to parse this content; Of which there are a few classes available. SS_HTML4Value is one, and SS_HTML5Value. Can you check which one your site is using?

I wonder if SS_HTML4Value is not properly closing the tags, and we need an SS_XHTML4Value?

MischaKr:
I've added a dirty workaround to the whitelistContent function to not parse the value based on input. 

Can't confirm which class is used on my site, but I can't find SS_HTML5VALUE in php files so it seems very unlikely that that does anything :)

tractorcow:
Oh right, it's in a separate module. Check out https://github.com/silverstripe/silverstripe-html5 to see if that's a better solution to your problem.

maxime-rainville:
Still looks like this is happening in SS4.

brynwhyman:
I believe with HTML5, there's no need for the closing tag ( `/>`). I'm guessing in CMS4 it's being parsed and saving with the 'correct' value (`>`).

I think we should close this issue as it's not reproducible on CMS v4 and raise another issue to update the label description to remove reference to the closing tag. Thoughts?

> Description label: "`HTML tags for additional meta information. For example <meta name="customName" content="your custom content here" />`"

dhensby:
agreed - does someone want to create a PR to update the docs?

brynwhyman:
Here you go @dhensby : https://github.com/silverstripe/silverstripe-cms/pull/2597

PR #2597, opened 2020-10-12
FIX remove old syntax from meta tag label

brynwhyman:
Resolves: https://github.com/silverstripe/silverstripe-cms/issues/1391


dhensby (approved):
thanks

PR #2602, opened 2020-10-21
MNT Fix unit tests not running on PGSQL build

Cheddam:
This matrix entry was missing `PHPUNIT_TEST=1`, and therefore wasn't running any tests. This was masking a test failure with PostgreSQL, and this test suite is [currently failing in silverstripe/installer](https://travis-ci.org/github/silverstripe/silverstripe-installer/jobs/737110775).


Cheddam:
I've added a fix for the failing PostgreSQL build. I'm still not sure this is a clear solution to the root cause - since the test was passing when run against MySQL - but this seems like a sensible adjustment anyway, since the tests were previously operating outside of any Stage.

emteknetnz (approved):
MOG

PR #2577, opened 2020-08-20
GraphQL4 compatibility

unclecheese:
See https://github.com/silverstripe/silverstripe-graphql/pull/266

PR #2607, opened 2020-11-02
NEW: GraphQL Compatibility

unclecheese:
Moved from: https://github.com/silverstripe/silverstripe-cms/pull/2577


_config/graphql.yml
```diff
@@ -1,18 +1,47 @@
...
-SilverStripe\GraphQL\Manager:
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\GraphQL\Schema\Registry\ResolverRegistry:
+    constructor:
+      cmsResolver: '%$SilverStripe\CMS\GraphQL\Resolver'
+  SilverStripe\GraphQL\Schema\Registry\PluginRegistry:
+    constructor:
+      linkable: '%$SilverStripe\CMS\GraphQL\LinkablePlugin'
+SilverStripe\CMS\Model\SiteTree:
+  graphql_blacklisted_fields:
```
chillu, line 14: Why did that change from a whitelist (`fields: [ID, LastEdited, AbsoluteLink]`) to a blacklist?

code/GraphQL/LinkablePlugin.php
```diff
@@ -0,0 +1,115 @@
...
+    public function apply(ModelQuery $query, Schema $schema, array $config = []): void
+    {
+        $class = $query->getModel()->getSourceClass();
+        // Only site trees have the get_by_link capability
+        if ($class !== SiteTree::class && !is_subclass_of($class, SiteTree::class)) {
+            return;
+        }
+        // if the query is intended to return a list, `link` doesn't apply here
+        if ($query->isList()) {
+            return;
```
chillu, line 50: Shouldn't those two conditionals throw instead? Silent errors are the devil.

chillu (changes requested):

chillu (changes requested):

_config/graphql.yml
unclecheese: Because this affects _all schemas_. It's an attribute of the SiteTree class itself. We need to throttle back what happens when you do `fields: '*'`, which is going to be very common with SiteTree.

code/GraphQL/LinkablePlugin.php
unclecheese: Yup, fixed.

code/GraphQL/_legacy/ReadOneResolver.php
```diff
@@ -0,0 +1,28 @@
...
+ * Shim to make readOnePage work like GraphQL 4
+ */
+class ReadOneResolver
+{
+    public static function resolve($obj, array $args, array $context, ResolveInfo $info)
+    {
+        $idKey = StaticSchema::inst()->formatField('ID');
+        $id = $args['filter'][$idKey]['eq'];
+        $readOne = Injector::inst()->createWithArgs(ReadOne::class, ['Page']);
+        $args = [$idKey => $id];
```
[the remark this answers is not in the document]
unclecheese: Fixed

_config/graphql.yml
chillu: Right, I've verified that there's exceptions when you try to include blacklisted fields, which is great since it's quite an opaque setting (outside of the GraphQL schema config). I've documented the status quo via https://github.com/silverstripe/silverstripe-framework/pull/9768, but also suggested a refactor via  https://github.com/silverstripe/silverstripe-graphql/issues/322. Doesn't block this merge though.

chillu (approved):

PR #2615, opened 2020-11-19
NEW compatibility with new stateless plugins

unclecheese:
See : https://github.com/silverstripe/silverstripe-graphql/pull/327


chillu:
The actual contents of this PR were already committed to the `4` branch through https://github.com/silverstripe/silverstripe-cms/commit/7b960c15374e74718b743c2378c23775fedcb159, but there was a regression where `LinkablePlugin` was accidentally removed. This PR reintroduces it.

PR #2617, opened 2020-11-23
NEW: Compliance with new SchemaConfig API

unclecheese:
See: https://github.com/silverstripe/silverstripe-graphql/pull/329

### DIFF
--- a/_graphql/config.yml
+++ b/_graphql/config.yml
@@ -1,0 +1,9 @@
+resolvers:
+  - 'SilverStripe\CMS\GraphQL\Resolver'
+modelConfig:
+  DataObject:
+    operations:
+      read:
+        plugins:
+          getByLink:
+            after: filter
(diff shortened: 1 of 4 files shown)


chillu (approved):

_graphql/config.yml
```diff
@@ -0,0 +1,9 @@
+resolvers:
```
michalkleiner, line 1: I get that this yaml might be used differently than all other `Config`-related yaml files, it just feels weird it doesn't have a root node/value either as an existing class or at least some string, like GraphQL or so. I know it might be redundant in terms of verbosity, but... Just a thought.

unclecheese: Yeah good point. This is a matter of preference.You can also do an arbitrary file name, like `schema.yml` containing any keys you like

```yaml
config: ...
types:... 

# etc..
```

PR #2620, opened 2020-12-07
BUGFIX: Restore, refactor getByLink plugin

unclecheese:
It had been accidentally removed from the registry.

Refactored so that it's always its own argument to the query, and should not apply to list queries, only single record queries.

### DIFF
--- a/_config/graphql.yml
+++ b/_config/graphql.yml
@@ -15,7 +15,3 @@ SilverStripe\GraphQL\Schema\Schema:
     admin:
       src:
         - 'silverstripe/cms: _graphql'
-SilverStripe\Core\Injector\Injector:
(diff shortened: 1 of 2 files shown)


_config/graphql.yml
```diff
@@ -15,3 +15,7 @@ SilverStripe\GraphQL\Schema\Schema:
     admin:
       src:
         - 'silverstripe/cms: _graphql'
+SilverStripe\Core\Injector\Injector:
```
chillu, line 18: I've already added this on the original location (a bit further up) in the same file: https://github.com/silverstripe/silverstripe-cms/blob/4/_config/graphql.yml#L6-L9

chillu:
I've sent a follow up pull request for docs on https://github.com/silverstripe/silverstripe-framework/pull/9793

PR #2639, opened 2021-03-05
NEW Add ability to redirect to a file on the website

wilr:


code/Model/RedirectorPage.php
```diff
@@ -39,19 +41,27 @@ class RedirectorPage extends Page
 
     private static $has_one = [
         "LinkTo" => SiteTree::class,
+        "LinkToFile" => File::class
+    ];
+
+    private static $owns = [
+        'LinkToFile'
```
michalkleiner, line 48: I don't think a redirect should own the file. If it's not published separately then the redirect won't work, but that'd be expected from my perspective.

code/Model/RedirectorPage.php
```diff
@@ -39,19 +41,27 @@ class RedirectorPage extends Page
 
     private static $has_one = [
         "LinkTo" => SiteTree::class,
+        "LinkToFile" => File::class
```
michalkleiner, line 44:
```suggestion
        "LinkToFile" => File::class,
```

package.json
```diff
@@ -9,7 +9,8 @@
     "node": "^10.x"
   },
   "scripts": {
-    "build": "yarn && yarn test && NODE_ENV=production webpack -p --bail --progress",
+    "build": "yarn && yarn test && yarn prod",
+    "prod": "NODE_ENV=production webpack -p --bail --progress",
```
michalkleiner, line 13: Unrelated but ok, let's keep this

code/Model/RedirectorPage.php
wilr: @michalkleiner Thanks, removed the `$owns`

code/Model/RedirectorPage.php
wilr: Added

michalkleiner (changes requested):
Thanks for the tweaks, @wilr.

Can you please add to/update the [RedirectorPageTest](https://github.com/silverstripe/silverstripe-cms/blob/4/tests/php/Model/RedirectorPageTest.php) as well to test the newly added ability? Would be much appreciated!

GuySartorelli:
@wilr I'm wondering what the reason is for changing from single quotes to double quotes? Iirc silverstripe mostly follows Airbnb longing rules which use single quotes.

wilr:
> @wilr I'm wondering what the reason is for changing from single quotes to double quotes? Iirc silverstripe mostly follows Airbnb longing rules which use single quotes.

Double quotes would have been automatically added via vscode for whatever the `.editorconfig` setup is.

michalkleiner:
> Double quotes would have been automatically added via vscode for whatever the .editorconfig setup is.

There's nothing around this in the `.editorconfig` file so these should be reset @wilr and only the necessary changes for this PR applied (adding the extra show/hide logic for the new field).

michalkleiner:
> > > Double quotes would have been automatically added via vscode for whatever the .editorconfig setup is.
> > 
> > 
> > There's nothing around this in the `.editorconfig` file so these should be reset @wilr and only the necessary changes for this PR applied (adding the extra show/hide logic for the new field).
> 
> I'm picking it comes from https://github.com/silverstripe/silverstripe-cms/blob/4/.eslintrc.js then perhaps?

Not that I can see either. It references https://github.com/silverstripe/eslint-config/blob/master/.eslintrc.js and there's no mention of a quote style rule.

michalkleiner:
As @GuySartorelli pointed out, the eslint config SS uses [extends the AirBnb one](https://github.com/silverstripe/eslint-config/blob/master/.eslintrc.js#L35) and that prefers single quotes https://airbnb.io/javascript/#strings--quotes.

package.json
GuySartorelli: I'm not sure what the point of this change is? Tests shouldn't be skipped when building out when it's ready to commit, and if developers want to skip tests during development that's what the `dev` script is for.

GuySartorelli (changes requested):
Some of the tests failed and there are some conflicts now - if you're willing to clean those up and either remove or justify the new yarn script I'll be happy to test and merge.

wilr:
@GuySartorelli The conflict was with the built `bundle.js` I've resolved those now and re-pushed

package.json
```diff
@@ -25,11 +26,15 @@
     "silverstripe",
     "cms"
   ],
+
```
GuySartorelli, line 29:
```suggestion
```

package.json
GuySartorelli: @wilr Is this needed? I'd prefer to keep this as it was (too likely someone will use `prod` instead of `build` when pushing their PR).

michalkleiner: I retract my previous comment. Let's revert this.

wilr: @GuySartorelli reverted that change now. I did that as the test fails for me. Because the tests fail the build doesn't execute. Some error locally about AnchorField.

package.json
wilr: Removed

tests/php/Model/RedirectorPageTest.php
```diff
@@ -117,4 +117,9 @@ public function testNoJSLinksAllowed()
         $page->write();
         $this->assertEmpty($page->ExternalURL);
     }
+
+    public function testFileRedirector()
+    {
+        $this->assertContains("example.pdf", $this->objFromFixture(RedirectorPage::class, 'file')->Link());
```
GuySartorelli, line 123:
```suggestion
        $this->assertStringContainsString("example.pdf", $this->objFromFixture(RedirectorPage::class, 'file')->Link());
```
This syntax changed with the upgrade to PHPUnit 9.5 (https://docs.silverstripe.org/en/4/upgrading/phpunit9/)

GuySartorelli (changes requested):
Couple of small issues to fix.

code/Model/RedirectorPage.php
```diff
@@ -198,6 +220,7 @@ public function getCMSFields()
                         _t(__CLASS__.'.YOURPAGE', "Page on your website"),
                         SiteTree::class
                     ),
+                    new UploadField('LinkToFileID', _t(__CLASS__.'.FILE', "File")),
```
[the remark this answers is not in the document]
wilr: @GuySartorelli done

tests/php/Model/RedirectorPageTest.php
wilr: Done

GuySartorelli:
@wilr Can you please fix the tests? I'm happy to merge on green.

GuySartorelli (approved):
Awesome, thanks @wilr

Issue #2641, opened 2021-03-17
Broken build - Publish a page behat test


emteknetnz:
Happens on both GraphQL 3 + GraphQL4 jobs, so probably unrelated to graphql

https://travis-ci.com/github/silverstripe/silverstripe-cms/jobs/490659470

tests/behat/features/publish-a-page.feature:74

Same failure is happening on the admin repo travis build which runs the same cms test

Broken on:
4.6.x-dev, 4.7.x-dev, 4.x-dev

### PRs
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2642

maxime-rainville:
Fixed by #2642

PR #2642, opened 2021-03-18
MNT Fix archive button behat test

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/2641

Targets 4.6, merge up to 4.7 and 4

### DIFF
--- a/tests/behat/features/publish-a-page.feature
+++ b/tests/behat/features/publish-a-page.feature
@@ -82,9 +82,11 @@ Feature: Publish a page
 
     When I press the "Unpublish" button, confirming the dialog
     And I click "More options" in the "#ActionMenus" element
-    Then I should see an "Archive" button
+    # Use a css-selector instead of the the "Archive" button otherwise it will get confused with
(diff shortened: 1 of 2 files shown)


maxime-rainville (commented):
I'm not sure about this, but wouldn't the proper fix here be to update [`BasicContext::findNamedButton()`](https://github.com/silverstripe/silverstripe-behat-extension/blob/efdd90b19724971d62dea9709da0c3a28e05e51b/src/Context/BasicContext.php#L394-L414) doesn't match links?

emteknetnz:
Probably, though seems risky.  Decent chance of downstream regressions which we won't pick up for up to a week after merging due to the lag of the travis crons

Also some old bespoke projects may still be running behat.

Given this is a bit of an edge case, I don't see too much problem doing what is a bit of a workaround

maxime-rainville (approved):
Fair enough. If we come across more instances where we accidentally match links instead of buttons, we can revisit.

PR #2647, opened 2021-04-01
UPDATE generate canonical via MetaComponents

### LINKED ISSUE #2637: BUG VirtualPage creates second canonical tag in MetaTags
This cropped up on an SEO report from a client of ours. It seems `VirtualPage` simply appends the `CopyContentFrom` page's relative URL to the `parent::MetaTags()` string value. The parent method seems to have already established the canonical URL value, so appending creates a second tag.

There are a couple options I think as a resolution:
- Confirm whether the `MetaTags` method is needed for `VirtualPage`. If the `parent` call generates the anticipated canonical it may not be needed at all.
- If the canonical from the `parent` method is not correct, replace the existing canonical with the correct canonical generated by the `VirtualPage`.

I'm happy to create a pull request for this issue once it's determined which option (or if other options) is preferred.

muskie9:
This replaces using the `MetaTags` method as replacing or removing a canonical requires string manipulation vs unsetting an array key, and is preferred.

Additionally, the virtual page can disable generating a canonical for the proxied page.

resolves #2637

### DIFF
--- a/code/Model/VirtualPage.php
+++ b/code/Model/VirtualPage.php
@@ -134,23 +134,22 @@ class VirtualPage extends Page
     }
 
     /**
-     * For VirtualPage, add a canonical link tag linking to the original page
-     * See TRAC #6828 & http://support.google.com/webmasters/bin/answer.py?hl=en&answer=139394
-     *
-     * @param boolean $includeTitle Show default <title>-tag, set to false for custom templating
-     * @return string The XHTML metatags
+     * @return array
      */
-    public function MetaTags($includeTitle = true)
+    public function MetaComponents()
     {
-        $tags = parent::MetaTags($includeTitle);
+        $tags = parent::MetaComponents();
+
         $copied = $this->CopyContentFrom();
         if ($copied && $copied->exists()) {
-            $tags .= HTML::createTag('link', [
-                'rel' => 'canonical',
-                'href' => $copied->Link()
-            ]);
-            $tags .= "\n";
+            $tags['canonical'] = [
+                'attributes' => [
+                    'rel' => 'canonical',
+                    'content' => $copied->AbsoluteLink(),
+                ],
+            ];
         }
+
         return $tags;
     }
 


code/Model/VirtualPage.php
```diff
@@ -78,6 +80,11 @@ class VirtualPage extends Page
 
     private static $db = [
         "VersionID" => "Int",
+        'UseProxiedForCanonical' => 'Boolean',
```
kinglozzer, line 83: We can’t add this in a patch release, it’d need to be a separate PR against the `4` branch. It’d also need to be `Boolean(1)`, as `private static $defaults` doesn’t affect existing records

kinglozzer (changes requested):
Happy to merge if you can remove the `UseProxiedForCanonical` stuff 😃

muskie9:
@kinglozzer sure thing, I wondered if introducing the new field wouldn't work in the targeted branch, good to know. Let me know if there are any other changes. I'll look at adding the new field once everything is merged up to `4` if that change still makes sense.

kinglozzer (approved):
Looks good 👍

dhensby:
Sorry for being a bit out of it, but why has the method `MetaTags` been replaced with `MetaComponents`?

muskie9:
Hi @dhensby,

The primary thought from our side was more consistent detection/replacement of the meta tags that may be generated by other modules. By implementing the canonical via the `MetaComponents()` modules simply check if the `canonical` key exists in `$tags` vs having to do a find and replace via regular expression or the like (by way of the extension points).

After looking at the methods a bit more I'd also have the opinion that `MetaComponents()` should really handle managing the components to be generated, and the `MetaTags()` method would simply assemble what is returned by `MetaComponents()`. As it stands now, both allow for manipulating the meta tags in different ways which can also cause confusion as to which extension point should be used.

The opinion of the managing vs assembling of the tags may not be valid, but was a thought I had recently so figured I'd throw it out there. When tracking down our initial issue on this I had no idea `MetaComponents()` was a method, and realized it could provide a much more consistent result if community modules utilized it as well.

Cheers!

PR #2663, opened 2021-05-31
NEW Return 404 when redirector page wants to link to missing page

ssmarco:
Issue https://github.com/silverstripe/silverstripe-cms/issues/2665

I recently got a ticket where content authors have quite a number of RedirectorPages 
and that they find it hard to track which ones to update whenever they unpublish the original page.
It makes more sense to them that whenever they unpublish a page, the RedirectorPage should display 404 as well.
This is quite opposite with the existing behaviour where the RedirectorPage maintains a published state.
Adding an extension point in this case provides additional flexibility.

- This extension is only applied whenever the link is null or not found
- One possible scenario is an internal link has been unpublished
- If lots of RedirectorPage then its hard to track and update
- An example use case is to change the content and status code to 404
- Another way is to redirect users to existing 404 page

### DIFF
--- a/code/Model/RedirectorPageController.php
+++ b/code/Model/RedirectorPageController.php
@@ -3,6 +3,7 @@ namespace SilverStripe\CMS\Model;
 
 use SilverStripe\Control\HTTPRequest;
 use PageController;
+use SilverStripe\Control\HTTPResponse_Exception;
 
 /**
  * Controller for the {@link RedirectorPage}.
@@ -12,18 +13,37 @@ class RedirectorPageController extends PageController
     private static $allowed_actions = ['index'];
 
     /**
+     * Should respond with HTTP 404 if the page or file being redirected to is missing
+     */
+    private static bool $missing_redirect_is_404 = true;
+
+    /**
      * Check we don't already have a redirect code set
      *
      * @param  HTTPRequest $request
      * @return \SilverStripe\Control\HTTPResponse
+     * @throws HTTPResponse_Exception
      */
     public function index(HTTPRequest $request)
     {
         /** @var RedirectorPage $page */
         $page = $this->data();
+
+        // Redirect if we can
         if (!$this->getResponse()->isFinished() && $link = $page->redirectionLink()) {
-            $this->redirect($link, 301);
+            return $this->redirect($link, 301);
         }
+
+        // Respond with 404 if redirecting to a missing file or page
+        if (($this->RedirectionType === 'Internal' && !$page->LinkTo()?->exists())
+            || ($this->RedirectionType === 'File' && !$page->LinkToFile()?->exists())
+        ) {
+            if (static::config()->get('missing_redirect_is_404')) {
+                $this->httpError(404);
+            }
+        }
+
+        // Fall back to a message about the bad setup
         return parent::handleAction($request, 'handleIndex');
     }
 
--- a/tests/php/Model/RedirectorPageTest.php
+++ b/tests/php/Model/RedirectorPageTest.php
@@ -8,6 +8,8 @@ use SilverStripe\CMS\Model\RedirectorPageController;
 use SilverStripe\Control\Director;
 use SilverStripe\Assets\File;
 use SilverStripe\Assets\Dev\TestAssetStore;
+use SilverStripe\Control\Controller;
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Dev\FunctionalTest;
 
 class RedirectorPageTest extends FunctionalTest
@@ -59,12 +61,30 @@ class RedirectorPageTest extends FunctionalTest
         );
     }
 
-    public function testEmptyRedirectors()
+    public function provideEmptyRedirectors()
     {
+        return [
+            'use 200' => [
+                'use404' => false,
+            ],
+            'use 404' => [
+                'use404' => true,
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider provideEmptyRedirectors
+     */
+    public function testEmptyRedirectors(bool $use404)
+    {
+        Config::modify()->set(RedirectorPageController::class, 'missing_redirect_is_404', $use404);
         // If a redirector page is misconfigured, then its link method will just return the usual
         // URLSegment-generated value
         $page1 = $this->objFromFixture(RedirectorPage::class, 'badexternal');
         $this->assertEquals('/bad-external', $page1->Link());
+        $response = $this->get($page1->Link());
+        $this->assertEquals(200, $response->getStatusCode());
 
         // An error message will be shown if you visit it
         $content = $this->get(Director::makeRelative($page1->Link()))->getBody();
@@ -73,8 +93,13 @@ class RedirectorPageTest extends FunctionalTest
         // This also applies for internal links
         $page2 = $this->objFromFixture(RedirectorPage::class, 'badinternal');
         $this->assertEquals('/bad-internal', $page2->Link());
-        $content = $this->get(Director::makeRelative($page2->Link()))->getBody();
-        $this->assertStringContainsString('message-setupWithoutRedirect', $content);
+        $response = $this->get(Director::makeRelative($page2->Link()));
+        $content = $response->getBody();
+        if ($use404) {
+            $this->assertNull($response->getBody());
+        } else {
+            $this->assertStringContainsString('message-setupWithoutRedirect', $content);
+        }
     }
 
     public function testReflexiveAndTransitiveInternalRedirectors()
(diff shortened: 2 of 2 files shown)


emteknetnz:
Another option for this is that we have a configurable private static along the lines of '404 on fail' that is false by default to retain the currently behaviour.  They would make it easier for other developers to change to this behaviour.

I think this may be a better option because I cannot think of any other useful behaviour that would be implemented in the extension hook.

code/Model/RedirectorPageController.php
```diff
@@ -24,7 +24,12 @@ public function index(HTTPRequest $request)
         if (!$this->getResponse()->isFinished() && $link = $page->redirectionLink()) {
             $this->redirect($link, 301);
         }
-        return parent::handleAction($request, 'handleIndex');
+
+        $response = parent::handleAction($request, 'handleIndex');
```
emteknetnz, line 28: Note to myself that having a `parent::` call in the middle of a function isn't completely unheard of https://github.com/silverstripe/silverstripe-framework/blob/4/src/Security/Member.php#L1705

emteknetnz:
I'm honestly finding the approach in this PR a bit weird

The existing logic on this function is:
1. attempt to redirect with a 301 status code
2. if that doesn't work, treat this like a normal page.  Since $page->Content is assumed to be blank, use the getContent() function on this with a 200 status code.

The reason for PR being raised as I understand it was because a cms author had changed a previous regular page into a redirector page, so it retained the old $page->Content (which they're no longer able to edit because that FormField is not shown in the CMS)

Repurposing an old page seems like a fairly reasonable thing for a CMS author to do with old content.  It became a problem when the page being redirected to was archived and the link no longer worked

I really don't think that adding an extension hook and attempting to fix this with custom code on a case by case basis is really the right solution here.

A 404 is also potentially wrong here since that should be returned for the URL of the archived target page, though the redirector page should still return a 301/302.  Unfortunately we no longer know what the target URL should be as the target page and corresponding URLSegment has been archived.

A redirect to the home page is probably not a bad default for Live pages where redirection is not working, though Draft previews should show the contents of getContent so that CMS authors get feedback that they need to do something

@clarkepaul do you have any views on this?

michalkleiner:
> I really don't think that adding an extension hook and attempting to fix this with custom code on a case by case basis is pretty sub-optimal

_do_ or _don't_ think it's sub-optimal?

From my perspective, a redirector page shouldn't really hold any content, so if the destination page is not found, return standard 404 using the error page if present. If the destination does exist, redirect with 301/302. Nothing more, nothing less.

clarkepaul:
I think a 404 page is better than sending them to the home page. 

I don't think the redirector page should be triggered to be unpublished unless the CMS user has opted into it.

I think we need to empower so "CMS authors get feedback that they need to do something", for instance when unpublishing they get a warning that other pages are linked to it prior to unpublish/archiving. This is probably a separate but related issue, reports page related, or reason to look into CMS notifications.

ssmarco:
I like your quote @clarkepaul about empowering CMS authors and thus makes sense to empower Devs as well if CMS authors needs help in this case as well as for backwards compatibility. This is the mean reason I think of extension hook which gives Devs flexibility whether the old behaviour is correct or wrong.

ssmarco:
@emteknetnz @michalkleiner @clarkepaul Thanks for the feedback. Have updated my pull request taking points from all.

tests/php/Model/RedirectorPageTest.php
```diff
@@ -117,4 +124,25 @@ public function testNoJSLinksAllowed()
         $page->write();
         $this->assertEmpty($page->ExternalURL);
     }
+
+    public function testOnMissingLinkResponse()
+    {
+        $page = $this->objFromFixture(RedirectorPage::class, 'goodinternal');
+        $this->assertEquals('/redirection-dest/', $page->Link());
+        $response = $this->get($page->Link());
```
dhensby, line 132: I note that other tests use `Director::makeRelative()` with this call, you may want to copy that

tests/php/Model/RedirectorPageTest.php
```diff
@@ -117,4 +124,25 @@ public function testNoJSLinksAllowed()
...
+        $this->assertEquals(200, $response->getStatusCode());
+
+        // Override default response code to 400
+        Config::forClass(RedirectorPageController::class)->set('response_code', 404);
+
+        // Unpublish the target page of this redirector page.
+        $targetPage = $this->objFromFixture(Page::class, 'dest');
```
dhensby, line 139: there is no test here for setting a response code and having a published page that responds successfully.

tests/php/Model/RedirectorPageTest.php
```diff
@@ -117,4 +124,25 @@ public function testNoJSLinksAllowed()
...
+        $targetPage->doUnpublish();
+
+        $response = $this->get($page->Link());
+        $this->assertEquals(404, $response->getStatusCode());
+
+        // Reverts configuration changes
+        Config::forClass(RedirectorPageController::class)->set('response_code', 200);
```
dhensby, line 146: this shouldn't be needed as all the tests are wrapped in `Config::nest()/unnest()`

dhensby (changes requested):

dhensby:
I agree with the overall approach that if a redirector page points to a missing page, then it should return 404.

Also we should provide a clear/easy way for these dependent pages to be found and/or automatically unpublished

code/Model/RedirectorPageController.php
```diff
@@ -24,6 +30,11 @@ public function index(HTTPRequest $request)
         if (!$this->getResponse()->isFinished() && $link = $page->redirectionLink()) {
             $this->redirect($link, 301);
```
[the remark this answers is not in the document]
ssmarco: Cheers, will update this one as well

code/Model/RedirectorPageController.php
```diff
@@ -11,6 +11,12 @@ class RedirectorPageController extends PageController
 {
     private static $allowed_actions = ['index'];
 
+    /**
+     * For backwards compatibility
+     * @var int http response code
+     */
+    private static $response_code = 200;
```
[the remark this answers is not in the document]
ssmarco: Nice one, will update. I like HTTP 410

code/Model/RedirectorPageController.php
ssmarco: That's a good example :)

code/Model/RedirectorPageController.php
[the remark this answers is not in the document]
ssmarco: Unfortunately there is not 410 error template so I stick with 404

dhensby:
My suggested approach to this would be:

1. Get a fix in now that leads RedirectorPages that don't point anywhere to 404.
2. Create a task / issue for getting a better CMS user experience for automatic unpublishing or flagging of dependant pages when a page is unpublished

tests/php/Model/RedirectorPageTest.php
```diff
@@ -18,7 +21,7 @@ public function setUp()
     {
         parent::setUp();
         Director::config()->update('alternate_base_url', 'http://www.mysite.com/');
-
+        Config::modify()->set(RedirectorPageController::class, 'should_respond_404', false);
```
michalkleiner, line 24: Probably not necessary to add this as it's the default.

code/Model/RedirectorPageController.php
```diff
@@ -11,19 +12,28 @@ class RedirectorPageController extends PageController
...
+     * @throws HTTPResponse_Exception
      */
     public function index(HTTPRequest $request)
     {
         /** @var RedirectorPage $page */
         $page = $this->data();
         if (!$this->getResponse()->isFinished() && $link = $page->redirectionLink()) {
-            $this->redirect($link, 301);
+            return $this->redirect($link, 301);
+        } elseif ($page->LinkToID && $this->config()->get('should_respond_404') === true) {
```
michalkleiner, line 33: This condition is a bit cryptic for me @ssmarco, can we refactor it so it's cleaner what/why it's testing? I would expect to test something like `!$page->LinkTo()->exists()` here.

michalkleiner: Or can we add a comment why we're not doing that and whether it needs to check for something else or relying on the previous condition etc.

GuySartorelli:
I've retargetted this to `5` and rebased, updated to current best practices and made the requested changes.

One test is still failing but I ran out of time to look into it - I'll look at it in the new year but @ssmarco this is still your PR, so if you want to keep working on it and fix that failing test before I get to it that'd be amazing.

code/Model/RedirectorPageController.php
```diff
@@ -11,19 +12,38 @@ class RedirectorPageController extends PageController
...
-            $this->redirect($link, 301);
+            return $this->redirect($link, 301);
         }
+
+        // Respond with 404 if redirecting to a missing file or page
+        if (($this->RedirectionType === 'Internal' && !$page->LinkTo()?->exists())
+            || ($this->RedirectionType === 'File' && !$page->LinkToFile()?->exists())
+        ) {
+            if (static::config()->get('missing_redirect_is_404')) {
+                return $this->httpError(404);
```
emteknetnz:
```suggestion
                $this->httpError(404);
```
RequestHandler::httpError() throws an exception, doesn't return anything

emteknetnz (changes requested):
Could you also add a quick unit test for the scenario for redirecting to a file

code/Model/RedirectorPageController.php
ChloeSartorelli: I don't think this really matters either way - but since I need to touch this PR anyway to add an extra test I'll also change this.
I personally like the `return` because it makes it super clear to the reader that this line will stop executing the rest of the method, but it's not really important.

GuySartorelli:
Added test scenario for files to the existing test via the dataprovider.

code/Model/RedirectorPageController.php
ChloeSartorelli: Done

emteknetnz (approved):
Tested locally, works good

PR #2666, opened 2021-06-16
NEW: getByLink plugin allows arrays of links

unclecheese:
Issue https://github.com/silverstripe/silverstripe-cms/issues/2674

Deserving of a refactor down the track, because this is is really inefficient, but it's a legitimate case for headless apps that want to bulk fetch pages.

### DIFF
--- a/code/GraphQL/LinkablePlugin.php
+++ b/code/GraphQL/LinkablePlugin.php
@@ -64,8 +64,8 @@ class LinkablePlugin implements ModelQueryPlugin
             return;
         }
         $singleFieldName = $this->config()->get('field_name');
-        $pluralFieldName = $this->config()->get('list_field_name');
-        $fieldName = $query->isList() ? $pluralFieldName : $singleFieldName;
+        $listFieldName = $this->config()->get('list_field_name');
+        $fieldName = $query->isList() ? $listFieldName : $singleFieldName;
         $type = $query->isList() ? '[String]' : 'String';
         $query->addArg($fieldName, $type);
         $query->addResolverAfterware(
@@ -80,9 +80,9 @@ class LinkablePlugin implements ModelQueryPlugin
     public static function applyLinkFilter($obj, array $args, array $context, ResolveInfo $info)
     {
         $singleFieldName = static::config()->get('field_name');
-        $pluralFieldName = static::config()->get('list_field_name');
-        $filterLink = $args['filter'][$singleFieldName] ?? ($args['filter'][$pluralFieldName] ?? null);
-        $argLink = $args[$singleFieldName] ?? ($args[$pluralFieldName] ?? null);
+        $listFieldName = static::config()->get('list_field_name');
+        $filterLink = $args['filter'][$singleFieldName] ?? ($args['filter'][$listFieldName] ?? null);
+        $argLink = $args[$singleFieldName] ?? ($args[$listFieldName] ?? null);
         $linkData = $filterLink ?: $argLink;
         if (!$linkData) {
             return $obj;
--- a/code/GraphQL/LinkablePlugin.php
+++ b/code/GraphQL/LinkablePlugin.php
@@ -12,6 +12,7 @@ use SilverStripe\GraphQL\Schema\Exception\SchemaBuilderException;
 use SilverStripe\GraphQL\Schema\Field\ModelQuery;
 use SilverStripe\GraphQL\Schema\Interfaces\ModelQueryPlugin;
 use SilverStripe\GraphQL\Schema\Schema;
+use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataList;
 
 if (!interface_exists(ModelQueryPlugin::class)) {
@@ -30,6 +31,11 @@ class LinkablePlugin implements ModelQueryPlugin
      * @config
      */
     private static $field_name = 'link';
+
+    /**
+     * @var array
+     */
+    private static $resolver = [__CLASS__, 'applyLinkFilter'];
 
     /**
      * @return string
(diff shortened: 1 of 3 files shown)


code/GraphQL/LinkablePlugin.php
```diff
@@ -31,6 +32,11 @@ class LinkablePlugin implements ModelQueryPlugin
      */
     private static $field_name = 'link';
 
+    /**
+     * @var array
+     */
+    private static $resolver = [__CLASS__, 'applyLinkFilter'];
```
emteknetnz, line 38: I don't think I've seen a private static written like this before.  Does this even need to be a private static?  Is it something we'd ever expect a project or module to change the configuration for?

Otherwise the old `$query->addResolverAfterware([static::class, 'applyLinkFilter']);` seems appropriate

code/GraphQL/LinkablePlugin.php
```diff
@@ -52,43 +58,44 @@ public function apply(ModelQuery $query, Schema $schema, array $config = []): vo
...
-        array $args,
-        array $context,
-        ResolveInfo $info,
-        callable $done
-    ) {
-        $fieldName = static::config()->get('field_name');
-        $filterLink = $args['filter'][$fieldName] ?? null;
-        $argLink = $args[$fieldName] ?? null;
-        $filterLink = $filterLink ?: $argLink;
+    public static function applyLinkFilter(array $context): callable
```
emteknetnz, line 76: The function signature has totally changed, under semver this requires a new major release which we obviously cannot do.

code/GraphQL/LinkablePlugin.php
```diff
@@ -52,43 +58,44 @@ public function apply(ModelQuery $query, Schema $schema, array $config = []): vo
         if ($class !== SiteTree::class && !is_subclass_of($class, SiteTree::class)) {
             return;
         }
-        Schema::invariant(
```
emteknetnz, line 55: With this removed does this still function still `@throws SchemaBuilderException` ?

emteknetnz (changes requested):
Need to do something about the function signature change of `applyLinkFilter()`

Also requires unit tests

code/GraphQL/LinkablePlugin.php
unclecheese: Ah, yup. Good catch.

code/GraphQL/LinkablePlugin.php
unclecheese: There is a 0.000000000000 chance that anyone is using this in the user space, but I've obeyed the semver gods and refactored this to be much uglier but backward compatible nonetheless. :)

It's probably worth discussing at some point that as soon as you add context to a resolver, it has to change its method signature, and that core resolvers should therefore be marked `@internal`. There's no good reason to be invoking those methods outside of a graphql request.

code/GraphQL/LinkablePlugin.php
unclecheese: You can pass in a custom resolver through the plugin config. Otherwise, it just falls back on the one that's assigned through static class config. 

You need `__CLASS__` because late static binding (`static::class`) isn't available at the time the static is defined.

code/GraphQL/LinkablePlugin.php
```diff
@@ -52,43 +63,42 @@ public function apply(ModelQuery $query, Schema $schema, array $config = []): vo
...
-            !$query->isList(),
-            'Plugin %s cannot be applied to queries that return lists. Query "%s" is a list',
-            static::getIdentifier(),
-            $query->getName()
+        $singleFieldName = $this->config()->get('field_name');
+        $pluralFieldName = $this->config()->get('list_field_name');
```
emteknetnz, line 67: `$pluralFieldName` should be `$listFieldName` to match the static field name

code/GraphQL/LinkablePlugin.php
```diff
@@ -52,43 +63,42 @@ public function apply(ModelQuery $query, Schema $schema, array $config = []): vo
...
-        $argLink = $args[$fieldName] ?? null;
-        $filterLink = $filterLink ?: $argLink;
-
-        if ($filterLink) {
-            $done();
-            return SiteTree::get_by_link($filterLink);
+    public static function applyLinkFilter($obj, array $args, array $context, ResolveInfo $info)
+    {
+        $singleFieldName = static::config()->get('field_name');
+        $pluralFieldName = static::config()->get('list_field_name');
```
emteknetnz, line 83: `$pluralFieldName` should be `$listFieldName` to match the static field name

emteknetnz (changes requested):
Approach seems fine

Minor comment about removing the duplicate 'plural' naming and only use 'list'

Will the graphqli explorer auto update to show that a 'links' filter is available?

Needs unit tests

code/GraphQL/LinkablePlugin.php
```diff
@@ -52,43 +63,42 @@ public function apply(ModelQuery $query, Schema $schema, array $config = []): vo
...
-    ) {
-        $fieldName = static::config()->get('field_name');
-        $filterLink = $args['filter'][$fieldName] ?? null;
-        $argLink = $args[$fieldName] ?? null;
-        $filterLink = $filterLink ?: $argLink;
-
-        if ($filterLink) {
-            $done();
-            return SiteTree::get_by_link($filterLink);
+    public static function applyLinkFilter($obj, array $args, array $context, ResolveInfo $info)
```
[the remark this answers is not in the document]
unclecheese: Apparently not. This type of stuff is all over the module. The graphql-php library will always pass four arguments to your resolvers (the graphql standard of `$obj, $args, $context, $info`), but if you only need the first one, that's fine.

Screen Shot 2021-09-06 at 16 59 26

With regard to the stack trace, yeah it's confusing. See in the apply() method that it adds `addResolverAfterware` that points to the resolver callable. Lots of magic like that in graphql, unfortunately. It's just a consequence of everything needing to be static.

PR #2675, opened 2021-06-29
[WIP] API Enable support for extensions to get anchor links

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/796

Related PR https://github.com/silverstripe/silverstripe-elemental/pull/911

Note: seems like the AnchorSelectField javascript only currently works now because something is a bit broken and prevents the response get cached in the redux state

`props.loadingState === anchorSelectorStates.DIRTY` for regular pages

For BlockPages, its always `anchorSelectorStates.SUCCESS`, so the cached "response" of an empty array is used.  Now response is actually made though, I think it's just the initial state.

I think all this may be because Elemental is more 'plugged into' the react/redux/apollo way of doing this that it "works"

Simply removing the caching allowed a BlockPage to actually fetch from the endpoint

There's currently some [code](https://github.com/silverstripe/silverstripe-cms/blob/4/client/src/components/AnchorSelectorField/AnchorSelectorField.js#L169) in AnchorSelectField.mapStateToProps that looks little non-standard?  I'm somewhat inclined to just remove it all. Anchor links on pages are currently not cached, and this is causing no real issues.

Previously the BlockPage was trying to fetch from `/admin/methodSchema/Modals/editorAnchorLink/6`

The correct URL is `/admin/Modals/editorAnchorLink/6/field/Anchor/anchors/2`


GuySartorelli:
Closing this in favour of #2741
There was a lot going on in this PR that I don't have time to unpick - it's possible some of the additional changes are enhancements that make sense to make - but #2741 resolves the main issue this PR was for, I believe.

PR #2676, opened 2021-07-16
PATCH: docblock for return value fixed for get_by_link

sunnysideup:


dhensby (approved):
FYI - these changes are not `PATCH` (indicates a bug fix) but are `DOCS` as they are documentation improvements

PR #2677, opened 2021-07-20
add a new extension hook to SiteTree::RelativeLink()

xini:
The extension hook on https://github.com/silverstripe/silverstripe-cms/blob/4/code/Model/SiteTree.php#L661 (`updateRelativeLink`) is called before a trailing slash is added to page links without action (see #2631). 

To give a dev the option to remove the trailing slash we need an extension hook after the `join_links`, but I don't want to move the existing one because of the different parameters. But I'm happy to change that if requested.

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -666,7 +666,11 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             $action = null;
         }
 
-        return Controller::join_links($base, '/', $action);
+        $link = Controller::join_links($base, '/', $action);
+
+        $this->extend('updateFinalRelativeLink', $link);
+
+        return $link;
     }
 
     /**


dhensby:
Given the overheads of extension and the significant use of `RelativeLink()` I'm really anxious about changing this - especially if the use-case is as basic as removing trailing slashes - the overhead to inflict on all users so that some can worry about aesthetics of URLs is not a trade-off I'd be willing to support.

dhensby (changes requested):
As per my comment - I'd rather change `join_links` to strip trailing slashes than do this, and I'm against that too.

xini:
@dhensby So what do you suggest I do? 

I agree that `join_links` should clean it up. But since `SiteTree:RelativeLink()` is the *only* method in the whole of SS that adds a trailing slash, changing `join_links` would be overhead?

You've closed #2631 which would solve the issue altogether?

All I'm asking here is an extension hook so that I can solve the problem of trailing slashes in an external module...

dhensby:
> You've closed #2631 which would solve the issue altogether?

But what is the issue exactly?

> All I'm asking here is an extension hook so that I can solve the problem of trailing slashes in an external module...

sure, but with a huge overhead for no actual gain. I'm not sure changing `join_links` (to not add a trailing slash) provides any overhead, not compared to adding an extension hook

xini:
> But what is the issue exactly?

The issue is that `SiteTree` creates inconsistent links, adding a trailing slash if there is no action. Nowhere else in Silverstripe a trailing slash is added anywhere. For consistency, it would be nice if we could align link generation to one common format.

> sure, but with a huge overhead for no actual gain.

The overhead is not that huge. There are 142 extension hooks in framework alone.

I would much prefer to move the existing hook in `RelativeLink` down and pass it the full link instead of adding another hook. This would keep the number of extension hooks and overhead the same, but it would change the API.

The gain would be consistency.

>  I'm not sure changing join_links (to not add a trailing slash) provides any overhead, not compared to adding an extension hook

I agree that changing `join_links` would be the cleaner option. I had suggested that change in #2631: 

```
diff --git a/src/Control/Controller.php b/src/Control/Controller.php
index 09e8e7e1a..2b8eaadbe 100644
--- a/src/Control/Controller.php
+++ b/src/Control/Controller.php
@@ -690,10 +690,10 @@ class Controller extends RequestHandler implements TemplateGlobalProvider
             }
             if ((is_string($arg) && $arg) || is_numeric($arg)) {
                 $arg = (string) $arg;
-                if ($result && substr($result, -1) != '/' && $arg[0] != '/') {
-                    $result .= "/$arg";
+                if (strlen($result) > 0) {
+                    $result .= '/' . trim($arg, '/');
                 } else {
-                    $result .= (substr($result, -1) == '/' && $arg[0] == '/') ? ltrim($arg, '/') : $arg;
+                    $result .= rtrim($arg, '/');
                 }
             }
         }
```
But then we are back where we ended up in #2631 with changed functionality and tests etc.

You said yourself in https://github.com/silverstripe/silverstripe-framework/issues/8882#issuecomment-760795629 that we should have more consistent URL generation, but also, that this should be solved outside the core. 

So please, give me hook to do so, even if you consider this an unimportant vanity issue. Thanks.

dhensby:
> The issue is that `SiteTree` creates inconsistent links, adding a trailing slash if there is no action. Nowhere else in Silverstripe a trailing slash is added anywhere. For consistency, it would be nice if we could align link generation to one common format.
> 
> >

OK - and I'm saying fixing that would be preferable to adding a second extension point here.

Just because there are 142 hooks already doesn't mean that adding another isn't additional overhead.

xini:
Then why did you close #2631? That PR would solve it.

sunnysideup:
Bump

GuySartorelli:
Closing in favor of #2770

Issue #2680, opened 2021-08-15
500 errors produced by SiteTree with url segment "0"


shanholmes:
**To reproduce:**
Hit your Silverstripe site with the route` /0/anything`. The `anything` in the URL can be literally any value but the URL must contain `0` as the first parameter.

https://www.loom.com/share/31c8626c785a432abd0adb4d0d312d68

**Expected behaviour:**
If the URL does not exist within the CMS or is not valid, the site should return a 404 error.

**Current behaviour:**
The site returns a 500 error.

**Environment:**
Tested with Silverstripe 4.8 (PHP 7.4), Silverstripe 3.7.3 (PHP 7.3).

## PRs
- https://github.com/silverstripe/silverstripe-cms/pull/2825

maxime-rainville:
The problem is that on this line, we assigned the URL segment to a variable and we check that if it's falsy. We need to do a strict check instead for `null`.

https://github.com/silverstripe/silverstripe-cms/blob/f892205b9bf6263e6b699c84fda13536860e36e1/code/Controllers/ModelAsController.php#L122

There's a potential follow up problem that the CMS won't let you assign `0` as a URL segment to a page. It's happy with `1`

Zazama:
We've been getting many bot requests lately that call '/0/...' URLs and cause 500 errors. I've seen that there was a PR (https://github.com/silverstripe/silverstripe-cms/pull/2681) but it has not been merged.

Is there something that was missing in the PR or would you accept a new one that checks if the value is null instead of falsy?
It would be cool to get rid of those errors without resorting to the webserver config.

shanholmes:
Yeah we have had a bit of a flood recently as well. Let me know if there is anything I can do to help.

shanholmes:
This is still an issue and has since been carried over to the 5 branch of the cms.

emteknetnz:
Linked PR has been merged, this will auto-tag on the 4.13 shortly and will auto-tag on the 5.1 branch within the next week

PR #2681, opened 2021-08-15
Support `0` url

adrhumphreys:
Should resolve: https://github.com/silverstripe/silverstripe-cms/issues/2680

I've kept the checks as broad as possible to ensure we don't get snagged but something unexpected


GuySartorelli:
> Just thought I'd note just in case: If unit tests end up being added for this, I've added a fixture for a page with a urlsegment '0' in #2685.

I've pulled the '0' segment support stuff out of that PR now. Looks like more work is needed upstream (which is out of scope for that PR) before `SiteTree::get_by_link` support can be added.

PR #2682, opened 2021-08-18
MNT Use logInAs() for unit tests

emteknetnz:
Issue https://github.com/silverstripe/recipe-cms/issues/47


maxime-rainville (approved):
Looks good. Probably worth mentioning that some of the `logOut` calls are not necessary as the tests should implicitly log you out anyway in between calls.

PR #2683, opened 2021-08-30
Make Batch Actions button look like a button

andrewandante:
It's really not immediately obvious that it is clickable - not certain `-info` is best but it's an improvement.

image

### DIFF
--- a/templates/SilverStripe/CMS/Controllers/Includes/CMSPagesController_ContentToolActions.ss
+++ b/templates/SilverStripe/CMS/Controllers/Includes/CMSPagesController_ContentToolActions.ss
@@ -4,7 +4,7 @@
             <a class="btn btn-primary cms-content-addpage-button tool-button font-icon-plus" href="$LinkPageAdd" data-url-addpage="{$LinkPageAdd('', 'ParentID=%s')}"><%t SilverStripe\CMS\Controllers\CMSMain.AddNewButton 'Add new' %></a>
 
             <% if $View == 'Tree' %>
-            <button type="button" class="cms-content-batchactions-button btn btn-secondary tool-button font-icon-check-mark-2 btn--last" data-toolid="batch-actions">
+            <button type="button" class="cms-content-batchactions-button btn btn-info tool-button font-icon-check-mark-2 btn--last" data-toolid="batch-actions">
                 <%t SilverStripe\CMS\Controllers\CMSPageHistoryController.MULTISELECT "Batch actions" %>
             </button>
             <% end_if %>


brynwhyman:
Nice! cc @silverstripeux :)

michalkleiner:
I wanted to have a look at how the secondary button is meant to look, as that might be the primary issue here the button doesn't look as expected because of the secondary button styling issue. Semantically it's not really an info button but I get it's the best that might be available at the moment.

michalkleiner:
Secondary with an outline might be better, but can't find what CSS class to use, may not be implemented?
https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5aa97977f21f2800116494f5

michalkleiner:
Default state
normal
Hover state
hover

andrewandante:
> Secondary with an outline might be better `.btn-outline-secondary`.
> https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5aa97977f21f2800116494f5

That looks like this: 

image

Not sure I prefer it, it's definitely an improvement on current. Not sure, feel like "action" is the most right in terms of functionality but it feels weird to have double green there.

Outline primary:

image


Double primary:

image


I dunno!

michalkleiner:
Yeah, I get the sentiment. Let's see what UX says.

clarkepaul:
I don't think its the fact that it doesn't look like its a button is the problem, but rather its the wrong UI for the action.

In the files area we decided not to follow the pages pattern as it wasn't the right interaction (not obvious) for whats happening. The interaction is more like show me some more options. From memory my favoured approach was to make the full page tree view just show the checkboxes all the time (along with every option) then there was no need for this weird interaction. On the smaller tree view there would be a link to edit tree or view more options (hope that makes sense) which leads to the full editable tree.

Currently we use secondary as the most common looking button to minimise the impact of all the actions available to the user. Basically if they were all looking like buttons the UI would be overflowing with them competing for attention. We are trying to make each area only have 1-2 max main actions.

I think I'd rather try to use a switch type element if we are looking for a quick fix.

andrewandante:
> I think the fact that it doesn't look like its a button isn't the problem, but rather its the wrong UI for the action.

That's absolutely true, I think - but as you say that's a larger piece of work. For some context, this came about because today I was doing a CMS training session over Zoom and the client had no idea that was a clickable item. Same problem came about with the tabs inside a Model Admin but that's a separate thing.

Is it worth making a minor improvement here, with a longer-term goal of a larger refactor? Secondary-with-outline might be a good halfway house of maintaining the "clean" look while still making it more obvious that it is clickable.

clarkepaul:
Was this button perhaps the first part of the UI that was introduced after the "Add page" button, or first instance of that style of button being seen? The reason I ask is that I think once more of the CMS is seen, the user develops more of an understanding about the areas for actions and other areas for content editing. 

I'm not comfortable to introduce this pattern consistently for buttons right of the main action (e.g. actions in the Security & Files area). To single out this one action doesn't seem right either, as it isn't an action that needs to be more prominent than other actions (this action is more likely for someone familiar with the CMS and comfortable making dramatic changes).

Its common place now for UIs with a lot of actions to minimise the styling as long as they are in areas which are clear to hold actions. I think this is more of a general CMS wide 'thing' rather than this one button which is better solved by a better defined areas, better labelling and better association/use of icons. 

> Is it worth making a minor improvement here, with a longer-term goal of a larger refactor? Secondary-with-outline might be a good halfway house of maintaining the "clean" look while still making it more obvious that it is clickable.

Sorry to say I have a different view.

In my mind changing the label and icon to something like "Edit tree" could be a better option.

clarkepaul:
For example, "Edit" being in the label makes it more obvious it is an action. 
image

Also there is an additional blank space at the beginning of the label "Batch actions" which makes a bigger gap between the label and icon.
Pasted_Image_1_09_21__4_24_PM

andrewandante:
> Sorry to say I have a different view.

No problem, that's why we have these discussions! I think it might have been the first of that button type that the client had their attention drawn to, they had been using the CMS a while though.

I like the mockup for `Edit Tree`, definitely implies a clickable "action" more than "Batch Actions" does, for whatever reason.

clarkepaul:
@andrewandante these types of fixes rarely make the priority list so really appreciate the intention. If you have UX type fixes, feel free to hit me up to see if we can align or maybe I have some thoughts already. I have user tested batch actions many moons ago and it [returned really bad results](https://www.silverstripe.org/assets/2014/Reports/Usability-report-on-SilverStripe-CMS-Main-actions-and-page-states.pdf) (page 5). Also good to see how v3 looked, I remember we opted to change the component because it didn't align to any other experience, unfortunately the replacement wasn't any better really.

andrewandante:
Hah that document is a blast from the past for sure! I'm going to make some sort of change in this area for an upcoming project, it will be good to see what the client says as we try options on, and we can use that feedback to inform where we aim to go from here.

clarkepaul:
@andrewandante nice one :), happy to help e.g. if you need alternative mockups etc.

GuySartorelli:
I'm going to close this PR - it sounds based on the discussion that the current PR is not suitable as a resolution to this UX problem. We also don't have a dedicated product UX team right now so it's unlikely this will get any attention in the near future.

PR #2685, opened 2021-09-21
FIX: Treat absolute links in a consistent manner in get_by_link.

### LINKED ISSUE #2580: SiteTree::get_by_link returns null instead of the homepage if passed the homepage absolute URL
Passing the homepage's absolute URL (excluding the home path e.g. `https://example.co.nz/`, not `https://example.co.nz/home/`) to `SiteTree::get_by_link` returns null.
I believe this should return the SiteTree object which represents the home page.

silverstripe-cms version: 4.5.1

GuySartorelli:
Fixes #2580 

The call to `Director::makeRelative` transforms absolute links into relative links, but is done separately to checking for a possible homepage URL. Previously, this meant that you could pass in "https://example.co.nz/about-us" or "about-us" and get the same result, but passing in "https://example.co.nz/" and "/"  would give _different_ results.
This commit performs the transformation to a relative link _before_ checking if the path should be for the home page, which leads to more consistent results.

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -415,9 +415,8 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         $urlSegmentExpr = sprintf('"%s"."URLSegment"', $tableName);
         $parentIDExpr = sprintf('"%s"."ParentID"', $tableName);
 
-        if (trim($link, '/')) {
-            $link = trim(Director::makeRelative($link), '/');
-        } else {
+        $link = trim(Director::makeRelative($link), '/');
+        if (!$link) {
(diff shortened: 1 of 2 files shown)


GuySartorelli:
> Seems like this could use a unit test or two

I'll put that in my to-do list. I don't know when I'll get around to it though, I'm quite a bit busier than when I raised the PR.

GuySartorelli:
It looks like more work is required to allow "0" routes in the first place that needs to be addressed separately, so I've removed the "0" route support from this PR - it probably doesn't really belong here anyway.

GuySartorelli:
@emteknetnz Finally got this in a good state. Please review.

emteknetnz (approved):

PR #2686, opened 2021-09-23
FIX updateAnchorsOnPage extension not always called

kevin-hine-innis:
Currently, if there are no anchors found in the page content, the updateAnchorsOnPage extension is not called.
This proposed change will call the updateAnchorsOnPage extension regardless of the page content.

PR #2687, opened 2021-09-23
SiteTreeLinkTracking_Parser should use the getAnchorsOnPage

kevin-hine-innis:
Currently, the SiteTreeLinkTracking_Parser only checks the page's content for anchors.
As a result any anchors that have been added or modified by the updateAnchorsOnPage extension in the getAnchorsOnPage method are marked ss-broken.


michalkleiner:
I suspect this can only be properly tested once https://github.com/silverstripe/silverstripe-cms/pull/2686 is resolved, but could you add a unit test where you'd update the list of anchors via the extension point?

kevin-hine-innis:
Okay, I think I did this correctly now, and I added additional unit tests as suggested.

I felt it would be better to pull the project and recreate the pull request. Sorry about that.
Please see #2697

PR #2688, opened 2021-09-23
FIX canonical tag on VirtualPage

xini:
The canonical tag should be a `link` tag not `meta`, see https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls#rel-canonical-link-method

### DIFF
--- a/code/Model/VirtualPage.php
+++ b/code/Model/VirtualPage.php
@@ -143,9 +143,10 @@ class VirtualPage extends Page
         $copied = $this->CopyContentFrom();
         if ($copied && $copied->exists()) {
             $tags['canonical'] = [
+                'tag' => 'link',
                 'attributes' => [
                     'rel' => 'canonical',
-                    'content' => $copied->AbsoluteLink(),
+                    'href' => $copied->AbsoluteLink(),
                 ],
             ];
         }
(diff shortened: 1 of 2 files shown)


michalkleiner:
Thanks @xini. Did you check if there's an existing unit test we can update as well or there's nothing covering this at this stage for the VirtualPage or any page in general?

xini:
@michalkleiner I haven't found any tests covering this, no.

michalkleiner:
Thanks @xini, yeah, it looks like there's nothing current. There's a test class for the whole VirtualPage, so I think we could take a subset from https://github.com/silverstripe/silverstripe-cms/blob/4/tests/php/Model/SiteTreeTest.php#L1389 and add it into a new test in https://github.com/silverstripe/silverstripe-cms/blob/4/tests/php/Model/VirtualPageTest.php#L19, potentially testing other expected headers as well, such as that the value of the canonical should point to the referenced page.

Do you think you can look into that, please?

dhensby (approved):
This change looks good. Whilst a test would be nice, we shouldn't block a fix waiting for a test if there is currently no test coverage anyway.

michalkleiner:
I was of two minds on that too, @dhensby. Let's give @xini a chance if he wants to/can add the test, and merge without it if not.

dhensby:
Thanks for the test @xini - I think there's some linting (tabs/spaces being mixed) to sort out in the test

xini:
@dhensby I noticed, thanks. Fixed now.

michalkleiner:
Thanks @xini, great work. Test failure is unrelated.

xini:
Thanks @michalkleiner

PR #2692, opened 2021-10-07
NEW Add preview state for non-versioned DataObjects.

GuySartorelli:
Currently only Versioned DataObjects are enabled for the CMS preview. Without a navigator item, non-versioned DataObjects aren't allowed to display in the preview panel even if they implement CMSPreviewable.

image

Note that this state will still need to be added to the list of allowed states in the .cms-preview entwine logic which can be done like so:
```js
$('.cms-preview').entwine('ss.preview').AllowedStates().push('Unversioned');
```
Allowing the state mannualy as above is unnecessary if silverstripe/silverstripe-admin#1258 gets merged in.

This isn't particularly useful without silverstripe/silverstripe-admin#1250 but I guess it would allow someone to preview non-versioned `SiteTree` objects (assuming it's even possible to remove the `Versioned` extension from `SiteTree` without causing issues) or to use non-versioned `DataObject`s in a custom `CMSMain` subclass controlller.

### DIFF
--- a/code/Controllers/SilverStripeNavigatorItem_Unversioned.php
+++ b/code/Controllers/SilverStripeNavigatorItem_Unversioned.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace SilverStripe\CMS\Controllers;
+
+use SilverStripe\CMS\Controllers\SilverStripeNavigatorItem;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Convert;
+use SilverStripe\Security\Member;
+use SilverStripe\Versioned\Versioned;
+
+class SilverStripeNavigatorItem_Unversioned extends SilverStripeNavigatorItem
+{
+    public function getHTML()
+    {
+        $recordLink = Convert::raw2att($this->getLink());
+        $linkTitle = _t('SilverStripe\\CMS\\Controllers\\ContentController.UNVERSIONEDPREVIEW', 'Preview');
+        return "<a class=\"current\" href=\"$recordLink\">$linkTitle</a>";
+    }
+
+    public function getLink()
+    {
+        return $this->getRecord()->PreviewLink();
+    }
+
+    public function getTitle()
+    {
+        return  _t(
+            'SilverStripe\\CMS\\Controllers\\ContentController.UNVERSIONEDPREVIEW',
+            'Preview',
+            'Used for the Switch between states (if any other other states are added). Needs to be a short label'
+        );
+    }
+
+    /**
+     * True if the record doesn't have the Versioned extension and is configured to display this item.
+     *
+     * @param Member $member
+     * @return bool
+     */
+    public function canView($member = null)
+    {
+        return (
+            !$this->getRecord()->hasExtension(Versioned::class)
+            && $this->showUnversionedLink()
+        );
+    }
+
+    /**
+     * True if the record is configured to display this item.
+     *
+     * @return bool
+     */
+    public function showUnversionedLink(): bool
+    {
+        return (bool) Config::inst()->get(get_class($this->record), 'show_unversioned_preview_link');
+    }
+
+    /**
+     * This item is always active, as there are unlikely to be other preview states available for the record.
+     *
+     * @return bool
+     */
+    public function isActive()
+    {
+        return true;
+    }
+}
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -134,6 +134,7 @@ en:
     Password: Password
     PostInstallTutorialIntro: 'This website is a simplistic version of a SilverStripe 3 site. To extend this, please take a look at {link}.'
     StartEditing: 'You can start editing your content by opening <a href="{link}">the CMS</a>.'
+    UNVERSIONEDPREVIEW: 'Preview'
     UnableDeleteInstall: 'Unable to delete installation files. Please delete the files below manually'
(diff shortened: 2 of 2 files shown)


GuySartorelli:
I've realised recently that this might be better as `NEW` or `API` rather than `ENH` - if a core contributor can advise which is appropriate, I'll make the change.

GuySartorelli:
Let me know if you want me to force push this with the commit message changed to `NEW`

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
```diff
@@ -0,0 +1,53 @@
...
+namespace SilverStripe\CMS\Controllers;
+
+use Config;
+use SilverStripe\CMS\Controllers\SilverStripeNavigatorItem;
+use SilverStripe\Core\Convert;
+use SilverStripe\Versioned\Versioned;
+
+class SilverStripeNavigatorItem_Unversioned extends SilverStripeNavigatorItem
+{
+    public function getHTML()
```
emteknetnz, line 12: For all of these methods, would you be able to add php dockblocks for all the `@param` + `@return`, and php7 return types, though omit php7 return types if existing parent class(es) API don't define php7 return types

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace SilverStripe\CMS\Controllers;
+
+use Config;
```
emteknetnz, line 5: I don't believe this is correct, possibly should be `use SilverStripe\Core\Config\Config;` ?

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
```diff
@@ -0,0 +1,53 @@
...
+            && $this->showUnversionedLink()
+        );
+    }
+
+    /**
+     * @return bool
+     */
+    public function showUnversionedLink()
+    {
+        return (bool)Config::inst()->get(get_class($this->record), 'show_unversioned_preview_link');
```
emteknetnz, line 46:
```suggestion
        return (bool) Config::inst()->get(get_class($this->record), 'show_unversioned_preview_link');
```

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return $this->getRecord()->PreviewLink();
+    }
+
+    public function getTitle()
+    {
+        return  _t(
+            'SilverStripe\\CMS\\Controllers\\ContentController.UNVERSIONEDPREVIEW',
+            'Preview',
+            'Used for the Switch between states (if any other other states are added). Needs to be a short label'
```
emteknetnz, line 29: What's this 3rd arg in `_t()` for?

emteknetnz (changes requested):
I haven't tested yet, though some code changes that could be made

NEW is correct

emteknetnz:
Just to confirm, in the description you've said:

```
Note that this state will still need to be added to the list of allowed states in the .cms-preview entwine logic which can be done like so:

$('.cms-preview').entwine('ss.preview').AllowedStates().push('Unversioned');
```

Would this be resolved due with this PR? https://github.com/silverstripe/silverstripe-admin/pull/1258/files

GuySartorelli:
> Would this be resolved no longer require once this PR is merged?

Yes, that's correct. The comment was made so that this PR could be merged without that one if necessary but that code is redundant if that PR gets merged.

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
GuySartorelli: I just copied what was done for the existing `SilverStripeNavigatorItem`s but the PHPDoc in `i18n::t()` says "Any other string argument after default is context for i18nTextCollector"

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
GuySartorelli: Yes! Oops.

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
GuySartorelli: Can do, though they're just copied from the superclass. I had omitted them originally because the PHPDoc will typically inherit from the superclass if omitted. Most methods in this class are overriding the methods in `SilverStripeNavigatorItem` and with the exception of `canView` I don't think there's anything to add to the PHPDoc from that class.

Let me know if you want me to add PHPDocs for inherited methods anyway and I'll do so.
In the meantime I'll add any for non-inherited methods and where I feel I have something to add.

GuySartorelli:
@emteknetnz I've made most of the requested changes - just pending your decision [about the PHPDocs for inherited methods](#discussion_r794074378).

Travis jobs failed to connect to packagist.

emteknetnz (approved):
Travis issue is nothing to worry about, some matrix jobs running PHPUNIT still ran, so I'm happy

Have manually tested using sample code at https://github.com/silverstripe/silverstripe-admin/pull/1250#issuecomment-1023709904 along with https://github.com/silverstripe/silverstripe-admin/pull/1258

PR #2696, opened 2021-10-13
Add ukrainian localisation

Petro-Ivvysoft:


brynwhyman:
Hey @Petro-Ivvysoft this is a bit of a late response, but I wanted to point out that translations can't be pulled in as pull requests, you'll want to add these strings via transifex.com which is then pulled in to the source code at each release - there's documentation on this here: https://docs.silverstripe.org/en/4/contributing/translations/#contributing-translations

I see you have done a lot of work here, so I hope it's not too late to have another go at this! 🇺🇦

ScopeyNZ:
As Bryn said, the translations are controlled externally by Transifex, which are pulled in for each release. Please feel free to contribute there.

Issue #2702, opened 2021-11-11
SiteTree::getCMSActions() produces invalid Unpublish button


gurucomkz:
## Affected Version
silverstripe-cms 4+

## Description

`SiteTree.php` has the following code
```php
FormAction::create('unpublish', _t('SiteTree.BUTTONUNPUBLISH', 'Unpublish'), 'delete')
```
This code seems to be dormant and usually superseded by VersionedGridFieldItemRequest::getFormActions() and all is fine.
Once actually called it results in throwing an exception because the third argument is meant to be a `Form` instance or `null` but a `string` is given  instead.

Most users don't have a problem with that because the code is not called unless we make a separate ModelAdmin for any type of Pages (i.e. to let the editors focus only on product pages) AND enable a plugin like [lekoala/silverstripe-cms-actions](https://github.com/lekoala/silverstripe-cms-actions) that somehow triggers the `getCMSActions()`.

### Create a CMS Section
```
<?php
namespace App;
use Page;
use SilverStripe\Admin\ModelAdmin;

class PagesAdmin extends ModelAdmin
{
    private static $url_segment = 'ppp';
    private static $menu_title = 'PPPPP';

    private static $managed_models = [
        Page::class,
    ];
}
```

### Install `lekoala/silverstripe-cms-actions`
```
composer require lekoala/silverstripe-cms-actions
```
### Open a page through that new ModelAdmin
* Navigate to /admin/ppp/ 
* Open any page on the list

gurucomkz:
I actually wonder, should those buttons be declared there anyway since `silverstrioe/versioned` provides them already.

michalkleiner:
Interesting find!
I guess it's possible to deactivate versioning even for SiteTree by removing the extension in _config.php, so I'd say they should stay.

gurucomkz:
So the solution is to just patch the line to remove the third argument?

michalkleiner:
It might be as simple as that, though I can't say that for sure. There might be more to it, or not.
It's probably also missing a test since it hasn't been caught yet, so might be about creating a failing test first and then fixing it.

gurucomkz:
You are right. The direct calls to `SiteTree::getCMSActions` are still made by the `CMSPagesController` and that part is very important.

@michalkleiner I've followed your suggestion and made a PR with 2 commits - a test and a fix: https://github.com/silverstripe/silverstripe-cms/pull/2703

PR #2703, opened 2021-11-12
FIX SiteTree::getCMSActions unpublish action

gurucomkz:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/2702


dhensby:
Would this be better in a patch release?

tests/php/Model/SiteTreeTest_CMSActions.php
```diff
@@ -0,0 +1,241 @@
...
+        SiteTreeTest_ClassB::class,
+        SiteTreeTest_ClassC::class,
+        SiteTreeTest_ClassD::class,
+        SiteTreeTest_ClassCext::class,
+        SiteTreeTest_NotRoot::class,
+        SiteTreeTest_StageStatusInherit::class,
+        SiteTreeTest_DataObject::class,
+    ];
+
+    public function testGetCMSActions()
```
emteknetnz, line 32: Could you please move `testGetCMSActions()` and `testGetCMSActionsWithoutForms()` to SiteTreeTest.php and remove SiteTreeTest_CMSActions.php.  

I know that SiteTreeTest.php is already a giant file, though adding these new tests to the existing file is consistent with our naming convention of <classname>Test.php

emteknetnz (changes requested):
Looks great, thanks for this contribution

I've updated the target branch from `4` to `4.10` as there are no API changes.  Then allows this fix to be be patch released

Just one requested change to move the new unit tests to the existing test file

tests/php/Model/SiteTreeTest_CMSActions.php
gurucomkz: Done.

emteknetnz (approved):
Fantastic thanks

PR #2709, opened 2021-12-01
ENH Switch Parent to getParent method to make SiteTree faster

sunnysideup:
This change will make Silverstripe much faster (i hope), because it uses SiteTree::getParent() instead of the magic / custom method Parent()

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -684,7 +684,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     public function RelativeLink($action = null)
     {
         if ($this->ParentID && self::config()->get('nested_urls')) {
-            $parent = $this->Parent();
+            $parent = $this->getParent();
             // If page is removed select parent from version history (for archive page view)
             if ((!$parent || !$parent->exists()) && !$this->isOnDraft()) {
                 $parent = Versioned::get_latest_version(self::class, $this->ParentID);
@@ -809,7 +809,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         }
 
         // Parent must exist and not be an orphan itself
-        $parent = $this->Parent();
+        $parent = $this->getParent();
         return !$parent || !$parent->exists() || $parent->isOrphaned();
     }
 
@@ -863,7 +863,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             if ($sectionName === $page->URLSegment) {
                 return true;
             }
-            $page = $page->Parent();
+            $page = $page->getParent();
         }
         return false;
     }
@@ -966,7 +966,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
                 $pages[] = $page;
             }
 
-            $page = $page->Parent();
+            $page = $page->getParent();
         }
 
         return new ArrayList(array_reverse($pages ?? []));
@@ -995,13 +995,13 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     /**
      * Get the parent of this page.
      *
-     * @return SiteTree Parent of this page
+     * @return SiteTree|null
      */
     public function getParent()
     {
         $parentID = $this->getField("ParentID");
         if ($parentID) {
-            return SiteTree::get_by_id(self::class, $parentID);
+            return SiteTree::get_by_id($parentID);
         }
         return null;
     }
@@ -1176,7 +1176,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         // check for inherit
         if ($this->CanViewType === InheritedPermissions::INHERIT) {
             if ($this->ParentID) {
-                return $this->Parent()->canView($member);
+                return $this->getParent()->canView($member);
             } else {
                 return $this->getSiteConfig()->canViewPages($member);
             }
@@ -1872,7 +1872,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         // Check known urlsegment blacklists
         if (self::config()->get('nested_urls') && $this->ParentID) {
             // Guard against url segments for sub-pages
-            $parent = $this->Parent();
+            $parent = $this->getParent();
             if ($controller = ModelAsController::controller_for($parent)) {
                 if ($controller instanceof Controller && $controller->hasAction($this->URLSegment)) {
                     return false;
@@ -2132,7 +2132,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
 
         $baseLink = Controller::join_links(
             Director::absoluteBaseURL(),
-            (self::config()->get('nested_urls') && $this->ParentID ? $this->Parent()->RelativeLink(true) : null)
+            (self::config()->get('nested_urls') && $this->ParentID ? $this->getParent()->RelativeLink(true) : null)
         );
 
         $urlsegment = SiteTreeURLSegmentField::create("URLSegment", $this->fieldLabel('URLSegment'))
@@ -2755,7 +2755,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
                 if ($instance instanceof HiddenClass) {
                     continue;
                 }
-                if (!$instance->canCreate(null, ['Parent' => $this->ParentID ? $this->Parent() : null])) {
+                if (!$instance->canCreate(null, ['Parent' => $this->ParentID ? $this->getParent() : null])) {
                     continue;
                 }
             }
@@ -3033,7 +3033,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     {
         $parent = $this;
         $stack = [$parent];
-        while (($parent = $parent->Parent()) && $parent->exists()) {
+        while (($parent = $parent->getParent()) && $parent->exists()) {
             array_unshift($stack, $parent);
         }
 
@@ -3048,7 +3048,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     public function getPageLevel()
     {
         if ($this->ParentID) {
-            return 1 + $this->Parent()->getPageLevel();
+            return 1 + $this->getParent()->getPageLevel();
         }
         return 1;
     }
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -683,7 +683,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      */
     public function RelativeLink($action = null)
     {
-        if ($this->ParentID && self::config()->get('nested_urls')) {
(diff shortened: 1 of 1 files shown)


sunnysideup:
@GuySartorelli - in my application it seems to make a lot of enhancements. However, even if it was the same, I would recommend implementing the change because `Parent` is a generically created method based on a `has_one`, while `getParent` is a specific SiteTree method that can be optimised for SiteTree.

sunnysideup:
@GuySartorelli - I would love you to try it on a larger site.  I just tried it on a random site we run.  I opened the home page with 
`?flush=all` so that the menu, etc... is not cached. Without this fix, the time was 1.60s to load, with this fix it was around 1.45s (running both around 10 times to get an average value). I'd be interested to see what you get.

wilr:
I can replicate the improvement on a stock standard install.

```
^ Parent()->Link took:
  0.001104

^ getParent()->Link took
  0.000040

^ Parent()->Link after cached (2nd call) took:
  0.000057

^ getParent()->Link after cached (2nd call) took
  0.000023
```

`SiteTree` defines a `getParent()`, so does `Hierarchy` (we should probably remove it from `SiteTree`). 

The performance improvement probably shouldn't be shocking - Both those methods bypass the magic layer of Silverstripe (`CustomMethods::__call()`) For comparison, `$page->getComponent('Parent')` is what happens under the hood on the `__call()` dataobject bypassing most of the config / extension work and that is slower

code/Model/SiteTree.php
```diff
@@ -1133,7 +1133,7 @@ public function canView($member = null)
         // check for inherit
         if ($this->CanViewType === InheritedPermissions::INHERIT) {
             if ($this->ParentID) {
-                return $this->Parent()->canView($member);
+                return $this->getParent()->canView($member);
```
GuySartorelli, line 1136: If there is no parent, this will attempt to call `canView` on `null`.

code/Model/SiteTree.php
```diff
@@ -2001,7 +2001,7 @@ public function getCMSFields()
 
         $baseLink = Controller::join_links(
             Director::absoluteBaseURL(),
-            (self::config()->get('nested_urls') && $this->ParentID ? $this->Parent()->RelativeLink(true) : null)
+            (self::config()->get('nested_urls') && $this->ParentID ? $this->getParent()->RelativeLink(true) : null)
```
GuySartorelli, line 2004: If there is no parent, this will attempt to call `RelativeLink` on `null`

code/Model/SiteTree.php
```diff
@@ -1740,7 +1740,7 @@ public function validURLSegment()
         // Check known urlsegment blacklists
         if (self::config()->get('nested_urls') && $this->ParentID) {
             // Guard against url segments for sub-pages
-            $parent = $this->Parent();
+            $parent = $this->getParent();
             if ($controller = ModelAsController::controller_for($parent)) {
```
GuySartorelli, line 1744: If there's no parent, we're passing in null - but the typehinting for `ModelAsController::controller_for` requires a `SiteTree` object.

There are probably other situations like this that need to be accounted for which is probably why tests are failing.

sunnysideup:
@GuySartorelli - what might be useful is if the automated tests also included some sort of timing feedback.  I guess that is super rough, but it may give some indication on how fast things run with the proposed changes.

code/Model/SiteTree.php
sunnysideup: Yes, I completely rewrote my work now as there were tons of errors.

Here is what I did in your instance:
https://github.com/silverstripe/silverstripe-cms/pull/2709/commits/c849252512ab83c60f068d197c02896a684365a7#diff-08b08699fad031e3fb13c3637d1a2b0a71a0b82f2bd69606b2244b12f89b6129L1741-L1747

sunnysideup:
@wilr - thank you for looking at this.  I think the best gains are made in PHP code  were we write:

```php

$parent = $this->Parent();
if($parent && $parent->exists()) {
    //....
}
```

we should encourage people to write:
```
$parent = $this->getParent();
if($parent) {
    //....
}
```

This is not really part of this proposed change, but it does implement that idea withi SiteTree itself to make the code simpler and more consistent.

sunnysideup:
Just running this again.  I will make sure, of course, all tests pass. Just relying on this to pick up any issues.

code/Model/SiteTree.php
```diff
@@ -641,10 +641,10 @@ public function getMimeType()
      */
     public function RelativeLink($action = null)
     {
-        if ($this->ParentID && self::config()->get('nested_urls')) {
-            $parent = $this->Parent();
+        $parent = $this->getParent();
+        if ($parent && self::config()->get('nested_urls')) {
             // If page is removed select parent from version history (for archive page view)
-            if ((!$parent || !$parent->exists()) && !$this->isOnDraft()) {
+            if (! $this->isOnDraft()) {
```
emteknetnz, line 647:
```suggestion
            if (!$this->isOnDraft()) {
```

code/Model/SiteTree.php
```diff
@@ -951,16 +946,26 @@ public function setParent($item)
...
      * Get the parent of this page.
+     * This is here to keep compatability with the has_one magic method
+     * that always returns an object, even if there is none for the specific relation.
      *
-     * @return SiteTree Parent of this page
+     * @return SiteTree Parent of this page or SiteTree singleton
+     */
+    public function Parent()
+    {
+        return $this->getParent() ?: Injector::inst()->get(self::class);
```
emteknetnz, line 956: So calling `$RootPage->Parent` would previously return a `SiteTree` singleton?  This seems a bit odd.  Was the previously the reason for all those `$parent->exists()` checks?

code/Model/SiteTree.php
```diff
@@ -951,16 +946,26 @@ public function setParent($item)
...
      */
     public function getParent()
     {
-        $parentID = $this->getField("ParentID");
-        if ($parentID) {
-            return SiteTree::get_by_id(self::class, $parentID);
+        if ($this->ParentID) {
+            return SiteTree::get_by_id(self::class, $this->ParentID);
         }
-        return null;
```
emteknetnz, line 963: Should keep the explicit return null for readability

code/Model/SiteTree.php
```diff
@@ -951,16 +946,26 @@ public function setParent($item)
...
+     *
+     * @return SiteTree|null Parent of this page
      */
     public function getParent()
     {
-        $parentID = $this->getField("ParentID");
-        if ($parentID) {
-            return SiteTree::get_by_id(self::class, $parentID);
+        if ($this->ParentID) {
+            return SiteTree::get_by_id(self::class, $this->ParentID);
```
emteknetnz, line 967:
```suggestion
            return SiteTree::get()->byID($this->ParentID);
```
Slightly nicer syntax

code/Model/SiteTree.php
```diff
@@ -1555,7 +1561,7 @@ protected function onBeforeWrite()
 
         // If Sort hasn't been set, make this page come after it's siblings
         if (!$this->Sort) {
-            $parentID = ($this->ParentID) ? $this->ParentID : 0;
+            $parentID = (int) ($this->ParentID ?: 0);
```
emteknetnz, line 1564: Are you worried about a `null` database value for ParentID?

Don't need to both cast to int and use `?: 0`  Personally I'd remove the cast as `$this->ParentID ?: 0` is used elsewhere

code/Model/SiteTree.php
```diff
@@ -2507,7 +2520,8 @@ public function onAfterRevertToLive()
      */
     protected function isParentArchived()
     {
-        if ($parentID = $this->ParentID) {
+        $parentID = $this->ParentID;
```
emteknetnz, line 2523: No need to use the $parentID var, just use $this->ParentID directly

code/Model/SiteTree.php
```diff
@@ -3157,7 +3173,7 @@ public function provideI18nEntities()
      */
     public function getParentType()
     {
-        return $this->ParentID == 0 ? 'root' : 'subpage';
+        return $this->ParentID === 0 ? 'root' : 'subpage';
```
emteknetnz, line 3176:
```suggestion
        return !$this->ParentID ? 'root' : 'subpage';
```
`===` will break if $this->ParentID is `null` for whatever reason

code/Model/SiteTree.php
```diff
@@ -760,14 +760,9 @@ public function isSection()
      */
     public function isOrphaned()
     {
-        // Always false for root pages
-        if (empty($this->ParentID)) {
```
emteknetnz, line 764: Retain this check

`return !$parent` below means we've flipped the logic here where a root page (with no parent) will mistakenly return `true` for `isOrphaned()` when it should return `false`

emteknetnz (changes requested):
I like this PR, there's some changes to make

code/Model/SiteTree.php
sunnysideup: There is a difference between byID and get_by_id - from the code:

```php
// The object returned is cached, unlike DataObject::get()->byID() {@link DataList::byID()}
```

code/Model/SiteTree.php
sunnysideup: yes, because a has_one returns a singleton if there is no relationship.

code/Model/SiteTree.php
sunnysideup: I guess I was worried about just whatever it would be.

sunnysideup:
I have made all the changes as requested.

michalkleiner (approved):
Looks ok from my perspective, nice improvement.

GuySartorelli:
Looks good in general. @sunnysideup can you please take a look at the tests? There's a lot of cases now where null is returned when the tests were previously getting pages.

dhensby:
I'm not a huge fan of this kind of change. In reality the magic methods should be more improved so they perform in a more optimal way (ie: make use of in-memory caches like this implementation does) rather than fixing this on a case-by-case basis.

This is one of the biggest regressions going from v3 to v4 of SS; the lack of caching of relationships/ORM queries.

GuySartorelli:
Retargetted to `5` and rebased to see what the CI tests think of this PR in its current state

GuySartorelli:
I've added some protection against null values to resolve _most_ of the test failures.

While I do agree with @dhensby that the ORM could afford to be more efficient, with perhaps some built in caching in places it doesn't currently have it, I don't think that's something we can sensibly do in a minor release.
This change is though - and since the `getParent()` method already exists it seems sensible to start using it.

## The failed test

The remaining failed test is failing because we've cached the query for the parent when the parent for that ID _doesn't actually exist_ (because it hasn't yet been restored) - and then after restoring it, when we query for the parent, we're told it doesn't exist (even though now it does).

I think this PR is opening ourselves to a bunch of small regressions like that one, and while it does represent a performance gain, I think there's too much chance of regression for my tastes here.

I'm going to close this PR for now - but if someone thinks of a way we can mitigate this class of regression, I'd be happy to discuss and potentially reopen the PR.

PR #2717, opened 2022-01-25
FIX SiteTree visibility field name

xini:
If not set explicitly, the field name is taken from the title and changes if the site uses a different locale.


kinglozzer (approved):
Thanks @xini, we definitely should have a consistent name here

Issue #2708, opened 2021-12-01
Include version number in generator meta tag


maxime-rainville:
# Description

We already provided a generator meta tag that tells the world that a site is running Silverstripe CMS. Adding a version number would allow us to better track adoption and upgrades in the real world

# Acceptance criteria
- [ ] [`SiteTree::MetaComponent`](https://github.com/silverstripe/silverstripe-cms/blob/c75448669300ce5ed94a6e20c0a386ff4219167d/code/Model/SiteTree.php#L1395) is updated to include the version number
- [ ] The version number is determined using a mechanism similar to [LeftAndMain's `VersionProvider`](https://github.com/silverstripe/silverstripe-admin/blob/d3fc48cde0aef36518d0b7efd11dd5711bef6be4/code/LeftAndMain.php#L161)
- [ ] The version number is provided by default
- [ ] Format is switch to "Silverstripe CMS 4.xx" without the patch release number
- [ ] It's possible to easily disable the generator meta tag
- [ ] There's clear guidance on how to disable this feature for project who consider this information sensitive.
- [ ] Validate performance and implement caching if necessary

# Related issue
- https://github.com/silverstripe/silverstripe-admin/issues/1256

# PRs
- [x] https://github.com/silverstripe/silverstripe-framework/pull/10235
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2721

lerni:
I understand the desire to better track version adoption but I consider adding version as a security risk. I know you can remove the tag but, but as default it'll remain for the majority of projects and makes vulnerability analysis much to easy.

michalkleiner:
I agree with lerni. In theory, exposing any version information on a fully patched and properly set up system is no risk, again, in theory, but the real world is different.

If we added this as an optional feature behind a config flag, off by default, I'd be happy to have it, and leave it up the site owners/site vendors to decide whether they want to turn it on.

maxime-rainville:
There's already plenty of ways to identify that a site is running Silverstripe CMS. Not disclosing that a site is running version X is just an extension of the "security by obscurity" fallacy. It provides no meaningful security layer and prevents us from collecting useful metadata about Silverstripe CMS usage in the wild.

emteknetnz:
Security by obscurity provides virtually zero additional security and it cannot be relied on whatsoever

[...]

PR #2721, opened 2022-02-16
NEW Include version number in meta generator tag

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/2708

Will require [this PR](https://github.com/silverstripe/silverstripe-framework/pull/10235) to be merged first

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -21,6 +21,7 @@ use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Manifest\ModuleResource;
 use SilverStripe\Core\Manifest\ModuleResourceLoader;
+use SilverStripe\Core\Manifest\VersionProvider;
 use SilverStripe\Core\Resettable;
 use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\CheckboxField;
@@ -339,7 +340,16 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      * @config
      * @var string
      */
-    private static $meta_generator = 'SilverStripe - https://www.silverstripe.org';
+    private static $meta_generator = 'Silverstripe CMS';
+
+    /**
+     * Whether to display the version portion of the meta generator tag
+     * Set to false if it's viewed as a concern.
+     *
+     * @config
+     * @var bool
+     */
+    private static $show_meta_generator_version = true;
 
     protected $_cache_statusFlags = null;
 
@@ -394,6 +404,11 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      * @var CacheInterface
      */
     protected $creatableChildrenCache;
+
+    /**
+     * @var VersionProvider
+     */
+    private $versionProvider;
 
     /**
      * Fetches the {@link SiteTree} object that maps to a link.
@@ -1401,13 +1416,13 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             'content' => $this->obj('Title')->forTemplate()
         ];
 
-        $generator = trim(Config::inst()->get(self::class, 'meta_generator'));
-        if (!empty($generator)) {
+        $generator = $this->getGenerator();
+        if ($generator) {
             $tags['generator'] = [
                 'attributes' => [
                     'name' => 'generator',
-                    'content' => $generator,
-                ],
+                    'content' => $generator
+                ]
             ];
         }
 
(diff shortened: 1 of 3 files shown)


code/Model/SiteTree.php
```diff
@@ -339,7 +340,16 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      * @config
      * @var string
      */
-    private static $meta_generator = 'SilverStripe - https://www.silverstripe.org';
+    private static $meta_generator = 'Silverstripe CMS';
+
+    /**
+     * Whether to display the version portion of the meta generator tag
+     * Set to false if it's viewed as a security convern.
```
michalkleiner, line 347:
```suggestion
     * Set to false if it's viewed as a security concern.
```

code/Model/SiteTree.php
```diff
@@ -1449,6 +1464,47 @@ public function MetaComponents()
...
+     * @return string
+     */
+    private function getGenerator(): string
+    {
+        $generator = trim(Config::inst()->get(self::class, 'meta_generator'));
+        if ($generator === '') {
+            return '';
+        }
+        if (self::config()->get('show_meta_generator_version')) {
+            $version = $this->getVersionProvider()->getModuleVersion('silverstripe/framework');
```
michalkleiner, line 1480: There was a whole exercise of fallbacks within CWP to show something more relevant than just the framework version or the single recipe version, is framework the right source of this information? It's the most reliable one, though it's not uncommon to use a newer framework version than CMS or the recipe.

michalkleiner (approved):

code/Model/SiteTree.php
```diff
@@ -339,7 +340,16 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
...
+    private static $meta_generator = 'Silverstripe CMS';
+
+    /**
+     * Whether to display the version portion of the meta generator tag
+     * Set to false if it's viewed as a concern.
+     *
+     * @config
+     * @var bool
+     */
+    private static $show_meta_generator_version = true;
```
maxime-rainville, line 352: Why not just keep tho old behavior and not output the meta tag if `meta_generator' is blank?

code/Model/SiteTree.php
maxime-rainville: It's an interesting question.

The version number displayed in the CMS is meant for actual CMS users. This one is really meant for robots, so I think it makes less sense to have the convoluted fallback logic.

Since this meta tag is coming from `silverstripe/cms`, maybe it would make more sense to track that module ... but I don't have string feelings either way.

maxime-rainville (approved):

maxime-rainville (changes requested):
Sorry ... meant to reject.

code/Model/SiteTree.php
emteknetnz: That behavior is still there, if you config `private static $meta_generator` to a blank string then there won't be a meta tag for generator

i.e.
```php

    public function MetaComponents()
...
        $generator = $this->getGenerator();
        if ($generator) {
            $tags['generator'] = [
...
    }

    private function getGenerator(): string
    {
        $generator = trim(Config::inst()->get(self::class, 'meta_generator'));
        if ($generator === '') {
            return '';
        }
        if (self::config()->get('show_meta_generator_version')) {
        ...
    }
```

code/Model/SiteTree.php
emteknetnz: At some point in the future we could theoretically expose this number on a non traditional site without pages that doesn't have CMS installed, possibly via an API endpoint (I'm speculating here).  In that scenario we'd get the version from framework, so it makes sense to use framework here so the number would be aligned.

code/Model/SiteTree.php
maxime-rainville: I just don't understand what kind of user would say "I don't want the version number there, but I'm fine advertising the site uses Silverstripe CMS".

My gut feeling is the vast majority of users who are bothered by the version number will want to remove the entire meta tag.

emteknetnz: Probably very few users.  However `private static $meta_generator = 'Silverstripe CMS';` is currently a configurable string, so we do support changing this to something else e.g. 'CMS from Mars'.  

If people want to report this as a different CMS for whatever reason, then I can see why they wouldn't want to include the version number

I'm not hugely attached to having these two things separately configurable, though I also don't see it as much extra hassle

maxime-rainville (approved):
Tested it locally and it works fine. I hacked my `composer.lock` file to test that the version is being picked up properly.

There's the slight question about what to do with "4.x-dev". Since this doesn't match the expected regex format, it just comes back with nothing. The help menu in the CMS renders "4.x". 

The behaviour in this PR is pretty defensible and I don't have any beef with it. Just think it's worth pointing out this is happening.

maxime-rainville (changes requested):
Actually, the composer file needs to be updated to require framework `^4.11` since we rely on the new API being present.

emteknetnz:
> it just comes back with nothing

That was intentional, some people will be running forks of framework, don't want those being reported.  I'll add an inline note about this

maxime-rainville (approved):

maxime-rainville:
Framework PR has been merged which should make this PR green. Just restarted the build

PR #2726, opened 2022-03-27
FIX Ensure search form start var is not a negative number

3Dgoo:
If a user changes the search form URL start parameter to a negative number (`&start=-10`), this will cause a server error.

This change fixes this problem by making sure start is always an int and is never less than zero.

### DIFF
--- a/code/Search/SearchForm.php
+++ b/code/Search/SearchForm.php
@@ -163,7 +163,7 @@ class SearchForm extends Form
         $keywords = $this->addStarsToKeywords($keywords);
 
         $pageLength = $this->getPageLength();
-        $start = $request->requestVar('start') ?: 0;
+        $start = max(0, (int)$request->requestVar('start'));
 
         $booleanSearch =
             strpos($keywords, '"') !== false ||


code/Search/SearchForm.php
```diff
@@ -163,7 +163,7 @@ public function getResults()
         $keywords = $this->addStarsToKeywords($keywords);
 
         $pageLength = $this->getPageLength();
-        $start = $request->requestVar('start') ?: 0;
+        $start = max(0, (int)$request->requestVar('start')) ?: 0;
```
dhensby, line 166: The ternary operator isn't needed any more

```diff
-$start = max(0, (int)$request->requestVar('start')) ?: 0;
+$start = max(0, (int)$request->requestVar('start'));
```

dhensby (changes requested):

dhensby:
This looks like a bug fix so can go against 4.10 branch

code/Search/SearchForm.php
3Dgoo: Sure thing. I have made this change

3Dgoo:
Sure thing. This is done and is ready to go.

GuySartorelli:
Looks like you've inadvertently added a lot of changes from the 4 branch when you rebased. Can you please try again, so that only your change is added?

dhensby:
You need to reset your branch to 4.10 and chrrry-pick your commit into it

dhensby (approved):

3Dgoo:
Thanks Dan

PR #2727, opened 2022-03-27
Fix search form ending in dash error

3Dgoo:
On the search form if a user's search ends in a dash it will cause a server error.

This change fixes this problem by trimming dashes and spaces from the end of the search term.

### DIFF
--- a/code/Search/SearchForm.php
+++ b/code/Search/SearchForm.php
@@ -146,7 +146,7 @@ class SearchForm extends Form
             }
         }
 
-        $keywords = $request->requestVar('Search');
+        $keywords = rtrim($request->requestVar('Search'), '- ');
 
         $andProcessor = function ($matches) {
             return ' +' . $matches[2] . ' +' . $matches[4] . ' ';


michalkleiner (changes requested):
To trim whitespaces is probably legitimate here, but I don't think we should be trimming potentially meaningful characters. It seems the problem here is that the hyphen in the keyword isn't escaped properly for the db query where it uses a minus to create a not predicate. So I'd prefer a fix taking that into account.

muskie9:
I've been looking at another similar scenario where FullTextSearch against something like ABC-123, where ABC-123 is in the content zone it yields no results. This is due to the not predicate from what I've seen thus far.

I'm wondering if making the modifiers more flexible as right now I believe it simply looks for + or - and assumes you know what those do within the keyword string and how it affects the query. As a developer I didn't know that was the case so I'd guess CMS and end users would have no idea.

3Dgoo:
The code that follows the initial keyword fetch shows that the code uses the ` -` to create a "not" search condition:
```
        $andProcessor = function ($matches) {
            return ' +' . $matches[2] . ' +' . $matches[4] . ' ';
        };
        $notProcessor = function ($matches) {
            return ' -' . $matches[3];
        };
        $keywords = preg_replace_callback('/()("[^()"]+")( and )("[^"()]+")()/i', $andProcessor, $keywords);
        $keywords = preg_replace_callback('/(^| )([^() ]+)( and )([^ ()]+)( |$)/i', $andProcessor, $keywords);
        $keywords = preg_replace_callback('/(^| )(not )("[^"()]+")/i', $notProcessor, $keywords);
        $keywords = preg_replace_callback('/(^| )(not )([^() ]+)( |$)/i', $notProcessor, $keywords);
        $keywords = $this->addStarsToKeywords($keywords);
```

I think escaping the `-` would break that code.

On the other hand, stripping any dashes on the end of the whole search string shouldn't break anything.

PR #2728, opened 2022-03-29
PATCH: make cms faster by not calling mapFn twice

sunnysideup:
This function should really be moved to Group.php!

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -2183,8 +2183,9 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             asort($map);
             return $map;
         };
+        
         $viewAllGroupsMap = $mapFn(Permission::get_groups_by_permission(['SITETREE_VIEW_ALL', 'ADMIN']));
-        $editAllGroupsMap = $mapFn(Permission::get_groups_by_permission(['SITETREE_EDIT_ALL', 'ADMIN']));
+        $editAllGroupsMap = $viewAllGroupsMap;
 
         $fields = new FieldList(
             $rootTab = new TabSet(


dhensby (changes requested):
But they do different things 😬 

`SITETREE_VIEW_ALL` vs `SITETREE_EDIT_ALL`

dhensby:
😅  which part do you think is useful? I'm not sure it has much generic purpose...

sunnysideup:
mapfn returns a list of groups ready for addition to a description of a field by providing an alternative breadcrumb title.   

https://github.com/silverstripe/silverstripe-cms/pull/2728/commits/06c4776530ca6d73a980f9eddc156796e9aa68c7#diff-08b08699fad031e3fb13c3637d1a2b0a71a0b82f2bd69606b2244b12f89b6129L2283-L2289

There are many ways to make this more "organised":
a. escape descriptions
b. a method providing the statement used in the description
c. a method that provides an escaped list
etc...

I guess all in all, it seems to me that it is helpful to make these "getCMSFields" / "getSettingsFields"  type methods shorter by moving out functionality that is more universal and makes the methods hard to read. 

Even just moving this stuff to its own method would be helpful I feel: https://github.com/silverstripe/silverstripe-cms/pull/2728/commits/06c4776530ca6d73a980f9eddc156796e9aa68c7#diff-08b08699fad031e3fb13c3637d1a2b0a71a0b82f2bd69606b2244b12f89b6129L2177-L2187

GuySartorelli:
@sunnysideup thanks for contributing.
Thir PR as it currently stands is broken, as has been pointed out already, and it sounds like the changes you're proposing now would belong in a separate PR.
If you would like to create a new PR that refactors the method in a way that makes potentially generally useful functionality more reusable I'd be happy to review it but I'll close this one for now.

PR #2730, opened 2022-04-04
ENH PHP 8.1 compatibility

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10250


GuySartorelli (approved):
For posterity: See discussion in silverstripe/silverstripe-admin#1294 about the general approach. tl;dr:
> the purposes of this PR is to avoid passing null (specifically null) from built-in functions that as of php 8.1 will throw deprecation warnings if null is passed in. The approach is necessarily heavy-handed, and while there are many situations where it isn't needed, it would be prohibitively laborious to find all of those situations by hand.

Looks good to me, but I'll hold off merging until the outstanding questions about `file_get_contents()` and `file_put_contents()` have been answered on the other PRs.

PR #2737, opened 2022-05-04
FIX Allow empty PreviewURLs for CMSPreviewable objects

GuySartorelli:
An empty PreviewURL will result in the "no preview available" message displaying instead of a 404 error.

Parent Issue
- silverstripe/silverstripe-admin#1304


GuySartorelli:
Changed base from `4` to `4.11` as this is a fix that should be included in the 4.11.0 release.

code/Controllers/SilverStripeNavigatorItem_ArchiveLink.php
```diff
@@ -46,10 +46,8 @@ public function getMessage()
 
     public function getLink()
     {
-        return Controller::join_links(
-            $this->record->PreviewLink(),
-            '?archiveDate=' . urlencode($this->record->LastEdited ?? '')
-        );
+        $link = $this->record->PreviewLink();
+        return $link ? Controller::join_links($link, '?archiveDate=' . urlencode($this->record->LastEdited ?? '')) : null;
```
sabina-talipova, line 50: Probably, we need cover it with unit tests.
And also `join_links()` method returns empty string. I'm not sure, but if somebody uses this method and doesn't process NULL, then he probably would get an Exception.

code/Controllers/SilverStripeNavigatorItem_ArchiveLink.php
```diff
@@ -62,6 +60,7 @@ public function canView($member = null)
             && $this->isArchived()
             // Don't follow redirects in preview, they break the CMS editing form
             && !($record instanceof RedirectorPage)
+            && $this->getLink()
```
sabina-talipova, line 63: Would be nice to test that canView() still works.

sabina-talipova:
Nice work! Just a few small comments. It would be nice to add test coverage for these two methods, but probably not necessary if there are not any other tests for these classes. And, possibly, also add return type for getLink() methods where it's required.

code/Controllers/SilverStripeNavigatorItem_LiveLink.php
```diff
@@ -48,7 +48,8 @@ public function getMessage()
 
     public function getLink()
     {
-        return Controller::join_links($this->getLivePage()->PreviewLink(), '?stage=Live');
+        $link = $this->getLivePage()->PreviewLink();
+        return $link ? Controller::join_links($link, '?stage=Live') : null;
```
sabina-talipova, line 52: Probably, we need cover it with unit tests

code/Controllers/SilverStripeNavigatorItem_LiveLink.php
```diff
@@ -60,6 +61,7 @@ public function canView($member = null)
             && $this->showLiveLink()
             && $record->hasStages()
             && $this->getLivePage()
+            && $this->getLink()
```
sabina-talipova, line 64: Probably, we need cover canView with unit tests.

code/Controllers/SilverStripeNavigatorItem_StageLink.php
```diff
@@ -66,6 +70,7 @@ public function canView($member = null)
             && $this->showStageLink()
             && $record->hasStages()
             && $this->getDraftPage()
+            && $this->getLink()
```
sabina-talipova, line 73: Probably, we need cover canView with unit tests.

code/Controllers/SilverStripeNavigatorItem_StageLink.php
```diff
@@ -50,8 +50,12 @@ public function getMessage()
     public function getLink()
     {
         $date = Versioned::current_archived_date();
+        $link = $this->record->PreviewLink();
+        if (!$link) {
+            return null;
+        }
         return Controller::join_links(
```
sabina-talipova, line 57: Probably, we need cover it with unit tests.

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
```diff
@@ -42,6 +42,7 @@ public function canView($member = null)
         return (
             !$this->getRecord()->hasExtension(Versioned::class)
             && $this->showUnversionedLink()
+            && $this->getLink()
```
sabina-talipova, line 45: Probably, we need cover canView with unit tests.

GuySartorelli:
I've completely redone the way these were being tested. The old tests were only partially testing the logic at best - and testing effectively dead code at worst.

Redoing these tests also pointed out another bug which I've fixed in the same PR. Note there are three commits - each has its own prefix as appropriate for the changelog, so squashing here won't be appropriate.

code/Controllers/SilverStripeNavigatorItem_ArchiveLink.php
GuySartorelli: Good catch, I've used an empty string now instead. Unit tests also included,

GuySartorelli:
Errors are due to recent changes in travis shared config - will force a push in hopes that fixes it.

sabina-talipova (approved):

PR #2742, opened 2022-06-09
FIX Always include unsaved anchors from the current WYSIWYG field

GuySartorelli:
Follow-up to https://github.com/silverstripe/silverstripe-cms/pull/2741

In that PR, I accidentally caused a regression where an anchor which has _not yet been saved_ would not be available to link from the same field.

This PR resolves that regression, while still fetching any additional anchors from the server for the current page (i.e. from other fields or DataObjects which _have_ been saved).

## Parent issue
- https://github.com/silverstripe/silverstripe-elemental/issues/186


client/src/legacy/TinyMCE_sslink-anchor.js
```diff
@@ -30,6 +30,14 @@ const plugin = {
   init(editor) {
     editor.addCommand(commandName, () => {
       const field = jQuery(`#${editor.id}`).entwine('ss');
+      // Get the anchors in the current field and save them as props for AnchorSelectorField
+      const currentPageID = Number(jQuery('#Form_EditForm_ID').val() || 0);
+      const validTargets = editor
+        .$('[id],[name]', editor.getBody())
+        .toArray()
+        .map((element) => element.id || element.name);
```
sabina-talipova, line 38: It still doesn't work as expected.
` const validTargets ` contains the anchors only from particular this WISIWYG editor on which we click on "Anchor on a page".

So it works perfect when we click on this action in Page's HTML Editor (not ElementalArea Content Block), but only if the page has one HTMLText field.
If Page has more than one HTMLText field, each field sees only its own anchors.

ElementalArea Content Block doesn't see any anchors from Page's HTML Editor, only from ElementalArea.

GuySartorelli: > `const validTargets` contains the anchors only from particular this WISIWYG editor on which we click on "Anchor on a page".

That _is_ working as expected. This PR is just to fix a regression I introduced with #2741, which behaved in that same way. There is no easy way to get the unsaved anchors for other WYSIWYG fields in the same form - much less to know for certain that those fields _will even be used on the front-end_ - plus there may be fields in other tabs, etc.... it's a much more complicated situation than this PR is intended to deal with.

This change effectively means the following should be true:

- Unsaved anchors _in the current field_ are included in the anchors list for the current page.
- SAVED anchors in _any other fields or DataObjects_ for this page (as discovered by `SiteTree::getAnchorsOnPage()`) are included in the anchors list for the current page.
- No change to how anchors on other pages are handled
- No change to how _UNSAVED_ anchors in other fields _for this page_ are handled.

sabina-talipova:
Just small request, could you update this test if it's needed https://github.com/silverstripe/silverstripe-cms/blob/77b5bfa497d5026ffc806f1071a6c5d04d65bfb4/client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js#L66

GuySartorelli:
I don't think it's feasible to test this via jest - instead I have added two new behat tests.
Because there are now several tests for anchor links, I've split that off into its own feature file.

sabina-talipova (approved):
Works as expected in local environment. Tests passed. APPROVED

PR #2753, opened 2022-07-25
ENH Keep state in Page's Breadcrumbs

sabina-talipova:
### Description
- Additional get parameters in Breadcrumbs links

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1314

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -1084,14 +1084,19 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         $ancestors = $record->getAncestors();
         $ancestors = new ArrayList(array_reverse($ancestors->toArray() ?? []));
         $ancestors->push($record);
+
+        $params = $this->getRequest()->getVars();
         /** @var SiteTree $ancestor */
         foreach ($ancestors as $ancestor) {
             $items->push(new ArrayData([
                 'Title' => $ancestor->getMenuTitle(),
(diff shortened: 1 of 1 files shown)


code/Controllers/CMSMain.php
```diff
@@ -1084,14 +1084,19 @@ public function Breadcrumbs($unlinked = false)
...
                 'Title' => $ancestor->getMenuTitle(),
                 'Link' => ($unlinked)
                     ? false
-                    : $ancestor->CMSEditLink()
-            ]));
+                    : Controller::join_links(
+                        $ancestor->CMSEditLink(),
+                        '?' . http_build_query($params ?? [])
+                    )
+                ]));
```
GuySartorelli, line 1099: I don't think we want this - we probably only want this on the "back" button in the breadcrumb rather than each link.

If we _do_ want this to apply to each link, we should do this from the `GridFieldDetailForm_ItemRequest::Breadcrumbs()` method. This already has a list of all of the breadcrumbs (`$items = $this->popupController->Breadcrumbs($unlinked);`) which we can loop through like this:

```php
foreach ($items as $item) {
    if ($item->Link) {
        $item->Link = $this->gridField->addOtherGridfieldStateToURL($item->Link);
    }
}
```
That way we only have this in the context of a gridfield, instead of arbitrarily adding all params to every breadcrumb.

GuySartorelli (changes requested):
Don't think we need this PR

PR #2757, opened 2022-07-29
MNT Use update-js action

emteknetnz:
Issue https://github.com/silverstripe/gha-update-js/issues/5


GuySartorelli (approved):
The travis failure was accompanied by github actions CI failures. Github actions is green after re-running it. Won't rerun travis because' we're getting rid of that anyway.

PR #2761, opened 2022-08-06
API Migrate SilverStripeNagivator classes

GuySartorelli:
These classes are useful with `silverstripe/admin` and `silverstripe/versioned` without needing `silverstripe/cms`.

See also https://github.com/silverstripe/silverstripe-versioned-admin/pull/253 and https://github.com/silverstripe/silverstripe-admin/pull/1340

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1339


code/Controllers/SilverStripeNavigatorItem_CMSLink.php
```diff
@@ -16,13 +17,13 @@ public function getHTML()
...
             $this->record->CMSEditLink(),
-            _t('SilverStripe\\CMS\\Controllers\\ContentController.CMS', 'CMS')
+            _t(ContentController::class . '.CMS', 'CMS')
         );
     }
 
     public function getTitle()
     {
-        return _t('SilverStripe\\CMS\\Controllers\\ContentController.CMS', 'CMS', 'Used in navigation. Should be a short label');
+        return _t(ContentController::class . 'CMS', 'CMS', 'Used in navigation. Should be a short label');
```
dhensby:
```suggestion
        return _t(ContentController::class . '.CMS', 'CMS', 'Used in navigation. Should be a short label');
```

code/Controllers/SilverStripeNavigatorItem_CMSLink.php
```diff
@@ -16,13 +17,13 @@ public function getHTML()
         return sprintf(
             '<a href="%s">%s</a>',
             $this->record->CMSEditLink(),
-            _t('SilverStripe\\CMS\\Controllers\\ContentController.CMS', 'CMS')
+            _t(ContentController::class . '.CMS', 'CMS')
```
dhensby: Should this not just use `$this->getTitle()` - if not, it should use the same fallback text as `getTitle()` (third param: `Used in navigation. Should be a short label`)

dhensby (dismissed):

code/Controllers/SilverStripeNavigatorItem_CMSLink.php
ChloeSartorelli: I'm not refactoring the class, I'm just migrating some of them and in this case changing a hardcoded string to `::class` syntax.

code/Controllers/SilverStripeNavigatorItem_CMSLink.php
ChloeSartorelli: Good catch, I'll fix that tomorrow

code/Controllers/SilverStripeNavigatorItem_CMSLink.php
dhensby: I think if you're going to change LoC, then they should be changed so they are fully correct, not partly changed.

By slipping these unrelated changes into a migration commit/PR you're risking regressions for no benefit. If we at least take the chance to improve the implementation when slipping these unrelated changes in, then that somewhat justifies the risks of regressions slip through the net.

ChloeSartorelli: I'll revert this change tomorrow then.

code/Controllers/SilverStripeNavigatorItem_CMSLink.php
ChloeSartorelli: Reverted instead per https://github.com/silverstripe/silverstripe-cms/pull/2761#discussion_r970544389

code/Controllers/SilverStripeNavigatorItem_CMSLink.php
ChloeSartorelli: Reverted

GuySartorelli:
Needs https://github.com/silverstripe/silverstripe-admin/pull/1340 to be merged before it will go green.

sabina-talipova (approved):
Preview works perfect with or without installed silverstripe-cms module.

PR #2764, opened 2022-08-23
API Use strong types from silverstripe/framework#10457

GuySartorelli:
Merge this before #10457 and then re-run CI on that PR, since that one has the greater number of changes and therefore is more likely to require changes.

## Parent issues
- https://github.com/silverstripe/silverstripe-framework/issues/10350
- https://github.com/silverstripeltd/product-issues/issues/583


GuySartorelli:
The strong-typing commit was removed from the parent PR.

PR #2160, opened 2018-05-04
In unit tests, use SiteTree instead of Page where possible

mikenz:
Developers can customise the way the Page and PageController classes work. Unit tests shouldn't rely on these classes.

PR #2766, opened 2022-08-24
Rescue Master Branch PR: MNT Use SiteTree instead of Page in tests

GuySartorelli:
Rescues https://github.com/silverstripe/silverstripe-cms/pull/2160/

Note that I've swapped out `::create()` syntax for using the `new` keyword, which is preferred for all tests that don't explicitly need to rely on `Injector`.

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10350

### DIFF
--- a/tests/php/Model/SiteTreeTest.php
+++ b/tests/php/Model/SiteTreeTest.php
@@ -3,7 +3,6 @@
 namespace SilverStripe\CMS\Tests\Model;
 
 use LogicException;
-use Page;
 use Psr\SimpleCache\CacheInterface;
 use ReflectionMethod;
 use SilverStripe\CMS\Model\RedirectorPage;
@@ -36,6 +35,9 @@ use SilverStripe\View\Parsers\ShortcodeParser;
 use SilverStripe\View\Parsers\URLSegmentFilter;
 use SilverStripe\View\Shortcodes\EmbedShortcodeProvider;
 use TractorCow\Fluent\Extension\FluentSiteTreeExtension;
+use Page;
+use PageController;
+
 use const RESOURCES_DIR;
 
 class SiteTreeTest extends SapphireTest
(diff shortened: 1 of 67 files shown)


tests/php/Model/SiteTreeTest.php
```diff
@@ -470,46 +479,47 @@ public function testGetByLink()
         );
     }
 
-    public function testGetByLinkAbsolute()
```
emteknetnz, line 473: Why is this being removed?

tests/php/Model/SiteTreeTest.php
```diff
@@ -1634,36 +1672,13 @@ public function testLinkExtension()
     }
 
     /**
-     * Test that the controller name for a SiteTree instance can be gathered by appending "Controller" to the SiteTree
+     * Test that the controller name for a Page instance can be gathered by appending "Controller" to the Page
      * class name in a PSR-2 compliant manner.
      */
     public function testGetControllerName()
     {
-        $class = new Page;
```
emteknetnz, line 1642: Why is all of this being removed?

emteknetnz (changes requested):
Seems fine, not sure why a couple of blocks of test code is being removed though

tests/php/Model/SiteTreeTest.php
ChloeSartorelli: Looks like a mistake when I was resolving conflicts. Will add back in.

tests/php/Model/SiteTreeTest.php
ChloeSartorelli: Looks like a mistake when I was resolving conflicts. Will add back in.

tests/php/Model/SiteTreeTest.php
ChloeSartorelli: Done

tests/php/Model/SiteTreeTest.php
ChloeSartorelli: Done

emteknetnz (approved):

PR #2770, opened 2022-08-26
SS5: move updateRelativeLink hook after concatination

xini:
This is to make the relative actually updatable. 

This picks up https://github.com/silverstripe/silverstripe-cms/pull/2677 again for SS5.

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -696,7 +696,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
 
         $link = Controller::join_links($base, '/', $action);
 
-        $this->extend('updateRelativeLink', $base, $action, $link);
+        $this->extend('updateRelativeLink', $link, $base, $action);
 
         return $link;
     }
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -688,15 +688,17 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             $base = $this->URLSegment;
         }
 
-        $this->extend('updateRelativeLink', $base, $action);
-
         // Legacy support: If $action === true, retain URLSegment for homepages,
         // but don't append any action
         if ($action === true) {
             $action = null;
         }
 
-        return Controller::join_links($base, '/', $action);
+        $link = Controller::join_links($base, '/', $action);
+
+        $this->extend('updateRelativeLink', $link, $base, $action);
+
+        return $link;
     }
 
     /**
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -696,7 +696,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
 
         $link = Controller::join_links($base, '/', $action);
 
-        $this->extend('updateRelativeLink', $link, $base, $action);
+        $this->extend('updateRelativeLink', $base, $action, $link);
 
         return $link;
     }
--- a/code/Model/SiteTreeExtension.php
+++ b/code/Model/SiteTreeExtension.php
@@ -77,13 +77,13 @@ abstract class SiteTreeExtension extends DataExtension
      *
      * @param string &$link The URL of this page relative to siteroot including
      * the action
-     * @param string &$base The URL of this page relative to siteroot, not including
+     * @param string $base The URL of this page relative to siteroot, not including
      * the action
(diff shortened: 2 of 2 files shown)


code/Model/SiteTree.php
```diff
@@ -688,15 +688,17 @@ public function RelativeLink($action = null)
...
         // Legacy support: If $action === true, retain URLSegment for homepages,
         // but don't append any action
         if ($action === true) {
             $action = null;
         }
 
-        return Controller::join_links($base, '/', $action);
+        $link = Controller::join_links($base, '/', $action);
+
+        $this->extend('updateRelativeLink', $link, $base, $action);
```
ChloeSartorelli, line 699: Add link as the last argument. Sure this is targeting a major release but we shouldn't make breaking changes for the hell of it.

GuySartorelli (changes requested):
There are CI failures to address.

GuySartorelli:
What, ultimately, is the intention here?

xini:
The intention is to be able to actually update the relative link. If the link is concatenated after the hook, you can't update the final link.

code/Model/SiteTree.php
xini: Actually, the correct hook should only have the $link parameter, because $base and $action are irrelevant at that point, right?

ChloeSartorelli: They both provide context that might be used in the logic of the extension hook.
They're also both already there so any hook that currently uses them will expect them there.

xini: Hm. True. But any existing extension that is using this would expect for link and action to be concatenated afterwards. So it's a breaking change anyway.

xini: I have moved $link to last

ChloeSartorelli: > any existing extension that is using this would expect for link and action to be concatenated afterwards

That's a good point. Maybe it would be better to have `$link` first then. Though I'm not overly fussed either way. If you want to put it at the beginning again then feel free to do so - otherwise just comment that you will leave it. Once you've reacted either way I'll take a quick look to see if any core logic uses this hook (I doubt it but worth checking) and then merge.

GuySartorelli:
We'll need some words about this in the changelog but the PR that adds the 5.0.0 changelog file hasn't been merged yet. I won't let that hold up this change - instead I'll make a note for myself to add that to the changelog once it exists.

code/Model/SiteTree.php
xini: Thanks. I have switched them back.

GuySartorelli:
`SiteTreeExtension` includes a method signature for `updateRelativeLink(&$base, &$action)` - can you please update both the method signature and the PHPdoc in that extension?

Looks like in core there's no other instance of that method being defined - there might be in some of the supported modules but we haven't started looking at those yet for CMS 5

xini:
Done. Sorry I missed that.

code/Model/SiteTreeExtension.php
```diff
@@ -75,13 +75,15 @@ public function canPublish($member)
...
+     * @param string &$link The URL of this page relative to siteroot including
+     * the action
      * @param string &$base The URL of this page relative to siteroot, not including
      * the action
      * @param string|boolean &$action The action or subpage called on this page.
      * (Legacy support) If this is true, then do not reduce the 'home' urlsegment
      * to an empty link
      */
-    public function updateRelativeLink(&$base, &$action)
+    public function updateRelativeLink(&$link, &$base, &$action)
```
ChloeSartorelli:
```suggestion
    public function updateRelativeLink(&$link, $base, $action)
```
Changes the base or action won't actually do anything (they're here for context only) so those shouldn't be passed by reference anymore. Phpdocs for them need to be updated too.

GuySartorelli (changes requested):

code/Model/SiteTreeExtension.php
xini: Well, why do we pass them at all? As you say, they don't do anything.

ChloeSartorelli: As I say, for context. :p
They show what was used to generate the link in the first place and make it easier to, for example, rebuild the link using the same base but a different action (or vice versa).

GuySartorelli (approved):

GuySartorelli:
Changelog PR: https://github.com/silverstripe/developer-docs/pull/75

PR #2778, opened 2022-09-13
ENH Use the new CMSEditLink extension on SiteTree

GuySartorelli:
Use the new `CMSEditLink` extension from https://github.com/silverstripe/silverstripe-admin/pull/1361 in `SiteTree` - this means a `has_many` relation from some page can set the page as its canonical edit owner and its `CMSEditLink()` method will automatically have a valid URL.

## Note
- CI won't be green until https://github.com/silverstripe/silverstripe-admin/pull/1361 has been merged.

## Related issue
- https://github.com/silverstripe/silverstripe-admin/issues/1354

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -375,6 +375,16 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         return $this->Link('listview');
     }
 
+    /**
+     * Get the link for editing a page.
+     *
+     * @see CMSEditLinkExtension::getCMSEditLinkForManagedDataObject()
+     */
+    public function getCMSEditLinkForManagedDataObject(SiteTree $obj): string
+    {
+        return Controller::join_links(CMSPageEditController::singleton()->Link('show'), $obj->ID);
+    }
+
     public function LinkPageEdit($id = null)
     {
         if (!$id) {
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -4,7 +4,9 @@ namespace SilverStripe\CMS\Model;
 
 use Page;
 use Psr\SimpleCache\CacheInterface;
+use SilverStripe\Admin\CMSEditLinkExtension;
 use SilverStripe\Assets\Shortcodes\FileLinkTracking;
+use SilverStripe\CMS\Controllers\CMSMain;
 use SilverStripe\CMS\Controllers\CMSPageEditController;
 use SilverStripe\CMS\Controllers\ContentController;
 use SilverStripe\CMS\Controllers\ModelAsController;
@@ -113,7 +115,6 @@ use SilverStripe\View\SSViewer;
  */
 class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvider, CMSPreviewable, Resettable, Flushable, MemberCacheFlusher
 {
-
     /**
      * Indicates what kind of children this page type can have.
      * This can be an array of allowed child classes, or the string "none" -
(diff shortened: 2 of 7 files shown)


code/Controllers/CMSMain.php
```diff
@@ -375,6 +375,16 @@ public function LinkListViewDeferred()
         return $this->Link('listview');
     }
 
+    /**
+     * Get the link for editing a page.
+     *
+     * @see CMSEditLinkExtension::getEditLinkForManagedDataObject()
+     */
+    public function getEditLinkForManagedDataObject(SiteTree $obj): string
```
emteknetnz:
```suggestion
    public function getCMSEditLinkForManagedDataObject(SiteTree $obj): string
```

code/Controllers/CMSMain.php
```diff
@@ -375,6 +375,16 @@ public function LinkListViewDeferred()
         return $this->Link('listview');
     }
 
+    /**
+     * Get the link for editing a page.
+     *
+     * @see CMSEditLinkExtension::getEditLinkForManagedDataObject()
```
emteknetnz:
```suggestion
     * @see CMSEditLinkExtension::getCMSEditLinkForManagedDataObject()
```

code/Model/SiteTree.php
```diff
@@ -215,6 +216,12 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      */
     private static $controller_name = null;
 
+    /**
+     * The class of the LeftAndMain controller where this class is managed.
+     * @see CMSEditLinkExtension::getCanonicalEditOwner()
+     */
+    private static string $canonical_edit_owner = CMSMain::class;
```
```suggestion
    private static string $cms_edit_owner = CMSMain::class;
```

code/Model/SiteTree.php
```diff
@@ -734,14 +742,11 @@ public function getAbsoluteLiveLink($includeStageEqualsLive = true)
      */
     public function CMSEditLink()
     {
-        $link = Controller::join_links(
-            CMSPageEditController::singleton()->Link('show'),
-            $this->ID
-        );
-        return Director::absoluteURL($link);
+        // This method has to be implemented here to satisfy the CMSPreviewable interface.
```
emteknetnz, line 745: This should be a method docblock comment rather than an inline comment

code/Controllers/CMSMain.php
```diff
@@ -375,6 +375,16 @@ public function LinkListViewDeferred()
...
+     */
+    public function getEditLinkForManagedDataObject(SiteTree $obj): string
+    {
+        return Controller::join_links(CMSPageEditController::singleton()->Link('show'), $obj->ID);
```
emteknetnz, line 385:
```suggestion
        return $obj->CMSEditLink();
```
?

emteknetnz: Does this method even need to exist on CMSMain?  Is it here simply to make CMSMain more like a ModelAdmin?

emteknetnz (changes requested):

code/Model/SiteTree.php
ChloeSartorelli: I disagree - it's a comment for our internal understanding, not something that needs to be highlighted in the IDE.

code/Controllers/CMSMain.php
ChloeSartorelli: It's here so that the `CMSEditLinkExtension` on `SiteTree` can get its `CMSEditLink`.

ChloeSartorelli: The recommended change will result in an unending recursive cycle. I can't make that change.

emteknetnz:
Have merged admin PR, re-running CI

emteknetnz (approved):

PR #2784, opened 2022-10-02
PATCH: cater for edge cases in CM Edit Link

sunnysideup:
image

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -744,7 +744,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     {
         // This method has to be implemented here to satisfy the CMSPreviewable interface.
         // See the actual implementation in CMSEditLinkExtension.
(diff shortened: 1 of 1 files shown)


sunnysideup:
This is not a good pull request.  I was thinking about it again and it should allow for other extensions to come through if required. Maybe like this:
```php
    /**
     * Generates a link to edit this page in the CMS.
     *
     * @return string
     */
    public function CMSEditLink()
        $link = CMSEditLinkExtension::get_link($this);
        $this->extend('CMSEditLinkUpdate', $link);
        return $link;
    }
```

GuySartorelli:
Can you please create an issue for this PR which explains what the problem is that you're trying to solve? The screenshot doesn't explain what you're solving, it just shows an error without providing any reproduction steps or an explanation of what's causing the error to occur.

After that, please update this PR - you've indicated that it's "not a good pull request" so please make it good ;p

GuySartorelli:
Closing as the requested changes were not made.

PR #2788, opened 2022-10-10
API Strongly-type action method signatures

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10466

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -23,6 +23,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
+use SilverStripe\Control\PjaxResponseNegotiator;
 use SilverStripe\Core\Cache\MemberCacheFlusher;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
@@ -204,7 +205,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         CMSBatchActionHandler::register('publish', CMSBatchAction_Publish::class);
     }
 
-    public function index($request)
+    public function index(HTTPRequest $request): HTTPResponse
     {
         // In case we're not showing a specific record, explicitly remove any session state,
         // to avoid it being highlighted in the tree, and causing an edit form to show.
@@ -215,7 +216,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         return parent::index($request);
     }
 
-    public function getResponseNegotiator()
+    public function getResponseNegotiator(): PjaxResponseNegotiator
     {
         $negotiator = parent::getResponseNegotiator();
 
@@ -653,11 +654,8 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
     /**
      * Get a subtree underneath the request param 'ID'.
      * If ID = 0, then get the whole tree.
-     *
-     * @param HTTPRequest $request
-     * @return string
      */
-    public function getsubtree($request)
+    public function getsubtree(HTTPRequest $request): HTTPResponse
     {
         $html = $this->getSiteTreeFor(
             $this->config()->get('tree_class'),
@@ -672,7 +670,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         $html = preg_replace('/^[\s\t\r\n]*<ul[^>]*>/', '', $html ?? '');
         $html = preg_replace('/<\/ul[^>]*>[\s\t\r\n]*$/', '', $html ?? '');
 
-        return $html;
+        return $this->getResponse()->setBody($html);
     }
 
     /**
@@ -680,11 +678,8 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
      * Similar to {@link getsubtree()}, but doesn't enforce loading
      * all children with the node. Useful to refresh views after
      * state modifications, e.g. saving a form.
-     *
-     * @param HTTPRequest $request
-     * @return HTTPResponse
      */
-    public function updatetreenodes($request)
+    public function updatetreenodes(HTTPRequest $request): HTTPResponse
     {
         $data = [];
         $ids = explode(',', $request->getVar('ids') ?? '');
@@ -752,17 +747,15 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
      * - 'SiblingIDs': Array of all sibling nodes to the moved node (incl. the node itself).
      *   In case of a 'ParentID' change, relates to the new siblings under the new parent.
      *
-     * @param HTTPRequest $request
-     * @return HTTPResponse JSON string with a
      * @throws HTTPResponse_Exception
      */
-    public function savetreenode($request)
+    public function savetreenode(HTTPRequest $request): HTTPResponse
     {
         if (!SecurityToken::inst()->checkRequest($request)) {
-            return $this->httpError(400);
+            $this->httpError(400);
         }
         if (!$this->CanOrganiseSitetree()) {
-            return $this->httpError(
+            $this->httpError(
                 403,
                 _t(
                     __CLASS__.'.CANT_REORGANISE',
@@ -775,14 +768,14 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         $id = $request->requestVar('ID');
         $parentID = $request->requestVar('ParentID');
         if (!is_numeric($id) || !is_numeric($parentID)) {
-            return $this->httpError(400);
+            $this->httpError(400);
         }
 
         // Check record exists in the DB
         /** @var SiteTree $node */
         $node = DataObject::get_by_id($className, $id);
         if (!$node) {
-            return $this->httpError(
+            $this->httpError(
                 500,
                 _t(
                     __CLASS__.'.PLEASESAVE',
@@ -794,7 +787,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         // Check top level permissions
         $root = $node->getParentType();
         if (($parentID == '0' || $root == 'root') && !SiteConfig::current_site_config()->canCreateTopLevel()) {
-            return $this->httpError(
+            $this->httpError(
                 403,
                 _t(
                     __CLASS__.'.CANT_REORGANISE',
@@ -1026,7 +1019,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         return $pageTypes;
     }
 
-    public function doSearch($data, $form)
+    public function doSearch(array $data, Form $form): HTTPResponse
     {
         return $this->getsubtree($this->getRequest());
     }
@@ -1551,17 +1544,14 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
     /**
      * Callback to request the list of page types allowed under a given page instance.
      * Provides a slower but more precise response over SiteTreeHints
-     *
-     * @param HTTPRequest $request
-     * @return HTTPResponse
      */
-    public function childfilter($request)
+    public function childfilter(HTTPRequest $request): HTTPResponse
     {
         // Check valid parent specified
         $parentID = $request->requestVar('ParentID');
         $parent = SiteTree::get()->byID($parentID);
         if (!$parent || !$parent->exists()) {
-            return $this->httpError(404);
+            $this->httpError(404);
         }
 
         // Build hints specific to this class
@@ -1748,12 +1738,9 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
     /**
      * Save and Publish page handler
      *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
      * @throws HTTPResponse_Exception
      */
-    public function save($data, $form)
+    public function save(array $data, Form $form): HTTPResponse
     {
         $className = $this->config()->get('tree_class');
 
@@ -1896,12 +1883,9 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
      *
      * @uses SiteTree->doRevertToLive()
      *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
      * @throws HTTPResponse_Exception
      */
-    public function revert($data, $form)
+    public function revert(array $data, Form $form): HTTPResponse
     {
         if (!isset($data['ID'])) {
             throw new HTTPResponse_Exception("Please pass an ID in the form content", 400);
@@ -1949,12 +1933,9 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
      *
      * @see deletefromlive()
      *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
      * @throws HTTPResponse_Exception
      */
-    public function delete($data, $form)
+    public function delete(array $data, Form $form): HTTPResponse
     {
         $id = $data['ID'];
         $record = SiteTree::get()->byID($id);
@@ -1984,12 +1965,9 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
     /**
      * Delete this page from both live and stage
      *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
      * @throws HTTPResponse_Exception
      */
-    public function archive($data, $form)
+    public function archive(array $data, Form $form): HTTPResponse
     {
         $id = $data['ID'];
         /** @var SiteTree $record */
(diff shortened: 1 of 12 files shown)


code/Controllers/CMSMain.php
```diff
@@ -672,19 +670,16 @@ public function getsubtree($request)
         $html = preg_replace('/^[\s\t\r\n]*<ul[^>]*>/', '', $html ?? '');
         $html = preg_replace('/<\/ul[^>]*>[\s\t\r\n]*$/', '', $html ?? '');
 
-        return $html;
+        return HTTPResponse::create()->setBody($html);
```
ChloeSartorelli: Why create a new response vs setting the body of the controller's existing response?
I think the existing behaviour of this would be that the body of the existing response would get set - and some of the other actions do that too (see updatetreenodes())

code/Controllers/CMSMain.php
```diff
@@ -775,14 +768,14 @@ public function savetreenode($request)
...
-            return $this->httpError(400);
+            $this->httpError(400);
         }
 
         // Check record exists in the DB
         /** @var SiteTree $node */
         $node = DataObject::get_by_id($className, $id);
         if (!$node) {
-            return $this->httpError(
+            $this->httpError(
```
ChloeSartorelli, line 778: There's another one being `return`ed below - I can't flag it directly but it needs to be updated as well.

code/Controllers/CMSMain.php
```diff
@@ -752,17 +747,15 @@ public function updatetreenodes($request)
      * - 'SiblingIDs': Array of all sibling nodes to the moved node (incl. the node itself).
      *   In case of a 'ParentID' change, relates to the new siblings under the new parent.
      *
-     * @param HTTPRequest $request
-     * @return HTTPResponse JSON string with a
      * @throws HTTPResponse_Exception
      */
-    public function savetreenode($request)
+    public function savetreenode(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 752: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array.

code/Controllers/CMSMain.php
```diff
@@ -1748,12 +1738,9 @@ public function currentPageID()
     /**
      * Save and Publish page handler
      *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
      * @throws HTTPResponse_Exception
      */
-    public function save($data, $form)
+    public function save(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 1743: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSMain.php
```diff
@@ -1896,12 +1883,9 @@ public function performPublish($record)
      *
      * @uses SiteTree->doRevertToLive()
      *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
      * @throws HTTPResponse_Exception
      */
-    public function revert($data, $form)
+    public function revert(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 1888: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSMain.php
```diff
@@ -1949,12 +1933,9 @@ public function revert($data, $form)
      *
      * @see deletefromlive()
      *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
      * @throws HTTPResponse_Exception
      */
-    public function delete($data, $form)
+    public function delete(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 1938: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSMain.php
```diff
@@ -1984,12 +1965,9 @@ public function delete($data, $form)
     /**
      * Delete this page from both live and stage
      *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
      * @throws HTTPResponse_Exception
      */
-    public function archive($data, $form)
+    public function archive(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 1970: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSMain.php
```diff
@@ -2017,14 +1995,14 @@ public function archive($data, $form)
...
-    public function publish($data, $form)
+    public function publish(array $data, Form $form): HTTPResponse
     {
         $data['publish'] = '1';
 
         return $this->save($data, $form);
     }
 
-    public function unpublish($data, $form)
+    public function unpublish(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 2005: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSMain.php
```diff
@@ -2064,12 +2042,8 @@ public function rollback()
 
     /**
      * Rolls a site back to a given version ID
-     *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
      */
-    public function doRollback($data, $form)
+    public function doRollback(array $data, Form $form): HTTPResponse
```
ChloeSartorelli: This method is called by `rollback()` which I think is a form action - and which in any case doesn't pass in a `Form`.

ChloeSartorelli: The fact that that didn't cause CI to fail is slightly worrying - we might need a new test for rollbacks.

ChloeSartorelli: Also this method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSMain.php
```diff
@@ -2161,10 +2135,8 @@ public function BatchActionList()
 
     /**
      * @deprecated 5.0 Please use custom logic for this
-     * @param $request
-     * @return HTTPResponse|string|void
      */
-    public function publishall($request)
+    public function publishall(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 2143: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSMain.php
```diff
@@ -2265,7 +2233,7 @@ public function restore($data, $form)
         return $this->getResponseNegotiator()->respond($this->getRequest());
     }
 
-    public function duplicate($request)
+    public function duplicate(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 2240: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSMain.php
```diff
@@ -2309,11 +2277,11 @@ public function duplicate($request)
         return new HTTPResponse("CMSMain::duplicate() Bad ID: '$id'", 400);
     }
 
-    public function duplicatewithchildren($request)
+    public function duplicatewithchildren(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 2284: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSPageAddController.php
```diff
@@ -192,12 +192,7 @@ public function AddForm()
         return $form;
     }
 
-    /**
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
-     */
-    public function doAdd($data, $form)
+    public function doAdd(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 195: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array

code/Controllers/CMSPageEditController.php
```diff
@@ -58,7 +58,7 @@ public function getClientConfig()
      * @param Form $form
      * @return DBHTMLText|HTTPResponse
      */
-    public function addtocampaign($data, $form)
+    public function addtocampaign(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 57: This method sometimes returns `null`

code/Controllers/ModelAsController.php
```diff
@@ -72,10 +72,8 @@ protected function beforeHandleRequest(HTTPRequest $request)
 
     /**
      * @uses ModelAsController::getNestedController()
-     * @param HTTPRequest $request
-     * @return HTTPResponse
      */
-    public function handleRequest(HTTPRequest $request)
+    public function handleRequest(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 74: This sometimes returns a result directly from an arbitrary `RequestHandler::handleRequest()` which doesn't guarantee an `HTTPResponse` - probably the returned value needs the same treatment that `Controller::handleRequest()` does on those.

code/Forms/SiteTreeURLSegmentField.php
```diff
@@ -68,7 +68,7 @@ public function Field($properties = [])
      * @param HTTPRequest $request
      * @return string
      */
-    public function suggest($request)
+    public function suggest(HTTPRequest $request)
```
ChloeSartorelli: Is it okay that this doesn't return `HTTPResponse`?

code/Search/ContentControllerSearchExtension.php
```diff
@@ -52,7 +54,7 @@ public function SearchForm()
      * @param SearchForm $form The form instance that was submitted
      * @param HTTPRequest $request Request generated for this action
      */
-    public function results($data, $form, $request)
+    public function results(array $data, Form $form): HTTPResponse
```
ChloeSartorelli: I don't think this is actually returning an `HTTPResponse`?
The fact that didn't cause CI to fail means we probably need a new test for this.

GuySartorelli (changes requested):

code/Controllers/CMSMain.php
emteknetnz: `Security::permissionFailure()` does not actually return an array. There is an anonymous function in in there that does return an array

emteknetnz: Updated

code/Controllers/CMSMain.php
emteknetnz: Updated

code/Controllers/CMSMain.php
emteknetnz: As noted above, it does not return an array

code/Controllers/CMSMain.php
emteknetnz: As noted above, it does not return an array

code/Controllers/CMSMain.php
emteknetnz: As noted above, it does not return an array

code/Controllers/CMSMain.php
emteknetnz: As noted above, it does not return an array

code/Controllers/CMSMain.php
emteknetnz: As noted above, it does not return an array

code/Controllers/CMSMain.php
emteknetnz: Have reverted

emteknetnz: As noted above, it does not return an array

code/Controllers/CMSMain.php
emteknetnz: As noted above, it does not return an array

code/Controllers/CMSMain.php
emteknetnz: As noted above, it does not return an array

code/Controllers/CMSMain.php
emteknetnz: As noted above, it does not return an array

code/Controllers/CMSPageAddController.php
emteknetnz: As noted above, it does not return an array

code/Controllers/CMSPageEditController.php
emteknetnz: Updated

code/Controllers/ModelAsController.php
emteknetnz: I've added strong typing to `ModelAsController::controller_for():ContentController` to ensure that the return type on `handleRequest()` is always `HTTPResponse`.  I've removed some excess code from this method too.

The phpdoc on that method says it's only supposed to only return a ContentController, so I'm happy for hard fails here on upgrade if something doesn't meet the new signature

code/Forms/SiteTreeURLSegmentField.php
emteknetnz: It's on a FormField class, should not have changed. I'll just revert this

code/Search/ContentControllerSearchExtension.php
emteknetnz: Reverted

emteknetnz: Regarding this any any other comments saying "we probably need a new test for this" - I'm not going to add these in this round of PRs, and I'd also prefer not to even raise a card to add this in isolation.  Reason being that there's a wider issue of code that lacks unit test coverage, from memory total code coverage was ~70% according to codecov, so it make more sense to me to raise this as a "very large" card, which starts with identifying how large of an issue it is. I don't think it's really the time to look at that, though feel free to raise a card yourself if you feel strongly about it.

code/Controllers/CMSPageEditController.php
```diff
@@ -53,25 +53,18 @@ public function getClientConfig()
...
     {
         $id = $data['ID'];
         $record = \Page::get()->byID($id);
 
         $handler = AddToCampaignHandler::create($this, $record);
         $results = $handler->addToCampaign($record, $data);
-        if (is_null($results)) {
-            return null;
```
ChloeSartorelli, line 69: Is it safe to do this? Shouldn't we check if there's a body to the result response and return the empty response if there is no body (instead of proceeding with the below)?

code/Controllers/ModelAsController.php
```diff
@@ -33,11 +33,9 @@ class ModelAsController extends Controller implements NestedController
      * Get the appropriate {@link ContentController} for handling a {@link SiteTree} object, link it to the object and
      * return it.
      *
-     * @param SiteTree $sitetree
      * @param string $action
-     * @return ContentController
      */
-    public static function controller_for(SiteTree $sitetree, $action = null)
+    public static function controller_for(SiteTree $sitetree, $action = null): ContentController
```
ChloeSartorelli, line 38: If anyone was using `ModelAsController` to get their controller for a non-sitetree this will break - we need to mention it on the changelog (or just not make the change - I don't know that it's valuable to make this change but I'll leave it to you to decide which to do)

ChloeSartorelli: Note: If we do revert this change then https://github.com/silverstripe/silverstripe-cms/pull/2788#discussion_r995163934 will need to be looked at again, but I've marked that resolved for now.

code/Search/ContentControllerSearchExtension.php
```diff
@@ -3,9 +3,11 @@
 namespace SilverStripe\CMS\Search;
 
 use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Extension;
 use SilverStripe\Forms\TextField;
 use SilverStripe\Forms\FieldList;
+use SilverStripe\Forms\Form;
```
ChloeSartorelli:
```suggestion
use SilverStripe\Core\Extension;
use SilverStripe\Forms\TextField;
use SilverStripe\Forms\FieldList;
```

code/Controllers/CMSMain.php
ChloeSartorelli: This hasn't been addressed?

GuySartorelli (changes requested):

code/Controllers/CMSPageEditController.php
emteknetnz: Yes it is safe

`addToCampaign()` had a strong HTTPResponse return type added - https://github.com/silverstripe/silverstripe-campaign-admin/pull/244/files#diff-005db379d07dad4da0ae604f8201249c122b800056b951f9b64b05ef977b7fdcR290

code/Search/ContentControllerSearchExtension.php
emteknetnz: Updated

code/Controllers/ModelAsController.php
emteknetnz: > get their controller for a non-sitetree

There was already strong typing for the `SiteTree $sitetree` param

I do not think this needs to be called out in the changelog

code/Controllers/CMSMain.php
emteknetnz: Updated

code/Controllers/CMSPageEditController.php
ChloeSartorelli: The body might be empty, which means we shouldn't do the processing below I think. We should probably check if `$results->getBody()` is empty, and if it is, just `return $results`.

GuySartorelli (changes requested):

code/Controllers/ModelAsController.php
ChloeSartorelli: Ahh yup for some reason I missed that it was already strongly typed. No worries then.

code/Controllers/CMSPageEditController.php
emteknetnz: Updated

GuySartorelli (approved):

PR #2800, opened 2022-11-28
API Remove deprecated code

sabina-talipova:
### Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10501


lang/ar.yml
```diff
@@ -167,8 +167,6 @@ ar:
     many_many_BackLinkTracking: 'متابعة التعقيبات'
     many_many_ImageTracking: 'متابعة الصور'
     many_many_LinkTracking: 'متابعة الروابط'
-  SilverStripe\CMS\Model\SiteTreeFileExtension:
```
emteknetnz: Can you please restore ALL the translation updates - we cannot delete keys from the yml files.  If we do, when we push the changes to transifex the translations will be deleted from transifex, so if the translations are pulled for older versions of silverstripe, then the translations will be deleted from the yml files of the older version of silverstripe

tests/php/Model/SiteTreeTest.php
```diff
@@ -888,7 +887,7 @@ public function testAuthorIDAndPublisherIDFilledOutOnPublish()
             "SELECT \"AuthorID\", \"PublisherID\" FROM \"SiteTree_Versions\"
             WHERE \"RecordID\" = ? ORDER BY \"Version\" DESC",
             [$about->ID]
-        )->first();
+        )->record();
```
emteknetnz, line 890: I'm surprised this used to work

emteknetnz (changes requested):

tests/php/Model/SiteTreeTest.php
sabina-talipova: Should I return back the method `first()`?

lang/ar.yml
sabina-talipova: Done

tests/php/Model/SiteTreeTest.php
emteknetnz: No it's correct, first() was just a deprecated alias for record() - sorry I should have checked that before commenting

emteknetnz (approved):

PR #2804, opened 2022-12-05
FIX Use orderBy() instead of sort()

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10449

Requires https://github.com/silverstripe/silverstripe-framework/pull/10600 to be merged first - but this was run though an installer CI run: https://github.com/emteknetnz/silverstripe-installer/actions/runs/3617074332


GuySartorelli (approved):
It feels a tad far to call this "Raw SQL" - but given we have control over the table name here I can't see a scenario where this will cause problems, so I'm okay with it.

Don't merge until the framework PR is merged in so we can rerun CI

PR #2806, opened 2022-12-08
FIX Cast absoluteUrl() argument to string

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/642

Similar to https://github.com/silverstripe/silverstripe-elemental/pull/1029

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -637,7 +637,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         if ($this->hasMethod('alternateAbsoluteLink')) {
             return $this->alternateAbsoluteLink($action);
         } else {
-            return Director::absoluteURL($this->Link($action));
+            return Director::absoluteURL((string) $this->Link($action));
         }
     }
 


code/Model/SiteTree.php
```diff
@@ -637,7 +637,7 @@ public function AbsoluteLink($action = null)
         if ($this->hasMethod('alternateAbsoluteLink')) {
             return $this->alternateAbsoluteLink($action);
         } else {
-            return Director::absoluteURL($this->Link($action));
+            return Director::absoluteURL((string) $this->Link($action));
```
ChloeSartorelli, line 640: Under what scenario does `$this->Link` return something other than a string? Looking at the method it should always be returning a link.... unless an extension changes it to something else, but then perhaps we should cast to string before returning from that method instead? I much prefer that solution, where `Link()` always reliably returns a string.

I'll let you respond to this before I look at the other related PRs as I suspect they are all similar in approach.

emteknetnz: This broke in CI, though looks like I forgot to link to the actual failure.

Regardless, $this->Link() can be overloaded by any page type, and has the `updateLink` extension hook and has a dynamic return type, so we cannot guarantee that it always returns a string.

If we don't cast it here, then we need to add a strongly typed `string` return type to `Link()`, which is certainly has more ways of going wrong / increasing upgrade friction

ChloeSartorelli: > has a dynamic return type, so we cannot guarantee that it always returns a string

> $this->Link() can be overloaded by any page type

These are both great reasons to strictly type the return type for `Link()` - but that's probably going to be a CMS 6 thing in the end. I'm okay with this given the time constraints we have to get CMS 5 go-ready.

GuySartorelli (approved):

PR #2812, opened 2023-01-09
PATCH: fixing jquery link to thirdparty/jquery-query/jquery.query.js

sunnysideup:


lozcalver (approved):

GuySartorelli:
jQuery.query is NOT jQuery. As of admin 4.12 jQuery is added via yarn, not stored in the thirdparty dir. 

That reference to jQuery should be removed, we should not be injecting jQuery into the front-end and afaik it is not needed for the Silverstripenavigator functionality. If it is, then imo that functionality should be rewritten to use native js. If that can't be done, then jQuery should be part of the dist js file for Silverstripe navigator. 

If we absolutely must inject jQuery from here, it should be done via a cdn.

lozcalver:
You’re right, it looks like `SilverStripeNavigator.js` doesn’t use this plugin at all, so we can probably just remove this require statement. I also can’t spot anywhere that the JS is used - the class names/IDs referenced don’t exist anymore

sunnysideup:
I have updated the pull request.

GuySartorelli (approved):

GuySartorelli:
Sorry, I'm not sure why this didn't get merged at the time. Merging now.

PR #2820, opened 2023-01-17
ENH Remove unused Translatable code

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10548

Failing unit test is unrelated

### DIFF
--- a/code/Controllers/ContentController.php
+++ b/code/Controllers/ContentController.php
@@ -30,7 +30,6 @@ use SilverStripe\View\ArrayData;
 use SilverStripe\View\Parsers\URLSegmentFilter;
 use SilverStripe\View\Requirements;
 use SilverStripe\View\SSViewer;
-use Translatable;
 
 /**
  * The most common kind of controller; effectively a controller linked to a {@link DataObject}.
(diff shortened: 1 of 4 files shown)


code/Search/SearchForm.php
```diff
@@ -133,18 +121,6 @@ public function getResults()
         // set language (if present)
         $locale = null;
         $origLocale = null;
```
michalkleiner:
```suggestion
```

emteknetnz: Updated

Failing unit test is unrelated

code/Controllers/ContentController.php
```diff
@@ -415,9 +386,7 @@ public function SiteConfig()
 
     /**
      * Returns an RFC1766 compliant locale string, e.g. 'fr-CA'.
-     * Inspects the associated {@link dataRecord} for a {@link SiteTree->Locale} value if present,
-     * and falls back to {@link Translatable::get_current_locale()} or {@link i18n::default_locale()},
-     * depending if Translatable is enabled.
+     * Inspects the associated {@link dataRecord} for a {@link SiteTree->Locale} value if present, and falls back to {@link i18n::default_locale()}
```
ChloeSartorelli: You've changed this to no longer look for a `SiteTree->Locale` value. We should either look for that value (if anything in core would add it) or remove the reference to that in the phpdoc

GuySartorelli (changes requested):

code/Controllers/ContentController.php
emteknetnz: Have removed the phpdoc

GuySartorelli (approved):
Failed unit tests do indeed seem unrelated. Changes look good to me.

PR #2825, opened 2023-01-20
fix: handle exceptions when using /0 as a URL

wilr:
Fix for https://github.com/silverstripe/silverstripe-cms/issues/2680

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -456,7 +456,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         $parentIDExpr = sprintf('"%s"."ParentID"', $tableName);
 
         $link = trim(Director::makeRelative($link) ?? '', '/');
-        if (!$link) {
+        if ($link === false || $link === null || $link === '') {
             $link = RootURLController::get_homepage_link();
         }
 
@@ -1633,6 +1633,11 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         }
     }
 
+    private function hasURLSegment(): bool
+    {
+        return $this->URLSegment !== false && $this->URLSegment !== null && $this->URLSegment !== '';
+    }
+
     protected function onBeforeWrite()
     {
         parent::onBeforeWrite();
@@ -1653,7 +1658,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             'New {pagetype}',
             ['pagetype' => $this->i18n_singular_name()]
         ));
-        if ((!$this->URLSegment || $this->URLSegment == $defaultSegment) && $this->Title) {
+        if ((!$this->hasURLSegment() || $this->URLSegment == $defaultSegment) && $this->Title) {
             $this->URLSegment = $this->generateURLSegment($this->Title);
         } elseif ($this->isChanged('URLSegment', 2)) {
             // Do a strict check on change level, to avoid double encoding caused by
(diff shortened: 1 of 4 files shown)


michalkleiner:
Thanks Will. Since this is fixing a bug, it can target `4.12`.
As per @GuySartorelli's [comment](https://github.com/silverstripe/silverstripe-cms/pull/2685#issuecomment-968432261), the fix may not be as simple as this, so unit test(s) would be needed as well.

shanholmes:
This is still an issue and has since carried over to the 5 branch of the CMS.

GuySartorelli:
Rebased, added tests, and made changes necessary for those tests to pass.

tests/php/Model/SiteTreeTest.php
```diff
@@ -459,6 +460,7 @@ public function testGetByLink()
         $about    = $this->objFromFixture('Page', 'about');
         $staff    = $this->objFromFixture('Page', 'staff');
         $product  = $this->objFromFixture('Page', 'product1');
+        $numeric0  = $this->objFromFixture('Page', 'numeric0');
```
michalkleiner: nitpick
```suggestion
        $numeric0 = $this->objFromFixture('Page', 'numeric0');
```

code/Model/SiteTree.php
```diff
@@ -1633,6 +1633,11 @@ public function requireDefaultRecords()
         }
     }
 
+    private function hasURLSegment(): bool
+    {
+        return $this->URLSegment !== null && $this->URLSegment !== '';
```
michalkleiner: I guess excluding `false` is not necessary here, but would qualify as having a URL segment.

tests/php/Model/SiteTreeTest.php
ChloeSartorelli: Done

tests/php/Model/SiteTreeTest.php
```diff
@@ -489,26 +493,30 @@ public function testGetByLinkAbsolute()
...
```
michalkleiner: nitpick
```suggestion
        $numeric0 = $this->objFromFixture('Page', 'numeric0');
```

tests/php/Model/SiteTreeTest.php
```diff
@@ -489,26 +493,30 @@ public function testGetByLinkAbsolute()
...
         $this->assertEquals($staff->ID, SiteTree::get_by_link(Controller::join_links($base, $staff->Link()), false)->ID);
         $this->assertEquals($product->ID, SiteTree::get_by_link(Controller::join_links($base, $product->Link()), false)->ID);
+        $this->assertEquals($numeric0->ID, SiteTree::get_by_link(Controller::join_links($base, $numeric0->Link()), false)->ID);
     }
 
     public function testRelativeLink()
     {
         $about    = $this->objFromFixture('Page', 'about');
         $staff    = $this->objFromFixture('Page', 'staff');
+        $numeric0    = $this->objFromFixture('Page', 'numeric0');
```
michalkleiner: cs
```suggestion
        $numeric0 = $this->objFromFixture('Page', 'numeric0');
```

michalkleiner (approved):

code/Model/SiteTree.php
```diff
@@ -456,7 +456,7 @@ public static function get_by_link($link, $cache = true)
         $parentIDExpr = sprintf('"%s"."ParentID"', $tableName);
 
         $link = trim(Director::makeRelative($link) ?? '', '/');
-        if (!$link) {
+        if ($link === false || $link === null || $link === '') {
```
ChloeSartorelli, line 459: Note that while `trim()` will only ever return a string, having the full check here avoids issues if we ever remove that trim.

code/Model/SiteTree.php
ChloeSartorelli: I don't think it's necessary since it's unlikely to ever be `false` - but it doesn't hurt to add to avoid that edge case.
Added.

tests/php/Model/SiteTreeTest.php
ChloeSartorelli: Done

tests/php/Model/SiteTreeTest.php
ChloeSartorelli: Done

emteknetnz (approved):
Tested locally, works good

PR #2826, opened 2023-01-23
fix url segment retrieval if it contains "0" (zero)

xini:
I had brute force attacks on multiple SS sites lately, trying to get into "/0/licence.txt". I haven't found out what software they were looking for, but a URL segment of "0" (zero) throws an 500 error in SS. This change fixes that.

### DIFF
--- a/code/Controllers/ModelAsController.php
+++ b/code/Controllers/ModelAsController.php
@@ -119,7 +119,8 @@ class ModelAsController extends Controller implements NestedController
     {
         $request = $this->getRequest();
 
-        if (!$URLSegment = $request->param('URLSegment')) {
+        $URLSegment = (string) $request->param('URLSegment');
+        if ($URLSegment == null || !strlen($URLSegment)) {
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Is this a duplicate of #2825?

code/Controllers/ModelAsController.php
```diff
@@ -119,7 +119,8 @@ public function getNestedController()
     {
         $request = $this->getRequest();
 
-        if (!$URLSegment = $request->param('URLSegment')) {
+        $URLSegment = (string) $request->param('URLSegment');
+        if ($URLSegment == null || !strlen($URLSegment)) {
```
michalkleiner, line 123: It will never be null if you retype it to string the line above.

michalkleiner:
It is a duplicate. I'd suggest we collaborate on one PR rather than splitting the focus into multiple similarly-looking solutions. The main thing we need for #2825 are tests.

xini:
Sorry, I didn't see the other one. all good, thanks.

PR #2827, opened 2023-01-24
MNT Update unit test to work with both HTML4Value and HTML5Value

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/665

Fixes https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/3987666536/jobs/6837802266#step:12:76
`1) SilverStripe\CMS\Tests\Model\SiteTreeTest::testSanitiseExtraMeta with data set #5 ('<link rel="canonical" href="v...ert(1)', '<link rel="canonical" href="v...gy="">', 'Invalid HTML is converted to ...parsed')
SilverStripe\ORM\ValidationException: Custom Meta Tags does not contain valid HTML`

recipe-sink will use HTML5Value (from silverstripe/html5) and installer will use the default HTML4Value

I'm honestly not sure why this only started breaking now

Have tested locally separately using both installer and recipe-kitchen-sink

### DIFF
--- a/tests/php/Model/SiteTreeTest.php
+++ b/tests/php/Model/SiteTreeTest.php
@@ -40,6 +40,9 @@ use SilverStripe\View\Shortcodes\EmbedShortcodeProvider;
 use TractorCow\Fluent\Extension\FluentSiteTreeExtension;
 use const RESOURCES_DIR;
 use SilverStripe\Dev\Deprecation;
+use SilverStripe\HTML5\HTML5Value;
+use SilverStripe\View\Parsers\HTMLValue;
+use SilverStripe\View\Parsers\HTML4Value;
 
 class SiteTreeTest extends SapphireTest
 {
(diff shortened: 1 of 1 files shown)


tests/php/Model/SiteTreeTest.php
```diff
@@ -2060,12 +2057,27 @@ public function provideSanitiseExtraMeta(): array
...
-                '<link rel="canonical" href="valid" ;;// somethingdodgy < onmouseover=alert(1)',
+            ]
+        ];
+    }
+
+    public function testSanatiseInvalidExtraMeta()
+    {
+        $siteTree = new SiteTree();
+        $siteTree->ExtraMeta = '<link rel="canonical" href="valid" ;;// somethingdodgy < onmouseover=alert(1)';
+        if (class_exists(HTML5Value::class)) {
```
maxime-rainville: `HTML5Value` could exists without being the default.

maxime-rainville (changes requested):
This seems like two distinct scenarios to me that should be covered by two distinct tests.

Is it practical to update the config on the spot to swap which parser is used, so that both test are run if the HTML5 module is present? If not, the test should be skip if the appropriate parser is not in used.

tests/php/Model/SiteTreeTest.php
emteknetnz: It makes itself the default - https://github.com/silverstripe/silverstripe-html5/blob/2/_config/html.yml#L6

The only scenario where it's not the default is if someone installs silverstripe/html5 and then manually configures to use HTML4Value ... which is not ever going to be an issue in CI

emteknetnz: Have updated

maxime-rainville (approved):
LGTM

PR #2830, opened 2023-01-27
Resolve a couple of behat issues

GuySartorelli:
The syntax for reactstrap confirm changed - see examples at https://algm.github.io/reactstrap-confirm/?path=/docs/confirm-dialog--alert
This should have been done upgrading initially.

Also behat tests changed to be more specific, so that modules which add additional treedropdown fields don't affect behat tests.

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/644


GuySartorelli:
JS failure is unrelated and will be covered in https://github.com/silverstripe/silverstripe-admin/issues/1421 at some future stage

emteknetnz (approved):

Issue #2816, opened 2023-01-12
ReferenceError: editorIdentifier is not defined


emteknetnz:
This behat failure - https://github.com/silverstripe/silverstripe-cms/actions/runs/3889666737/jobs/6638125642#step:12:538

Running behat test locally with JS with `yarn dev`

`STACKTRACE:
ReferenceError: editorIdentifier is not defined
    at http://bugfix6.test/_resources/vendor/silverstripe/asset-admin/client/dist/js/TinyMCE_sslink-file.js?m=1673495003:24724:4
    at http://bugfix6.test/_resources/vendor/silverstripe/asset-admin/client/dist/js/TinyMCE_sslink-file.js?m=1673495003:24830:2
    at http://bugfix6.test/_resources/vendor/silverstripe/asset-admin/client/dist/js/TinyMCE_sslink-file.js?m=1673495003:24831:12`
    
### PRs
- https://github.com/silverstripe/silverstripe-framework/pull/10665
- https://github.com/silverstripe/silverstripe-cms/pull/2831
- https://github.com/silverstripe/silverstripe-admin/pull/1450
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1327

maxime-rainville:
Note from @GuySartorelli: `editorIdentifier` was added as a global JS variable when upgrading TinyMCE. Looks like there's race condition somewhere.

GuySartorelli:
"global" there meaning it was added as a variable to the tinymce combined js, i.e. it's "global" only in that it's defined in that file, and everything which is part of that file should have access to it. My assumption at the time was that one copy of this is created for each WYSIWYG in a form, but that may not be accurate.

It'll be worth also confirming if this _should_ even be a "global" in the way it has been done - and if not, devise an appropriate way for the plugins that need this information to get it (e.g. fetch it from the appropriate data attribute from the `textarea` element). IIRC this was done originally because the `editorconfig` used to be exposed by the tinymce instance itself but isn't anymore, or at least isn't in the way it had been originally.

Edit: Turns out I didn't make it "global" to the combiner - [that was the case in CMS 4 as well](https://github.com/silverstripe/silverstripe-framework/blob/c430011f19011c52cacd48620e30cc5978de5a50/src/Forms/HTMLEditor/TinyMCECombinedGenerator.php#L139)!

GuySartorelli:
@sabina-talipova I haven't been able to reproduce this from a fresh install. Is this actually a problem?

[...]

PR #2831, opened 2023-01-29
FIX Remove editorIdentifier from global scope

sabina-talipova:
### Description
Removed `var editorIdentifier` from global scope in framework PR https://github.com/silverstripe/silverstripe-framework/pull/10665
### Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2816


sabina-talipova:
Failed JS tests are unrelated to changes were done.

client/src/legacy/TinyMCE_sslink-anchor.js
```diff
@@ -157,6 +158,9 @@ jQuery.entwine('ss', ($) => {
 });
 
 // Adds the plugin class to the list of available TinyMCE plugins
-tinymce.PluginManager.add(commandName, (editor) => plugin.init(editor));
+tinymce.PluginManager.add(commandName, (editor) => { 
+  editorIdentifier = editor.getParam('editorIdentifier');
+  plugin.init(editor)
+});
```
ChloeSartorelli, line 164: This happens too late in the process. By the time we do this, we've _already_ passed in `undefined` to `TinyMCEActionRegistrar.addAction` which means it doesn't know what editor it's creating the action for.

You can verify that by adding `console.log({configId});` into the `addAction` function.

I don't think there's a way to know _which_ editor we want at this stage. It also doesn't seem like there's any downside to _not_ specifying an editor.... so maybe we can just remove `configId` from the `TinyMCEActionRegistrar` altogether.

The only other alternative I can think of is to add the action inside the plugin's `init()` function (right before `addCommand()`). Try this before trying to remove it altogether.

ChloeSartorelli: This comment applies to everywhere else you've done this as well.

GuySartorelli (changes requested):

client/src/legacy/TinyMCE_sslink-anchor.js
sabina-talipova: > The only other alternative I can think of is to add the action inside the plugin's `init()` function (right before `addCommand()`). Try this before trying to remove it altogether.

This doesn't work if user has more than one Content Editor component on the page. I tried few different approaches.

sabina-talipova: > I don't think there's a way to know _which_ editor we want at this stage. It also doesn't seem like there's any downside to _not_ specifying an editor.... so maybe we can just remove `configId` from the `TinyMCEActionRegistrar` altogether.

We cannot get rid of this argument, since if user has it's own WYSIWYG editor configuration and has special name, I think TinyMCEActionRegistrar will not be able to recognise where it should append action.

ChloeSartorelli: > We cannot get rid of this argument, since if user has it's own WYSIWYG editor configuration and has special name, I think TinyMCEActionRegistrar will not be able to recognise where it should append action.

~~My assumption is that this doesn't actually matter. I tried this with multiple tinymce editors on a page and couldn't see any problems resulting from this. Note that the actions have an anonymous function in `onAction` which takes the editor as an argument - that editor will be the correct instance the plugin is being activated from. So at the time the action is actually _performed_ it has the correct instance of the editor. I don't think it needs it outside of that.~~
Ignore that... I think it does matter if, for example, one of the editors should have less options in its action dropdown field (maybe one can't have links to a file for example).

I see also that my init idea doesn't work because `sslink` is then loaded before the other plugins have initialised and therefore it doesn't have their actions loaded in time. The same thing happens if we move that logic down into the `tinymce.PluginManager.add()` anonymous function...

GuySartorelli:
@sabina-talipova Are you actually able to reproduce the original issue with a fresh install of CMS 5? I can't. It might just not be a problem anymore.

sabina-talipova:
> @sabina-talipova Are you actually able to reproduce the original issue with a fresh install of CMS 5? I can't. It might just not be a problem anymore.

I could not reproduce this issue in my local. Probably, it was just warning, the same as we have for `tinymc` variable.
But, it looks like we cannot completely remove `editorIdentifier` without additional investigation. I created two different `TinyMCEConfig` configuration and different identifiers, created page that has its own `Content` block and nested Child item that has its own WYSIWYG with different configuration. But `editorIdentifier` for all WYSIWYG on a page had the same `editorIdentifier` value.
Despite, that `tinymce-cms-xxxxxx.js` file present only one identifier `cms`, it works almost fine with two different configuration and follow instruction that were provided in `_config.php` file for each WYSIWYG. And it looks like we can get rid of this value, since any other configuration just modify `TinyMCEConfig::get('cms')` configuration. 
But I think this identifier was created especially to work with `sslink` tool, but we have some kind of bug or weakness in this tool, since this identifier is not unique and both `textarea` have the same `editorIdentifier`.  It prevents (sometime) users to create or add different set of `sslink` menu items (User should refresh page to see short list). 
I think for now we can close this issue and I crate new ticket about the problem that I describe above.

GuySartorelli:
I agree that it sounds like what you're describing is a separate issue - let's close this one and create a new one as you suggest. IT definitely sounds like it can be dealt with post-beta.

> I created two different TinyMCEConfig configuration and different identifiers, created page that has its own Content block and nested Child item that has its own WYSIWYG with different configuration. But editorIdentifier for all WYSIWYG on a page had the same editorIdentifier value.

In your new issue, can you please include the code you used to test that? It will make it a lot easier to reproduce.

> It prevents (sometime) users to create or add different set of sslink menu items (User should refresh page to see short list).

Can you please also include clear steps to reproduce that? I wasn't able to see that happen in my testing.

sabina-talipova:
> Can you please also include clear steps to reproduce that? I wasn't able to see that happen in my testing.
> In your new issue, can you please include the code you used to test that? It will make it a lot easier to reproduce.

Yep! I'll do and attach link to this issue.

sabina-talipova:
CLOSE

PR #2840, opened 2023-03-01
API Mark moved classes in silverstripe-admin module as deprecated

sabina-talipova:
### Description
The following classes have been moved from silverstripe/cms module in silverstripe/admin. They are no longer exist in cms module and were marked as deprecated. Direction, about classes that replace deprecated classes and should be used instead,  was provided.
- SilverStripe\CMS\Controllers\SilverStripeNavigator
- SilverStripe\CMS\Controllers\SilverStripeNavigatorItem
- SilverStripe\CMS\Controllers\SilverStripeNavigatorItem_Unversioned
- SilverStripe\CMS\Controllers\SilverStripeNavigatorItem_ArchiveLink
- SilverStripe\CMS\Controllers\SilverStripeNavigatorItem_LiveLink
- SilverStripe\CMS\Controllers\SilverStripeNavigatorItem_StageLink

### Parent issue
- https://github.com/silverstripeltd/product-issues/issues/699

### DIFF
--- a/code/Controllers/SilverStripeNavigator.php
+++ b/code/Controllers/SilverStripeNavigator.php
@@ -4,6 +4,7 @@ namespace SilverStripe\CMS\Controllers;
 
 use SilverStripe\ORM\CMSPreviewable;
 use SilverStripe\Core\ClassInfo;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\SS_List;
@@ -18,6 +19,9 @@ use SilverStripe\View\ViewableData;
  * New item types can be defined by extending the {@link SilverStripeNavigatorItem} class,
  * for example the "cmsworkflow" module defines a new "future state" item with a date selector
  * to view embargoed data at a future point in time. So the item doesn't always have to be a simple link.
+ *
+ * Class will be moved from `silverstripe/cms` to `silverstripe/admin`
+ * @deprecated 4.13.0 Will be renamed SilverStripe\Admin\Navigator\SilverStripeNavigator
  */
 class SilverStripeNavigator extends ViewableData
 {
@@ -32,6 +36,13 @@ class SilverStripeNavigator extends ViewableData
      */
     public function __construct(CMSPreviewable $record)
     {
+        Deprecation::withNoReplacement(function () {
+            Deprecation::notice(
+                '4.13.0',
+                'Will be renamed SilverStripe\Admin\Navigator\SilverStripeNavigator',
+                Deprecation::SCOPE_CLASS
+            );
+        });
         parent::__construct();
         $this->record = $record;
     }
--- a/code/Controllers/SilverStripeNavigatorItem.php
+++ b/code/Controllers/SilverStripeNavigatorItem.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\CMS\Controllers;
 
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\ORM\CMSPreviewable;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\Versioned\Versioned;
@@ -12,6 +13,9 @@ use SilverStripe\View\ViewableData;
  * Navigator items are links that appear in the $SilverStripeNavigator bar.
  * To add an item, extend this class - it will be automatically picked up.
  * When instanciating items manually, please ensure to call {@link canView()}.
+ *
+ * Class have been moved from `silverstripe/cms` to `silverstripe/admin` and renamed.
+ * @deprecated Will be renamed SilverStripe\Admin\Navigator\SilverStripeNavigatorItem
  */
 abstract class SilverStripeNavigatorItem extends ViewableData
 {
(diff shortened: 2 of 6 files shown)


code/Controllers/SilverStripeNavigator.php
```diff
@@ -18,6 +19,9 @@
  * New item types can be defined by extending the {@link SilverStripeNavigatorItem} class,
  * for example the "cmsworkflow" module defines a new "future state" item with a date selector
  * to view embargoed data at a future point in time. So the item doesn't always have to be a simple link.
+ *
+ * Class have been moved from `silverstripe/cms` to `silverstripe/admin` and renamed.
+ * @deprecated 4.13.0 Use SilverStripe\Admin\Navigator\SilverStripeNavigator instead
```
ChloeSartorelli:
```suggestion
 * @deprecated 4.13.0 Will be renamed SilverStripe\Admin\Navigator\SilverStripeNavigator
```
Or "Will be moved to" - I'm not sure which is better, so either will do.
But saying "Use [the new name] instead" is misleading since that's not available in CMS 4.

Please make the same change for the deprecation notice itself, and for all of the other classes.

code/Controllers/SilverStripeNavigator.php
```diff
@@ -32,6 +36,11 @@ class SilverStripeNavigator extends ViewableData
      */
     public function __construct(CMSPreviewable $record)
     {
+        Deprecation::notice(
```
ChloeSartorelli: Please wrap this in a call to `Deprecation::withNoReplacement()` since we can't avoid calling this in CMS 4 and don't want to flood logs with it. Same with the other classes.

code/Controllers/SilverStripeNavigatorItem.php
```diff
@@ -29,6 +33,12 @@ abstract class SilverStripeNavigatorItem extends ViewableData
      */
     public function __construct(CMSPreviewable $record)
     {
+        Deprecation::notice(
+            '4.13.0',
+            'Use SilverStripe\Admin\Navigator\SilverStripeNavigatorItem instead',
+            Deprecation::SCOPE_CLASS
+        );
```
ChloeSartorelli:
```suggestion
        Deprecation::withNoReplacement(function () {
            switch (static::class) {
                case 'SilverStripe\CMS\Controllers\SilverStripeNavigatorItem_Unversioned':
                case 'SilverStripe\CMS\Controllers\SilverStripeNavigatorItem_ArchiveLink':
                case 'SilverStripe\CMS\Controllers\SilverStripeNavigatorItem_LiveLink':
                case 'SilverStripe\CMS\Controllers\SilverStripeNavigatorItem_StageLink':
                    // These will have their own appropriate deprecation notices
                    break;
                default:
                    Deprecation::notice(
                        '4.13.0',
                        'Will be renamed SilverStripe\Admin\Navigator\SilverStripeNavigatorItem',
                        Deprecation::SCOPE_CLASS
                    );
            }
        });
```
Each class should have it's _own_ appropriate deprecation notice, stating the correct new FQCN.

code/Controllers/SilverStripeNavigatorItem_ArchiveLink.php
```diff
@@ -9,6 +9,10 @@
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\Versioned\Versioned;
 
+/**
+ * Class have been moved from `silverstripe/cms` to `silverstripe/admin` and renamed.
+ * @deprecated Use SilverStripe\VersionedAdmin\Navigator\SilverStripeNavigatorItem_ArchiveLink instead
+ */
 class SilverStripeNavigatorItem_ArchiveLink extends SilverStripeNavigatorItem
```
ChloeSartorelli, line 18: Add a new constructor for this with the appropriate deprecation notice
```php
public function __construct(CMSPreviewable $record)
{
    /* deprecation notice here */
    parent::__construct($record);
}
```
Same with all the other `SilverStripeNavigatorItem` subclasses.

GuySartorelli (changes requested):

code/Controllers/SilverStripeNavigator.php
michalkleiner: I would prefer "Will be moved to" since the (short) name of the class stays, in this case only the namespace changes.

sabina-talipova: DONE

code/Controllers/SilverStripeNavigator.php
sabina-talipova: DONE

code/Controllers/SilverStripeNavigatorItem.php
sabina-talipova: DONE

code/Controllers/SilverStripeNavigatorItem_ArchiveLink.php
sabina-talipova: DONE

code/Controllers/SilverStripeNavigator.php
ChloeSartorelli: Still need to update the message in all of the deprecation notices, as well. The deprecation notice should have the same message as the deprecation annotation.

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
```diff
@@ -5,11 +5,30 @@
...
+ * @deprecated 4.13.0 Will be renamed SilverStripe\Admin\Navigator\SilverStripeNavigatorItem_Unversioned
+ */
 class SilverStripeNavigatorItem_Unversioned extends SilverStripeNavigatorItem
 {
+    /**
+     * @param DataObject|CMSPreviewable $record
+     */
+    public function __construct(CMSPreviewable $record)
+    {
+        Deprecation::notice(
```
ChloeSartorelli: Wrap in `Deprecation::withNoReplacement(` - _all_ of these notices need to be wrapped like that, because we don't have a replacement in CMS 4.

GuySartorelli (changes requested):

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
```diff
@@ -5,11 +5,30 @@
 use SilverStripe\CMS\Controllers\SilverStripeNavigatorItem;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
+use SilverStripe\Dev\Deprecation;
+use SilverStripe\ORM\CMSPreviewable;
```
ChloeSartorelli, line 9: Why has `CMSPreviewable` been added?

sabina-talipova: For strong typing ` __constructor(CMSPreviewable $record)` argument

code/Controllers/SilverStripeNavigatorItem_Unversioned.php
sabina-talipova: DONE

code/Controllers/SilverStripeNavigator.php
```diff
@@ -32,6 +36,13 @@ class SilverStripeNavigator extends ViewableData
      */
     public function __construct(CMSPreviewable $record)
     {
+        Deprecation::withNoReplacement(function () {
+            Deprecation::notice(
+                '4.13.0',
+                'Use SilverStripe\Admin\Navigator\SilverStripeNavigator instead',
```
ChloeSartorelli: This message still needs to be updated.

GuySartorelli (changes requested):

GuySartorelli (approved):

GuySartorelli:
Looks good, merge on green

PR #2845, opened 2023-03-28
FIX undefined index error in CMS

NightJar:
With the CMS 4.12 update functionality was altered to utilise an Extension to obtain the CMS Edit link for a page, rather than having SiteTree do it internally. Unfortunately the default return case for `extend` (see Extensible) is an _empty_ array. This leaves code potentially referencing an array offset that doesn't exist ([0]). PHP 8 is less forgiving than its predecessors on this kind of behaviour. We should check that the responses from extensions exist before trying to reference them.

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -739,13 +739,15 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     /**
      * Generates a link to edit this page in the CMS.
      *
+     * Implemented here to satisfy the CMSPreviewable interface, but data is intended to be loaded via Extension
+     *
+     * @see SilverStripe\Admin\CMSEditLinkExtension
+     *
      * @return string
      */
     public function CMSEditLink()
     {
-        // This method has to be implemented here to satisfy the CMSPreviewable interface.
-        // See the actual implementation in CMSEditLinkExtension.
-        return $this->extend('CMSEditLink')[0];
+        return $this->extend('CMSEditLink')[0] ?? '';
     }
 
     /**


code/Model/SiteTree.php
```diff
@@ -739,13 +739,16 @@ public function getAbsoluteLiveLink($includeStageEqualsLive = true)
...
+     *
      * @return string
      */
     public function CMSEditLink()
     {
-        // This method has to be implemented here to satisfy the CMSPreviewable interface.
-        // See the actual implementation in CMSEditLinkExtension.
-        return $this->extend('CMSEditLink')[0];
+        $extensionsResult = $this->extend('CMSEditLink');
+        return count($extensionsResult) ? $extensionsResult[0] : '';
```
```suggestion
        return $extensionsResult[0] ?? '';
```
This is just a little clearer - as we don't really care about the number of items in the array so much as we care that there is a value in the `[0]` index.

GuySartorelli (changes requested):
Thank you, this is great.
Just a small change.

code/Model/SiteTree.php
NightJar: `extend` _always_ returns an `array`, never `null`.
***
Wait sorry, didn't see the dereference in the suggestion.
Does PHP not error on that still, as it's accessing an undefined index? Or does `??` squash errors?

ChloeSartorelli: `??` squashes the error - basically says "If there's nothing there, use the next value"
It's called "null coalescing operator" but it actually does an `isset()` check.
https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op gives this as an explicitly recommended use case.

NightJar: Neat thanks, I knew it was null coalescing, but always forget if it also handles the "null pointer" reference too.

code/Model/SiteTree.php
```diff
@@ -739,13 +739,15 @@ public function getAbsoluteLiveLink($includeStageEqualsLive = true)
...
+     * @see SilverStripe\Admin\CMSEditLinkExtension
+     *
      * @return string
      */
     public function CMSEditLink()
     {
-        // This method has to be implemented here to satisfy the CMSPreviewable interface.
-        // See the actual implementation in CMSEditLinkExtension.
-        return $this->extend('CMSEditLink')[0];
+        return $this->extend('CMSEditLink') ?? '';
```
```suggestion
        return $this->extend('CMSEditLink')[0] ?? '';
```
Still need to get the 0th result

GuySartorelli (changes requested):

code/Model/SiteTree.php
NightJar: oops haha 😅

NightJar:
Is there anything left to do here?
Can you rerun the behat task please? I feel like this failure is due to typical behat test flakeyness, not this change.

GuySartorelli (approved):
Sorry, I got distracted and lost sight of this. Thank you for submitting it!

NightJar:
Thanks for restarting the automated test run, it's a relief to see it all pass :)

PR #2849, opened 2023-04-18
MNT Use React Testing Library

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1419


package.json
```diff
@@ -34,6 +34,7 @@
   "dependencies": {
     "@apollo/client": "^3.7.1",
     "@popperjs/core": "^2.11.6",
+    "@testing-library/react": "^14.0.0",
```
ChloeSartorelli: Testing library should be added as a dev dependency

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
```diff
@@ -1,103 +1,110 @@
...
+    },
+    pageId: 4,
+    anchors: ['anchor1', 'anchor2'],
+    value: 'selectedanchor',
+    loadingState: anchorSelectorStates.SUCCESS,
+    actions: {
+      anchorSelector: {
+        beingUpdating: () => {},
+        updated: () => {},
+        updateFailed: () => {},
```
ChloeSartorelli: These all get overwritten in every test, so no sense adding them here. Though their being here isn't a _problem_ per se so I won't block merging for this.

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
ChloeSartorelli, line 1: The `Merges value with page anchors` test seems to be missing?

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
```diff
@@ -1,103 +1,110 @@
...
-            .toHaveBeenCalledWith(4);
-          expect(props.actions.anchorSelector.updated)
-            .toHaveBeenCalledWith(4, ['anchor3', 'anchor4']);
-          expect(props.actions.anchorSelector.updateFailed)
-            .not
-            .toHaveBeenCalled();
-          expect(result).toEqual(['anchor3', 'anchor4']);
-        });
-    });
+test('AnchorSelectorField ensurePagesLoaded Triggers loading on dirty', async () => {
```
ChloeSartorelli, line 89: The `expect(result).toEqual(['anchor3', 'anchor4']);` part of this test is missing.
It's not super clear to me what that was testing, but it does seem apparent that whatever it was isn't being tested here.

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
```diff
@@ -1,103 +1,110 @@
...
-            .not
-            .toHaveBeenCalled();
-          expect(props.actions.anchorSelector.updateFailed)
-            .not
-            .toHaveBeenCalled();
-          expect(result).toBe(undefined);
-        });
-    });
-  });
+test('AnchorSelectorField ensurePagesLoaded Does not trigger updating', async () => {
```
ChloeSartorelli, line 110: The `expect(result).toBe(undefined);` part of this test is missing. As above, it's not clear what that was testing exactly but we should probably either explicitly state what it was and why it _shouldn't_ be tested, or find a way to test it.

GuySartorelli (changes requested):

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
emteknetnz: Added in

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
emteknetnz: Removed

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
emteknetnz: It is being tested - `expect(updated).toBeCalledWith(4, ['anchor3', 'anchor4']);`

The difference is how it's being tested, ReactTestUtils (one of the old libraries) was calling a class method `ensurePagesLoaded()` directly, which you cannot do with testing-library.

Instead we're testing that the `updated` mock callback (`jest.fn()`) was called with particular params

package.json
emteknetnz: Updated

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
emteknetnz: `expect(result).toBe(undefined);` was essentially testing the return value of the class method `ensurePagesLoaded()

We're not testing the class method directly anymore

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
ChloeSartorelli: So the `result` value is definitely always exactly the same as the second argument passed into the updated callback?

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
ChloeSartorelli: > We're not testing the class method directly anymore

That's my point exactly. I haven't looked at the source code but I assume that method is called within the component, and its return value is used somewhere for some purpose, and therefore that value is important.

If that's not the case then fine, but if that is the case then we need to find some way to test that this is returning the correct value - or that the end result of whatever the value is used for is correct.

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
ChloeSartorelli: Upon further discussion offline, it has been decided that there would likely be more effort required to either validate this or implement a way to continue testing what was being tested than it's worth.

client/src/components/AnchorSelectorField/tests/AnchorSelectorField-test.js
ChloeSartorelli: Upon further discussion offline, it has been decided that there would likely be more effort required to either validate this or implement a way to continue testing what was being tested than it's worth.

GuySartorelli (approved):

Issue #2853, opened 2023-05-02
admin jQuery is gone, $SilverStripeNavigator breaks site for authors, devs, etc.


NightJar:
Not sure what to do with this. I'm not a fan of remote script inclusion, as it has knock-on effects with e.g. CORS, etc.
I have [worked around it](https://gist.github.com/NightJar/7d3c0053c89f6fbc6c7de9dccfb32d83) instead, should anyone else want a 'drop in' solution for the time being.

https://github.com/silverstripe/silverstripe-cms/blob/2afad17237080b7335834b2a20bf0164bdeccad4/code/Controllers/ContentController.php#L324

## PR
- https://github.com/silverstripe/silverstripe-cms/pull/2854

GuySartorelli:
Thanks for opening this. It has been in the back of my mind for a while but this has lit a fire under me, PR incoming.

PR #2854, opened 2023-05-04
FIX Don't require jquery on the frontend

GuySartorelli:
None of this javascript is actually used in the built-in silverstripe navigator functionality anymore - but someone might be relying on this javascript being there, so just removing the file would be a breaking change. Instead, I've rewritten the jquery functionality using pure javascript.

You can test it using this in a template. The new js should behave the same way the old js did in 4.11. Note that the popup css puts the popup above the visible area of the window by default, so I've added an `!important` style to the element below to counteract this for testing purposes.

```html
$SilverStripeNavigator

<div id="SilverStripeNavigatorLink">
    <a>This is a link thing apparently?</a>
</div>
<div id="SilverStripeNavigatorLinkPopup" style="top:0!important;">
    This is the popup! It'll pop up.
    <a class="close">close</a>
    <br>
    <input value="this test will be selected when you click the input">
</div>
```

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2853


emteknetnz (approved):
Tested locally, seems fine

emteknetnz:
Released as [4.12.2](https://github.com/silverstripe/silverstripe-cms/releases/tag/4.12.2) and [4.13.1](https://github.com/silverstripe/silverstripe-cms/releases/tag/4.13.1)

Issue #2609, opened 2020-11-05
Duplicating a page does not duplicate viewer/editor groups


maxime-rainville:
# Steps to reproduce
- Create a page.
- Set some viewer groups and editor groups
- Duplicate the page

Expected result: The same groups are assigned as editors and viewers.
Actual result: No group is assigned.

## Acceptance criteria
- When duplicating a page, the duplicate retains the same canView and canEdit groups.
- Fix is targeted to the next minor of CMS 5 and is called out in the changelog. 

## PR
- https://github.com/silverstripe/silverstripe-cms/pull/2859
- https://github.com/silverstripe/silverstripe-framework/pull/10806
- https://github.com/silverstripe/developer-docs/pull/282
- https://github.com/silverstripe/silverstripe-cms/pull/2858
- https://github.com/silverstripe/developer-docs/pull/280

emteknetnz:
PRs have been merged

PR #2858, opened 2023-06-07
FIX Duplicate page keeps original pages canView and canEdit permission

sabina-talipova:
## Description
- Add property `cascade_duplicates` to SiteTree class with specified relation names to duplicate and save against the new clone object. 

## Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2609

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -270,6 +270,11 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         'VirtualPages',
     ];
 
+    private static $cascade_duplicates = [
+        'ViewerGroups',
+        'EditorGroups',
+    ];
+
     private static $casting = [
         "Breadcrumbs" => "HTMLFragment",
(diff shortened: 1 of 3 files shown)


code/Model/SiteTree.php
```diff
@@ -270,6 +270,11 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         'VirtualPages',
     ];
 
+    private static $cascade_duplicates = [
```
emteknetnz, line 273: Seems like this should actually go directly on `InheritedPermissionsExtension` rather than SiteTree?

emteknetnz (changes requested):
Should target `5` rather than `5.0` since it's a behavioral change

code/Model/SiteTree.php
sabina-talipova: Updated:
- https://github.com/silverstripe/silverstripe-framework/pull/10806

PR #2862, opened 2023-06-15
FEAT update SiteTree permissions in CMS

andrewandante:
Linked to https://github.com/silverstripe/silverstripe-framework/pull/10819

Allows `SiteTree` objects to have permissions assigned to a individual users, rather than having to assign it directly to groups.

## Issue
- silverstripe/silverstripe-framework#10817

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -35,6 +35,7 @@ use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\GridField\GridFieldDataColumns;
 use SilverStripe\Forms\GridField\GridFieldLazyLoader;
 use SilverStripe\Forms\HTMLEditor\HTMLEditorField;
+use SilverStripe\Forms\ListboxField;
 use SilverStripe\Forms\LiteralField;
 use SilverStripe\Forms\OptionsetField;
 use SilverStripe\Forms\Tab;
@@ -1187,6 +1188,14 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
             return true;
         }
 
+        // check for specific users
+        if ($this->CanViewType === InheritedPermissions::ONLY_THESE_MEMBERS
+            && $member
+            && $this->ViewerMembers()->filter('ID', $member->ID)->count() > 0
+        ) {
+            return true;
+        }
+
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


andrewandante:
This is now all working alongside the Framework PR; however, the `ViewerMembers` and `EditorMembers` fields always show in the CMS, instead of dynamically show/hide depending on the `ViewType`/`EditType`. Can't seem to pinpoint what is doing that, any tips?

EDIT: Oh no, it's entwine? https://github.com/silverstripe/silverstripe-cms/pull/1747/files#diff-f8b125682490aaedb06618d1784af3223e62462cade2ad266673d6688399ca56R205-R237

EDIT EDIT: Fixed over here: https://github.com/silverstripe/silverstripe-admin/pull/1531/files

GuySartorelli:
There's a couple of PHP linting issues in the CI run which need to be ironed out, please

code/Model/SiteTree.php
```diff
@@ -1187,6 +1188,14 @@ public function canView($member = null)
             return true;
         }
 
+        // check for specific users
+        if ($this->CanViewType === InheritedPermissions::ONLY_THESE_MEMBERS
+            && $member
+            && \in_array($member->ID, $this->ViewerMembers()->column('ID'), true)
```
```suggestion
            && $this->ViewerMembers()->filter('ID', $member->ID)->count() > 0
```
Slightly more efficient, both the db query itself and not having to iterate over IDs. It won't be significant in 99% of cases, but I can imagine someone adding _all_ of the members from a large organisation into this field at some point.

code/Model/SiteTree.php
```diff
@@ -2343,17 +2367,27 @@ public function getSettingsFields()
 
         if (!Permission::check('SITETREE_GRANT_ACCESS')) {
             $fields->makeFieldReadonly($viewersOptionsField);
-            if ($this->CanEditType === InheritedPermissions::ONLY_THESE_USERS) {
+            if ($this->CanViewType === InheritedPermissions::ONLY_THESE_USERS) {
```
ChloeSartorelli, line 2370: Good catch on this bug 👍

GuySartorelli (changes requested):
Just one small change here, otherwise it all looks okay to me

GuySartorelli (changes requested):
For some reason when I try this locally, the `Viewer Users` and `Editor Users` fields don't save. It's probably more to do with the changes you made to ListboxField than anything here, but given that's merged already I'll put the note here instead.
Can you please investigate and assuming you can reproduce this, find out where that bug's coming from? The value _is_ being passed in as part of the form submission request, I can see that from my browser.

It does work fine in react contexts - just the entwine context that's borked.

andrewandante:
> Can you please investigate and assuming you can reproduce this, find out where that bug's coming from?

Yep can reproduce, I saw it last night and have spent a decent chunk of today implementing XDebug in a new dev environment 😅 so I'll see what I can find - I think it's likely do with the validation step. Thanks for all your reviewing thus far!

andrewandante:
[This Framework commit](https://github.com/silverstripe/silverstripe-framework/pull/10819/commits/a03d0fdf684e00d388b67af6ce9bdc99877709af) I think fixes the issue - locally I now have both the entwine and react versions working 🤞

GuySartorelli (approved):
LGTM

PR #2870, opened 2023-07-28
drop duplicated call

### LINKED ISSUE #2869: LeftAndMainPageIconsExtension::generatePageIconsCss is called twice
Version : ss5

Description
---

Not a big deal, but LeftAndMainPageIconsExtension::generatePageIconsCss is called twice

once due to the init method in controller->extend('init')

https://github.com/silverstripe/silverstripe-admin/blob/c0ed41ce469d7c9edc2df9d4df86fb6eba732a33/code/LeftAndMain.php#L747
https://github.com/silverstripe/silverstripe-cms/blob/c831b1be6b2ad35b26f8b0061ffef78f32a9d769/code/Controllers/LeftAndMainPageIconsExtension.php#L27-L30

once hardcoded in cmsmain
https://github.com/silverstripe/silverstripe-cms/blob/c831b1be6b2ad35b26f8b0061ffef78f32a9d769/code/Controllers/CMSMain.php#L190

Solution
---

Drop the line in cmsmain

PR
---

https://github.com/silverstripe/silverstripe-cms/pull/2870

lekoala:
fixes https://github.com/silverstripe/silverstripe-cms/issues/2869

already called due to init being called in the extension by leftandmain

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -187,7 +187,6 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         Requirements::javascript('silverstripe/cms: client/dist/js/bundle.js');
         Requirements::javascript('silverstripe/cms: client/dist/js/SilverStripeNavigator.js');
         Requirements::css('silverstripe/cms: client/dist/styles/bundle.css');
-        Requirements::customCSS($this->generatePageIconsCss(), self::PAGE_ICONS_ID);
 
         Requirements::add_i18n_javascript('silverstripe/cms: client/lang', false);
 


GuySartorelli:
> already called due to init being called in the extension by leftandmain

Can you please just explain that a little clearer? I'm not sure what extension this refers to, nor why calling init would result in this _already_ being called - since this _is in_ init? Are you saying that this line exists somewhere else as well? I don't understand.

lekoala:
i have explained it here https://github.com/silverstripe/silverstripe-cms/issues/2869

GuySartorelli:
Ahh, sorry, I didn't notice the link to the issue 😅

GuySartorelli (approved):
Makes sense to me. Didn't know that extension even existed but it's applied directly to `LeftAndMain` so this seems sensible

Issue #2875, opened 2023-08-22
SiteTree link tracking can cause fatal error when using “Cancel draft changes”


lozcalver:
To reproduce:

```php
<?php

class SiteConfigExtension extends Extension
{
    private static $has_one = [
        'Page' => SiteTree::class
    ];
}
```

Assign the link in the CMS, edit the page you chose and save as draft, then click More Actions > Cancel draft changes.

Error is:

> [Emergency] Uncaught BadMethodCallException: Object->__call(): the method 'writeWithoutVersion' does not exist on 'SilverStripe\SiteConfig\SiteConfig'

The call comes from this block here: https://github.com/silverstripe/silverstripe-cms/blob/14037a776fb10f04b78a7e5de3f5fa9af5680d19/code/Model/SiteTree.php#L2613-L2617

It seems that `SiteTree::DependentPages()` returns non-SiteTree items, which goes against it’s PHPDoc return type of `ArrayList|SiteTree[]`. The non-SiteTree items are added to the result here: https://github.com/silverstripe/silverstripe-cms/blob/14037a776fb10f04b78a7e5de3f5fa9af5680d19/code/Model/SiteTree.php#L2013-L2016

One potential fix is to change the above to:

```php
// We merge all into a regular SS_List, because DataList doesn't support merge
$contentLinks = $this->BackLinkTracking()->filterByCallback(function ($item) {
    return $item instanceof SiteTree;
});
if ($contentLinks->count()) {
    $items->merge($contentLinks);
}
```

## PR
- https://github.com/silverstripe/silverstripe-cms/pull/2879

sabina-talipova:
Hi @kinglozzer,
I'm trying to reproduce this issue in my local env, but I couldn't reproduce this issue neither in CMS 5 or CMS 4.13. All steps work correctly. 
Could you confirm that the following steps is correct, please. 
1. I have the following code:
```php
<?php
namespace App\Extensions;

use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\Core\Extension;

class SiteConfigExtension extends Extension
{
    private static $has_one = [
        'Page' => SiteTree::class,
    ];
}
```  
```yml
SilverStripe\SiteConfig\SiteConfig:
  extensions:
    - App\Extensions\SiteConfigExtension
```
2. I create Page and Publish it.
3. I add this Page in SiteConfig section.
4. I make changes on Page and Save it.
5. I click More Actions > Cancel draft changes.
[...]

PR #2879, opened 2023-08-31
FIX SiteTree::DependentPages method returns non-SiteTree instance

sabina-talipova:
## Description
Check instance is type of SiteTree before invoke method.
## Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2875

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -2612,8 +2612,10 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
 
         // Need to update pages linking to this one as no longer broken
         foreach ($stageSelf->DependentPages() as $page) {
-            /** @var SiteTree $page */
-            $page->writeWithoutVersion();
+            if ($page->hasExtension(Versioned::class)) {
+                /** @var Versioned $page */
+                $page->writeWithoutVersion();
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


code/Model/SiteTree.php
```diff
@@ -2612,8 +2612,10 @@ public function onAfterRevertToLive()
 
         // Need to update pages linking to this one as no longer broken
         foreach ($stageSelf->DependentPages() as $page) {
-            /** @var SiteTree $page */
-            $page->writeWithoutVersion();
+            if ($page instanceof SiteTree) {
+                /** @var SiteTree $page */
+                $page->writeWithoutVersion();
+            }
```
```suggestion
            if ($page->hasExtension(Versioned::class)) {
                /** @var Versioned $page */
                $page->writeWithoutVersion();
            }
```
That method is provided by the `Versioned` extension, so lets allow this to work for any versioned record instead. This will avoid regressions if people were doing funky things like having `DataObject` records that behave like pages but aren't `SiteTree` subclasses explicitly.

GuySartorelli (changes requested):

code/Model/SiteTree.php
sabina-talipova: Updated

lozcalver (approved):
I think this is the best fix, my suggested fix in the issue (patching `SiteTree::DependentPages()`) would’ve had the knock-on effect of stopping things like SiteConfig appearing in the “Dependent pages” tab in the CMS. The PHPDoc for that method is incorrect, but it appears to work, so let’s leave well-enough alone 😄

GuySartorelli (approved):
Works as expected locally.

PR #2880, opened 2023-09-01
FIX Cannot link an image in TinyMCE

sabina-talipova:
## Description
We should check that user selected an element and if element doesn't have inner content return outer HTML. 
## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1548


client/src/legacy/TinyMCE_sslink-anchor.js
sabina-talipova, line 1: I copied changes from CMS 5.

sabina-talipova:
I copied changes from CMS 5.

sabina-talipova:
[This PR](https://github.com/silverstripe/silverstripe-admin/pull/1571) is required to pass test

GuySartorelli:
Superceded by https://github.com/silverstripe/silverstripe-cms/pull/2883

Issue #2881, opened 2023-09-05
VirtualPage uses hardcoded 'admin' in the edit link


mirche97:
VirtualPage uses hardcoded 'admin' in the edit link:

https://github.com/silverstripe/silverstripe-cms/blob/5/code/Model/VirtualPage.php#L249

This breaks if AdminRootController's url_base is changed.

## PRs
- https://github.com/silverstripe/silverstripe-cms/pull/2882
- https://github.com/silverstripe/silverstripe-cms/pull/2888

GuySartorelli:
Hi,

Thanks for reporting that. Would you like to try raising a pull request to fix this bug?

mirche97:
Yes, I have already done it in https://github.com/silverstripe/silverstripe-cms/pull/2882

GuySartorelli:
A fix has been merged and will be automatically tagged by GitHub Actions

PR #2882, opened 2023-09-05
fixed VirtualPage generating incorrect CMS edit links

mirche97:
VirtualPage  now uses AdminRootController::get_admin_route() to generate edit links

Fixes the following issue: https://github.com/silverstripe/silverstripe-cms/issues/2881

### DIFF
--- a/code/Model/VirtualPage.php
+++ b/code/Model/VirtualPage.php
@@ -3,7 +3,6 @@
 namespace SilverStripe\CMS\Model;
 
 use Page;
-use SilverStripe\Admin\AdminRootController;
 use SilverStripe\Core\Convert;
 use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\FieldList;
--- a/code/Model/VirtualPage.php
+++ b/code/Model/VirtualPage.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\CMS\Model;
 
 use Page;
+use SilverStripe\Admin\AdminRootController;
 use SilverStripe\Core\Convert;
 use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\FieldList;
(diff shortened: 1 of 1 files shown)


code/Model/VirtualPage.php
```diff
@@ -246,7 +247,7 @@ public function getCMSFields()
                     'a',
                     [
                         'class' => 'cmsEditlink',
-                        'href' => 'admin/pages/edit/show/' . $this->CopyContentFromID,
+                        'href' => AdminRootController::get_admin_route() . '/pages/edit/show/' . $this->CopyContentFromID,
```
ChloeSartorelli:
```suggestion
                        'href' => $this->CopyContentFrom()->CMSEditLink(),
```
Since this is hardcoding a link to the pages admin, we can assume there's no allowance for this to be anything other than `SiteTree` - so we may as well stop hardcoding any of the link at all.

code/Model/VirtualPage.php
```diff
@@ -3,6 +3,7 @@
 namespace SilverStripe\CMS\Model;
 
 use Page;
+use SilverStripe\Admin\AdminRootController;
```
ChloeSartorelli:
```suggestion
```
Not needed as per https://github.com/silverstripe/silverstripe-cms/pull/2882#discussion_r1317833538

GuySartorelli (changes requested):
Thanks for submitting this. There's just a small change to make below.

Please also target this at the `5.0` branch (or, if it affects CMS 4.13, target the `4.13` branch) so we can release it as a patch.

You may need to reset the commits on this branch after retargetting it so that the pull request only contains your one commit. I'm happy to guide you through that if you need it.

mirche97:
@GuySartorelli , I ran into the issue in 4.13, but I checked 5.0 and 5.1, it should be fixed there too. Do I only need to rebase and retarget it from/to 4.13 or all 3 branches?

GuySartorelli:
@mirche97 Please only retarget the pull request to 4.13 and then reset the commits. After your pull request is merged in there, it will be merged _up_ for the CMS 5 branches automatically.

You only ever need to raise multiple pull requests if the solution for fixing a bug differs between branches, which seems unlikely in this case.
If you need help resetting commits after retargetting it, I've [created this gist](https://gist.github.com/GuySartorelli/74627843aeb75e12b777b1f44cb22d2f) with instructions.

mirche97:
Hi @GuySartorelli ,
I've created a new PR https://github.com/silverstripe/silverstripe-cms/pull/2888 so I'm closing this one.

Issue #2885, opened 2023-09-11
CMS 5.1.0-beta1. Regression testing: Broken UI for "URL segment" buttons "OK" and "Cancel"


sabina-talipova:
## Description
Click Edit for Change URL segment results with graphics by OK and Cancel buttons (see attachment)
-https://studio.cucumber.io/projects/301855/test-runs/835224/folder-snapshots/8880145/scenario-snapshots/30498378/test-snapshots/41412650

![screencapture-cmstestm-prod-sites-silverstripe-admin-pages-edit-show-1542-2023-09-11-15_41_15](https://github.com/silverstripe/silverstripe-cms/assets/87288324/9823cd11-e7dc-479b-8f35-b6f8ed8d7152)

## PR
- https://github.com/silverstripe/silverstripe-admin/pull/1581
- https://github.com/silverstripe/silverstripe-cms/pull/2886

GuySartorelli:
Please include the image in a way that it can be viewed on GitHub instead of a private zenhub image,

maxime-rainville:
All done

PR #2886, opened 2023-09-14
FIX Regression issue UI URLSegment buttons

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2885

### DIFF
--- a/templates/SilverStripe/CMS/Forms/SiteTreeURLSegmentField.ss
+++ b/templates/SilverStripe/CMS/Forms/SiteTreeURLSegmentField.ss
@@ -13,12 +13,12 @@
     <div class="input-group">
         <input $AttributesHTML />
         <div class="input-group-append">
-            <button role="button" data-icon="accept" type="button" class="btn btn-primary update">
+            <button role="button" type="button" class="btn btn-primary update">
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Not doing a proper review, just looking at this in passing - but does this mean we can't have those icons on any buttons anywhere anymore? That seems like a regression that goes deeper than this PR would resolve.

maxime-rainville (approved):
Looks good. Tested in chrome and Firefox

maxime-rainville:
> Not doing a proper review, just looking at this in passing - but does this mean we can't have those icons on any buttons anywhere anymore? That seems like a regression that goes deeper than this PR would resolve.

The icon still works fine on all the other buttons. Those split buttons were never meant to have icons and have some weird padding applied to them which was cause this strange behaviour.

Looking at the pattern lib, the you can still apply icons to them just fine.

https://silverstripe.github.io/silverstripe-pattern-lib/?path=/docs/admin-buttons-button--documentation

GuySartorelli:
The pattern library is for react components. The template (which is what this PR affects) is for the entwine implementation, I believe.

PR #2888, opened 2023-09-21
FIX Don't hardcode edit link for virtual page

mirche97:
VirtualPage  now uses $this->CopyContentFrom()->CMSEditLink() to generate edit links

Replaces https://github.com/silverstripe/silverstripe-cms/pull/2882

Fixes the following issue: https://github.com/silverstripe/silverstripe-cms/issues/2881


GuySartorelli (approved):
Perfect, works great. Thank you

Issue #1805, opened 2017-04-27
Add SiteTree::$hide_pagetype setting


sminnee:
## Acceptance criteria
* SiteTree has a `$hide_pagetype` config value which accepts SiteTree subclass class names in an array
* `$hide_ancestor` is deprecated
* The new configuration property gets defined directly on `SiteTree` via yaml config - defining it on subclasses has no effect
* classes in that array will not be available in the CMS when selecting page type, for any user.
* Unit tests pass
* Docs exist and are correct

## PRs
- https://github.com/silverstripe/silverstripe-cms/pull/2902
- https://github.com/silverstripe/developer-docs/pull/393

tractorcow:
This is essentially a tweak of hide_ancestor, so it should be a simple rename / refactor.

tractorcow:
```
/**
     * If you extend a class, and don't want to be able to select the old class
     * in the cms, set this to the old class name. Eg, if you extended Product
     * to make ImprovedProduct, then you would set $hide_ancestor to Product.
     *
     * @config
     * @var string
     */
    private static $hide_ancestor = null;
```

sminnee:
Yeah I think we can ditch hide_ancestor favour of this mechanism.

chillu:
Removing from milestone, it's a new API. The old (replaced) API can be deprecated instead of removed once we get around to that in the 4.x release line.

pxwee5:
As of SS4.1@stable, there's no depreciation notice on `$hide_ancestor` or any implementation on `$hide_pagetype`.

As much as I believe `$hide_pagetype` is more intuitive to use and have no dependencies on parent classes, please give enough depreciation notice/time before we completely drop the `$hide_ancestor` configuration.

tractorcow:
Don't worry, if we deprecate it, it will still work perfectly fine until 5.0 at least. Deprecation notices won't show on deprecated APIs unless you opt in by the way.

```
// Set to 5.0.0 to show APIs marked for removal at that version
Deprecation::notification_version('4.0.0');
```

sabina-talipova:
Solution was implemented. Issue is closed.

Issue #2207, opened 2018-07-07
ENHANCEMENT: make it easier to hide pages in the list of "canCreate"


sunnysideup:
**What is this about?**

When creating new pages, you often end up with a ton of pages that can't actually be created.  Sometimes that is because of a rule specific to position or situation (e.g. top level / parents / children / can_create in combo with content editor) but other times this is because the page is actually superfluous.  

There are two levels of "can not create"

A. greyed out (e.g. not allowed to be a child of x, not allowed as top level page, etc...)

B. hidden altogether -> as can currently be achieved by using `SiteTree.hide_ancestor` (and now `SiteTree.hide_pagetypes`)

I basically would like to make it easier to achieve (B) because there are lots of situation where you want (B) but you end up using (A) creating fluff for the content editor. 

**Example**

The page may be included with a module but you do not actually want to use it.   For example, the CWP module contains a ton of page types that you may want to exclude.  At the moment, there is no straightforwarded way to exclude them from the "add new" list.  While you can not create them, showing them as "uncreatable" pages will annoy content editors.  

**What it looks like?**

![image](https://user-images.githubusercontent.com/167154/42406141-fd3b614a-81f5-11e8-9cdc-fede1c284e9e.png)

Here is the key method: 

https://github.com/silverstripe/silverstripe-cms/blob/2d9432789b9e0f8dbcef30464d21dd123a4121b3/code/Model/SiteTree.php#L470-L506

**Solutions**

1. extension point in `SiteTree::page_type_classes`
1. extension point in `ClassInfo::getValidSubClasses()`
1. extension point in `SiteTree::getClassDropdown`
1. extension point in `CMSMain::PageTypes`
1. exclude "canCreate === false" pages
1. add new `private static $admin_can_create`  or something along those lines (similar to SiteTree::$can_create, but if set false even admin can not create it). 

I also looked at `SiteTree::$need_permission` in combination with the can method, but because the ADMIN can always do everything this is not so easy.

You could also look at hiding pages that can not be created (A). The basic question is: does it matter that the content editor is aware what pages he/she can not create "at the moment".  Here in lies the crux.  The difference between (A) and (B).  It is really easy to achieve (A) _- e.g. you can not create a blog entry under the home page_ or _you can not create another home page because you have already created one_, but it is hard to create (B) ... _you can't create a blog page because we have not themed this yet and you actually dont need it_.

kinglozzer:
[...]

PR #2902, opened 2023-11-02
API Add new SiteTree.hide_pagetypes configuration

GuySartorelli:
Allows clearly defining all pagetypes that should be hidden in a single place, instead of having to spread them out across different pagetypes with MyPage.hide_ancestor

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/1805

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -203,10 +203,19 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      * in the cms, set this to the old class name. Eg, if you extended Product
      * to make ImprovedProduct, then you would set $hide_ancestor to Product.
      *
+     * @deprecated 5.2.0 Use hide_pagetypes instead
+     *
      * @config
      * @var string
      */
     private static $hide_ancestor = null;
+
+    /**
+     * Any fully qualified class names added to this array will be hidden in the CMS
+     * when selecting page types, e.g. for creating a new page or changing the type
+     * of an existing page.
+     */
+    private static array $hide_pagetypes = [];
 
     /**
      * You can define the class of the controller that maps to your SiteTree object here if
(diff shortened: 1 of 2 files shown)


code/Model/SiteTree.php
```diff
@@ -551,7 +560,7 @@ public static function page_type_classes()
             unset($classes[$baseClassIndex]);
         }
 
-        $kill_ancestors = [];
+        $kill_ancestors = self::config()->get('hide_pagetypes', Config::UNINHERITED) ?? [];
```
michalkleiner, line 563: Should we add an extension point if we're touching this? We had a few projects where it would come handy to be able to do this dynamically, e.g. for subsites, different user groups etc.

ChloeSartorelli: I'd be happy for that to be implemented as a separate PR, but it's out of scope for this one. My intention here is to explicitly make the configuration property easier to work with since `hide_ancestor` is basically being abused at the moment for hiding arbitrary non-ancestor classes.

tests/php/Model/SiteTreeTest.php
```diff
@@ -1223,6 +1230,11 @@ public function testPageTypeClasses()
             $newClasses,
             'Setting hide_ancestor to a boolean (incorrect) value caused a page class to be hidden'
         );
+
+        // Testing what happens if a valid config value is set
+        Config::modify()->set(SiteTreeTest_ClassA::class, 'hide_ancestor', 'Page');
+        $classes = SiteTree::page_type_classes();
+        $this->assertNotContains('Page', $classes);
```
ChloeSartorelli, line 1237: Without this addition, we don't actually have any test coverage validating that the old config even works 😅

code/Model/SiteTree.php
ChloeSartorelli: In fact there's a whole separate card for improving the way that hidden pages is defined which includes scope for extensions: https://github.com/silverstripe/silverstripe-cms/issues/2207

sabina-talipova (approved):
LGTM.

PR #2914, opened 2024-01-11
ENH Add generic types

GuySartorelli:
Requires https://github.com/silverstripe/silverstripe-framework/pull/11108 and the associated PR in silverstripe/recipe-cms for these generics to take effect.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11066

### DIFF
--- a/code/Controllers/OldPageRedirector.php
+++ b/code/Controllers/OldPageRedirector.php
@@ -10,9 +10,11 @@ use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Core\Extension;
 
+/**
+ * @extends Extension<ContentController|ModelAsController>
+ */
 class OldPageRedirector extends Extension
 {
-
     /**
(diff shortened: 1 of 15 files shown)


code/Controllers/OldPageRedirector.php
```diff
@@ -10,9 +10,11 @@
 use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Core\Extension;
 
+/**
+ * @extends Extension<Controller>
```
emteknetnz:
```suggestion
 * @extends Extension<ContentController|ModelAsController>
```

emteknetnz (changes requested):

code/Controllers/OldPageRedirector.php
ChloeSartorelli: Those are both subclasses of `Controller` - and there's no reason this couldn't be applied to someone's custom controller.

But given that would be a pretty unlikely edge case, and for the sake of avoiding ping pong, I've made the change.

emteknetnz (approved):
Behat failures are existing

PR #2929, opened 2024-04-19
MNT Replace sass-lint with stylelint

sabina-talipova:
## Description
Stylelint has about 100 rules, which are turned on by default ([See](https://stylelint.io/user-guide/rules/)). 
This PR removed support of old `sass-lint` and implemented new `stylelint`. New file `.stylelintrc.yml` add some configuration for default rules and also turn off some rules that we doesn't support. Probably they required additional review. 
## Issues
- https://github.com/silverstripe/webpack-config/issues/69


GuySartorelli (approved):
LGTM

Issue #2426, opened 2019-05-08
Restore batch action is unusable - Remove it


robbieaverill:
### Overview

The restore batch action in the site tree is unusable and therefore not required. Let's remove this option.

### Steps to see that the option is unusable

**SS 4.4.x-dev** (vanilla)

* Create some pages
* Publish them all
* Go to batch actions and select some pages
* Choose "Unpublish & Archive" and click "Go"
* Choose "Restore" from the select

Expected: I should be able to select the pages I just archived and reinstate them again

Actual: they don't show up in the site tree

If this is expected behaviour, I'm not sure what this batch action is supposed to be for. If you can't restore archived pages with it because they don't show up in the site tree when they're archived, what is its purpose?

## Acceptance criteria
- "Restore" option is removed from bulk site tree option in CMS 5.
- User help is reviewed and updated to remove any reference to the "Restore" bulk site tree option.
- Any related API that are not used by anything else are marked as deprecated in CMS 5 and removed from CMS 6.
- Updated CMS 5.3 changelog to add a brief mention of this change.

## PRs
- https://github.com/silverstripe/silverstripe-cms/pull/2938 (CMS 5)
- https://github.com/silverstripe/silverstripe-userhelp-content/pull/181
- https://github.com/silverstripe/developer-docs/pull/506
- https://github.com/silverstripe/silverstripe-cms/pull/2940 (CMS 6)

brynwhyman:
Experiencing this in 4.5.alpha1 too. 

This functionality may have predated the Archives panel in the CMS. Perhaps it's something that can be removed in lieu of that functionality? _Although that doesn't yet have batch actions..._

cc @silverstripeux

clarkepaul:
I think its best to remove the Restore action for now. Batch actions in the Archive and within search results would be the way to bring the functionality back.

When search results of the tree moved to be presented as a list (rather than a tree) would have been the time the functionality was lost.

clarkepaul:
Related https://github.com/silverstripe/silverstripe-admin/issues/173

brynwhyman:
Thanks for the feedback. Given these thoughts, I'm updating this issue to focus on removing this (unusable) functionality.

maxime-rainville:
If the thing is not working at all and we have no intention of fixing it, it seems a bit weird to call this an API break.

GuySartorelli:
non-draft PRs merged.
Reassigned to Steve for draft PR.

[...]

PR #2938, opened 2024-05-09
FIX Remove broken restore batch action

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/2426


tests/php/Controllers/CMSBatchActionsTest.php
ChloeSartorelli, line 1: Don't remove the tests - we're just deprecating it and removing it from the list. Someone else might want to add it back to the list if they've got it working somehow, or they might be using the `CMSBatchAction_Restore` class itself to do some work, so that needs to not regress.
Tests will be removed when the class gets deleted.

emteknetnz: Done

GuySartorelli (approved):
LGTM

Issue #2933, opened 2024-05-05
SPIKE Analyse how LeftAndMain and CMSMain are put together and how they could work better


maxime-rainville:
CMSMain, LeftAndMain, ModelAdmin and other related Admin controllers have been around and have accumulated quite a bit of technical debt.

## Timebox
3 days

## Objectives of this SPIKE
- Review the architecture of LeftAndMain, CMSMain, ModelAdmin and any other related class.
- Identify key methods and their purpose and structure them into inter-related groups.
- Propose a more sensible architecture and break down. This may include:
  - moving method between classes
  - creating brand new classes
  - splitting concerns into composable services or traits
- Identify other concerns or discrepancies that could be address in a future major release
- Report findings at a CMS Architecture catch up

## Related cards
- https://github.com/silverstripe/silverstripe-cms/issues/2896
- https://github.com/silverstripe/silverstripe-admin/issues/348

## POC PRs
For demonstrative purposes only
- https://github.com/silverstripe/silverstripe-cms/pull/2939

GuySartorelli:
# `LeftAndMain`

## tl;dr

`LeftAndMain` could be broken into the following:

- abstract `AdminRouteController`
    - Possibly a new class, or possibly dump stuff into the existing `AdminRootController`
    - Handles permissions and routing against `/admin/*`
    - Any controller which should have a `/admin/*` route (i.e. it's intended to be used in the CMS, not on the front-end) should be a subclass of this.
- abstract `FormSchemaController`
    - Either a subclass of `AdminRouteController` or a trait
    - Has all the stuff we need to handle formschema modal stuff.
    - The `LinkFieldController` class would either use this trait or be a subclass of this class.
- abstract `LeftAndMain`
    - Implements whatever is needed for classes to be in the "left" menu, and have a "main" edit form

## Recommended changes in depth

Move all properties to the top of the class. This can be done in CMS 5. For some reason there's properties that are below some of the methods, which makes them harder to find.

`LeftAndMain` should be made `abstract` - the few cases that currently require a singleton of `LeftAndMain` should be refactored to not require that. From what I could see, those are mostly related to formschema currently.

### Break into multiple classes

#### abstract `AdminRouteController`

- Possibly a new class, or possibly dump stuff into the existing `AdminRootController`
- Handles permissions and routing against `/admin/*`
- No coupling with templates
[...]

PR #2939, opened 2024-05-09
POC DO NOT MERGE - Make CMSMain work with other models

GuySartorelli:
This is a proof of concept as part of https://github.com/silverstripe/silverstripe-cms/issues/2933 to see what's needed to make `CMSMain` work with classes that aren't `SiteTree`.

To play around with this, install [silverstripe/taxonomy](https://github.com/silverstripe/silverstripe-taxonomy/) and add the following YAML config to your project:

```yml
SilverStripe\Taxonomy\TaxonomyTerm:
  cms_edit_owner: SilverStripe\CMS\Controllers\CMSMain
  extensions:
    - SilverStripe\Admin\CMSEditLinkExtension

SilverStripe\CMS\Controllers\CMSMain:
  tree_class: SilverStripe\Taxonomy\TaxonomyTerm
```

Your `/admin/pages` will now be working with taxonomy terms, rather than pages.

Theoretically this could work with any `DataObject` class which has `CMSEditLinkExtension` and `Hierarchy` applied.

## What would still need to be done besides this

1. Refactor `CMSPagesController`, `CMSPageEditController`, `CMSPageSettingsController`, `CMSPageHistoryViewerController`, and `CMSPageAddController` so you could have multiple instances of this going at once, e.g. one admin for pages, one for taxonomies, etc - without having to subclass all of these controllers as well.
    - Ideally these wouldn't be subclasses of `LeftAndMain` at all - either this could all be handled in the main `HierarchyModelAdmin` class (see below) or these could all be more generic controllers.
    - Either way they need to not be subclasses of `CMSMain`, but instead get any information they need by calling methods on `CMSMain` (i.e. have the `CMSMain` instance passed to their constructor) or having that information passed through to them via setters or via their constructor.
1. Have this logic all in a superclass called something more generic, like `HierarchyModelAdmin` (which would belong in `silverstripe/admin` along with all appropriate JS), and have `CMSMain` be a subclass of `HierarchyModelAdmin` with its `tree_class` set to `SiteTree`
    - While moving the JS to silverstripe/admin, it'd be worth finding all JS logic that's duplicated and making it so that all works in the same way for `ModelAdmin` and this new class.
1. Obviously we'd need to add in appropriate type checking, make sure the tree class is hierarchical and has a cms edit link.
1. In a few places it'd pay to add early returns when versioned isn't applied to the record
1. This POC does a lot of lazy work to avoid refactoring - e.g. there's a lot of methods now duplicated between `SiteTree` and `CMSMain`. These should go into a new class (probably an extension) that can be applied to any classes that want to live in this admin type. Things like tree title vs menu title vs title, icon, status flags, tree node classes, etc
    - This will end up meaning if you want to add a model to this admin, it must have `Hierarchy`, `CMSPageEditController`, and this new extension. But that also means you can just apply 3 extensions to any model and boom, you're ready to go.
    - Some things like the breadcrumbs for list view shouldn't be on the model - just leave that in the controller or fully template-ify it
    - Some things like status flags should probably have default functionality on the controller but use `invokeWithExtensions` (or just a `hasMethod()` check) to allow the model to update it
1. There's a lot of wording to adjust that assumes this is for pages.
1. Batch actions assume `Versioned` is applied and probably also have other assumptions that make them not work
1. The way search works for `CMSMain` is just completely incompatible with these changes. It has a hardcoded form, and a different code path for the actual search functionality. We should marry up the way `GridFieldFilterHeader` and the filter/search in here work.
    - Probably have a new controller for CMS search. This gets the form using similar logic to the current `GridFieldFilterHeader` logic
    - Current fields in `CMSMain::getSearchForm()` get moved into `SiteTree::searchableFields()`
    - `SiteTree` gets a custom `SearchContext` subclass which leverages `CMSSiteTreeFilter` for the field that uses it (if that's necessary - I haven't looked too deep into the search code path for sitetree)
1. I haven't really looked at all at the history tab - it might need some work. MVP would be to just not show that for anything other than `SiteTree` and handle making it work as a separate effort. 
1. A whole bunch of new unit and behat tests would need to be written
1. The root of the site tree is the site name, which makes sense for pages but doesn't really make sense for much of anything else. Might be worth just removing that root, or making it configurable, or something.

## Stretch goal

Make this handle multiple classes in separate tabs, like `ModelAdmin` does via its `managed_models` config. Then you could still have a single "Taxonomies" CMS section, but it could handle both `TaxonomyTerm` and `TaxonomyType` with this hierarchical style.

Note that `TaxonomyType` isn't hierarchical - so either we'd need to change that, or better yet AAALLL of this logic could just get bundled together with `ModelAdmin`, so you can have for example one tab for managing `TaxonomyTerm` which uses the tree structure, and one tab for managing `TaxonomyType` which uses the normal `ModelAdmin` grid structure.

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -25,6 +25,7 @@ use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Control\PjaxResponseNegotiator;
 use SilverStripe\Core\Cache\MemberCacheFlusher;
+use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Environment;
@@ -73,6 +74,7 @@ use SilverStripe\Versioned\Versioned;
 use SilverStripe\VersionedAdmin\Controllers\CMSPageHistoryViewerController;
 use SilverStripe\View\ArrayData;
 use SilverStripe\View\Requirements;
+use SilverStripe\View\SSViewer;
 
 /**
  * The main "content" area of the CMS.
@@ -178,6 +180,11 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
      */
     protected $hintsCache;
 
+    public function getTreeClass(): string
+    {
+        return static::config()->get('tree_class');
+    }
+
     protected function init()
     {
         parent::init();
@@ -252,7 +259,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
      * Overloads the LeftAndMain::ShowView. Allows to pass a page as a parameter, so we are able
      * to switch view also for archived versions.
      *
-     * @param SiteTree $page
+     * @param DataObject $page
      * @return array
      */
     public function SwitchView($page = null)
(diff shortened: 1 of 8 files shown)


GuySartorelli:
Closing - no need to keep it open, the code will still be there even when it's closed.

PR #2955, opened 2024-06-05
ENH Use class name instead of self

emteknetnz:
Issue https://github.com/silverstripe/developer-docs/issues/486

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -128,11 +128,11 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      * @var array
      */
     private static $allowed_children = [
-        self::class
+        SiteTree::class
     ];
 
     /**
-     * Used as a cache for `self::allowedChildren()`
+     * Used as a cache for `SiteTree::allowedChildren()`
      * Drastically reduces admin page load when there are a lot of page types
      * @var array
      */
@@ -457,7 +457,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
     public static function get_by_link($link, $cache = true)
     {
         // Compute the column names with dynamic a dynamic table name
-        $tableName = DataObject::singleton(self::class)->baseTable();
+        $tableName = DataObject::singleton(SiteTree::class)->baseTable();
         $urlSegmentExpr = sprintf('"%s"."URLSegment"', $tableName);
         $parentIDExpr = sprintf('"%s"."ParentID"', $tableName);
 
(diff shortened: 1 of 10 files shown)


code/Model/SiteTree.php
```diff
@@ -471,16 +471,16 @@ public static function get_by_link($link, $cache = true)
         // Grab the initial root level page to traverse down from.
         $URLSegment = array_shift($parts);
         $conditions = [$urlSegmentExpr => rawurlencode($URLSegment ?? '')];
-        if (self::config()->get('nested_urls')) {
+        if (static::config()->get('nested_urls')) {
```
ChloeSartorelli, line 474: This (and any other invocations of this particular config property) really should be `SiteTree::config()` - this config property seems very much intended to have a single value that affects the system as a whole, rather than resulting in different behaviours for different subclasses based on config for each subclass.

That said, it's possible that this could be (ab)used right now to suit project-specific use cases - so changing this to `SiteTree` now might cause bugs.

@emteknetnz Do you have any thoughts on this?

code/Model/SiteTree.php
```diff
@@ -552,12 +552,12 @@ public static function page_type_classes()
...
         $classes = ClassInfo::getValidSubClasses();
 
-        $baseClassIndex = array_search(self::class, $classes ?? []);
+        $baseClassIndex = array_search(SiteTree::class, $classes ?? []);
         if ($baseClassIndex !== false) {
             unset($classes[$baseClassIndex]);
         }
 
-        $kill_ancestors = self::config()->get('hide_pagetypes', Config::UNINHERITED) ?? [];
+        $kill_ancestors = static::config()->get('hide_pagetypes', Config::UNINHERITED) ?? [];
```
```suggestion
        $kill_ancestors = SiteConfig::config()->get('hide_pagetypes', Config::UNINHERITED) ?? [];
```
This is a [new config as of 5.2](https://github.com/silverstripe/silverstripe-cms/pull/2902) and was very much intended to be just a single config value across the whole class hierarchy.

This change should be made regardless of the outcome of https://github.com/silverstripe/silverstripe-cms/pull/2955#discussion_r1642018370 - the risk of the change is way smaller, since it's both very new and has a much narrower affect on code paths.

GuySartorelli (changes requested):

code/Model/SiteTree.php
emteknetnz: I've been careful to change self::config() to static::config() (they behave identically) and to leave MyClass::config() as it current is. The simple reason is that this is refactoring and not a bugfix/feature, so it should behave exactly the same as before.

If you want to change this, it should be done separate from this set of PRs

code/Model/SiteTree.php
emteknetnz: Updated

GuySartorelli (approved):
LGTM

PR #2956, opened 2024-06-05
ENH Use class name instead of self

emteknetnz:
Issue https://github.com/silverstripe/developer-docs/issues/486


GuySartorelli:
@emteknetnz i assume this is a dupe of https://github.com/silverstripe/silverstripe-cms/pull/2955?
May pay to check if the other repos have duplicates too.

emteknetnz:
Yup dupe. Had a look through rhino couldn't find any other dupes

PR #2970, opened 2024-06-19
ENH Auto-scaffold SiteTree relations with tree fields

GuySartorelli:
These form fields do have lazy-loading applied after a configurable number of records exist, so they should be sensibly performant even for large sites. They also give the most appropriate UI for selecting pages to save into a relation, IMO.

## Issue
- https://github.com/silverstripe/.github/issues/262


emteknetnz (approved):
Tested locally, works well

PR #2971, opened 2024-06-19
MNT Tweak behat test to look for tag with specific attribute itstead of looking for a specifc string in the HTML

maxime-rainville:
This test was failing because it was looking for a very specific string in the HTML. Just checking for a tag with a specific attribute is more reliable because it doesn't depend on the order of parameter.

## Parent issue
- https://github.com/silverstripe/.github/issues/230

### DIFF
--- a/tests/behat/features/edit-a-page.feature
+++ b/tests/behat/features/edit-a-page.feature
@@ -103,11 +103,11 @@ Feature: Edit a page
     And I press the "Publish" button
     And I go to "/about-us"
     # insert from files
-    Then the rendered HTML should contain "<img src=\"/assets/file1.jpg\""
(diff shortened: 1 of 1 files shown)


tests/behat/features/edit-a-page.feature
```diff
@@ -77,6 +77,7 @@ Feature: Edit a page
     When I go to "/admin/pages"
     And I click on "About Us" in the tree
 
+  @onlyme
```
ChloeSartorelli:
```suggestion
```

emteknetnz (changes requested):
Need to remove the `@onlyme` annotation

maxime-rainville:
Fixed my booboo.

emteknetnz (approved):

Issue #2974, opened 2024-07-10
Duplicate breadcrumb labels when navigating site tree in ListView


maxime-rainville:
### Module version(s) affected

4.13.13 and 5.2.2

### Description

When navigating the sitetree in list view, the "crumbs" in the "breadcumbs" get doubled up.

### How to reproduce

1. Access the CMS
2. Create a top level page with the title of "First Level".
3. Create a child page under "First Level" with the title of "Second Level".
4. Create a child page under "Second Level" with the title of "Third Level".
5. Create a child page under "Third Level" with the title of "Fourth Level".
6. Switch to List View.
7. Navigate to Third Level.

Expected: The breadcrumb looks like `First Level / Second Level / Third Level`
Actual: The breadcrumb looks like `First Level / Second Level / Third Level / First Level / Second Level / Third Level`

![Screenshot 2024-07-10 at 4 38 21 PM](https://github.com/silverstripe/silverstripe-cms/assets/1168676/b2e6726e-691d-4582-b277-1fe41ea71e4f)

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## Pull request
- https://github.com/silverstripe/silverstripe-cms/pull/2976

maxime-rainville:
I haven't look at the underlying cause. I'll see if there's a quick fix there tonight.

maxime-rainville:
Don't have fix yet, but I think I narrowed the problem down a bit.

The problem is only apparent on PJAX request. After step 7, if you refresh, your breadcrumb will look like "Page / First Level / Second Level / Third Level". That's probably the correct end state.

Part of the problem is in the interrelation between [`CMSMain::Breadcrumbs`](https://github.com/silverstripe/silverstripe-cms/blob/7e10f298177b08b84723811467aa20baf04ad9e7/code/Controllers/CMSMain.php#L1039-L1089) and [`CMSPageController::Breadcrumbs`](https://github.com/silverstripe/silverstripe-cms/blob/7e10f298177b08b84723811467aa20baf04ad9e7/code/Controllers/CMSPagesController.php#L35)

When doing a regular request, CMSMain will got through this `if` statement. `updateBreadcrumbs` will be called with an array containing only the a "Pages" link.
[...]

PR #2976, opened 2024-07-21
BUG Don't double up the breadcrumb on list views

maxime-rainville:
## Description
This PR refactors how the CMS breadcrumb is generated.

It fix a bug causing crumbs to be duplicated on the list view.

It also ensures the "Pages" top crumb is always shown on the list view.

It adds missing unit tests.

## Manual testing steps
- Validate that the breadcrumb is still shown correctly:
  - on the page edit form
  - on the list view when access directly
  - on the list view when access via a PJAX
  - on the page search results.
- Validate that crumbs link where you expect them to link.

## Issues
- https://github.com/silverstripe/silverstripe-cms/issues/2974

## Pull request checklist

- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] CI is green

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -1040,54 +1040,93 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
 
     public function Breadcrumbs($unlinked = false)
     {
-        $items = new ArrayList();
+        $items = ArrayList::create();
 
-        if ($this->TreeIsFiltered()) {
-            $items->push(new ArrayData([
-                'Title' => CMSPagesController::menu_title(),
-                'Link' => ($unlinked) ? false : $this->LinkPages()
-            ]));
-            $items->push(new ArrayData([
-                'Title' => _t('SilverStripe\\CMS\\Controllers\\CMSMain.SEARCHRESULTS', 'Search results'),
-                'Link' => ($unlinked) ? false : $this->LinkPages()
-            ]));
+        if (($this->getAction() !== 'index') && ($record = $this->currentPage())) {
+            // The page is being edited
+            $this->buildEditFormBreadcrumb($items, $record, $unlinked);
+        } else {
+            // Ensure we always have the "Pages" crumb first
+            $this->pushCrumb(
+                $items,
+                CMSPagesController::menu_title(),
+                $unlinked ? false : $this->LinkPages()
+            );
 
-            $this->extend('updateBreadcrumbs', $items);
-
-            return $items;
-        }
-
-        // Check if we are editing a page
-        /** @var SiteTree $record */
-        $record = $this->currentPage();
-        if (!$record) {
-            $items->push(new ArrayData([
-                'Title' => CMSPagesController::menu_title(),
-                'Link' => ($unlinked) ? false : $this->LinkPages()
-            ]));
-
-            $this->extend('updateBreadcrumbs', $items);
-
-            return $items;
-        }
-
-        // Add all ancestors
-        $ancestors = $record->getAncestors();
-        $ancestors = new ArrayList(array_reverse($ancestors->toArray() ?? []));
-        $ancestors->push($record);
-        /** @var SiteTree $ancestor */
-        foreach ($ancestors as $ancestor) {
-            $items->push(new ArrayData([
-                'Title' => $ancestor->getMenuTitle(),
-                'Link' => ($unlinked)
-                    ? false
-                    : $ancestor->CMSEditLink()
-            ]));
+            if ($this->TreeIsFiltered()) {
+                // Showing search results
+                $this->pushCrumb(
+                    $items,
+                    _t(CMSMain::class . '.SEARCHRESULTS', 'Search results'),
+                    ($unlinked) ? false : $this->LinkPages()
+                );
+            } elseif ($parentID = $this->getRequest()->getVar('ParentID')) {
+                // We're navigating the listview. ParentID is the page whose
+                // children are currently displayed.
+                if ($page = SiteTree::get()->byID($parentID)) {
+                    $this->buildListViewBreadcrumb($items, $page);
+                }
+            }
         }
 
         $this->extend('updateBreadcrumbs', $items);
 
         return $items;
+    }
+
+    /**
+     * Push the provided an extra breadcrumb crumb at the end of the provided List
+     */
+    private function pushCrumb(ArrayList $items, string $title, string|false $link): void
+    {
+        $items->push(ArrayData::create([
+            'Title' => $title,
+            'Link' => $link
+        ]));
+    }
+
+    /**
+     * Build Breadcrumb for the Edit page form. Each crumb links back to its own edit form.
+     */
+    private function buildEditFormBreadcrumb(ArrayList $items, SiteTree $page, bool $unlinked): void
+    {
+        // Find all ancestors of the provided page
+        $ancestors = $page->getAncestors(true);
+        $ancestors = array_reverse($ancestors->toArray() ?? []);
+        foreach ($ancestors as $ancestor) {
+            // Link to the ancestor's edit form
+            $this->pushCrumb(
+                $items,
+                $ancestor->getMenuTitle(),
+                $unlinked ? false : $ancestor->CMSEditLink()
+            );
+        }
+    }
+
+    /**
+     * Build Breadcrumb for the List view. Each crumb links to the list view for that page.
+     */
+    private function buildListViewBreadcrumb(ArrayList $items, SiteTree $page): void
+    {
+        // Find all ancestors of the provided page
+        $ancestors = $page->getAncestors(true);
+        $ancestors = array_reverse($ancestors->toArray() ?? []);
+        
+        //turns the title and link of the breadcrumbs into template-friendly variables
+        $params = array_filter([
+            'view' => $this->getRequest()->getVar('view'),
+            'q' => $this->getRequest()->getVar('q')
+        ]);
+        
+        foreach ($ancestors as $ancestor) {
+            // Link back to the list view for the current ancestor
+            $params['ParentID'] = $ancestor->ID;
+            $this->pushCrumb(
+                $items,
+                $ancestor->getMenuTitle(),
+                Controller::join_links($this->Link(), '?' . http_build_query($params ?? []))
+            );
+        }
     }
 
     /**
--- a/code/Controllers/CMSPagesController.php
+++ b/code/Controllers/CMSPagesController.php
@@ -31,35 +31,4 @@ class CMSPagesController extends CMSMain
     {
         return false;
     }
-
-    public function Breadcrumbs($unlinked = false)
-    {
-        $this->beforeExtending('updateBreadcrumbs', function (ArrayList $items) {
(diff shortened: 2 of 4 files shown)


code/Controllers/CMSMain.php
```diff
@@ -1038,56 +1038,93 @@ public function getBreadcrumbsBackLink()
...
-                'Title' => CMSPagesController::menu_title(),
-                'Link' => ($unlinked) ? false : $this->LinkPages()
-            ]));
-            $items->push(new ArrayData([
-                'Title' => _t('SilverStripe\\CMS\\Controllers\\CMSMain.SEARCHRESULTS', 'Search results'),
-                'Link' => ($unlinked) ? false : $this->LinkPages()
-            ]));
-
-            $this->extend('updateBreadcrumbs', $items);
+        if (($this->getAction() !== 'index') && ($record = $this->currentPage())) {
```
maxime-rainville, line 1045: Checking the action here is necessary because we end up with a record in some of the listview context

code/Controllers/CMSMain.php
```diff
@@ -1038,56 +1038,93 @@ public function getBreadcrumbsBackLink()
...
 
-        // Check if we are editing a page
-        /** @var SiteTree $record */
-        $record = $this->currentPage();
-        if (!$record) {
-            $items->push(new ArrayData([
-                'Title' => CMSPagesController::menu_title(),
-                'Link' => ($unlinked) ? false : $this->LinkPages()
-            ]));
+        $this->extend('updateBreadcrumbs', $items);
```
maxime-rainville, line 1072: The original implementation would build the breadcrumb for a specific scenario, call `updateBreadcrumbs` and return.

Seems more logical to build to use `if` statements to make sure only one breadcrumb is built and then call `updateBreadcrumbs` once.

code/Controllers/CMSPagesController.php
```diff
@@ -31,35 +31,4 @@ public function isCurrentPage(DataObject $record)
     {
         return false;
     }
-
-    public function Breadcrumbs($unlinked = false)
```
maxime-rainville, line 35: This bit of logic was the reason why we were ending up with duplicate crumbs. Basically, the `CMSMain` would want to build the crumb once and this would build it again. 

I really don't see why this was put on a separate controller. Having all the Breadcrumb logic on CMSMain simplifies everything.

code/Controllers/CMSPagesController.php
```diff
@@ -31,35 +31,4 @@ public function isCurrentPage(DataObject $record)
...
-                $pages = [];
-                while ($page) {
-                    array_unshift($pages, $page); //add to start of array so that array is in reverse order
-                    $page = $page->Parent;
-                }
-
-                //turns the title and link of the breadcrumbs into template-friendly variables
-                $params = array_filter([
-                    'view' => $this->getRequest()->getVar('view'),
-                    'q' => $this->getRequest()->getVar('q')
```
maxime-rainville, line 52: These parameters are useless ... I couldn't find anyway to get the CMS to output them for me. Even if I manually put them in, nothing happens with them. My guess is that they are a left over from a time someone was hoping to allow the search to work in conjunction with a tree view or something.

I just took them out from my final implementation.

code/Controllers/CMSMain.php
```diff
@@ -1038,56 +1038,93 @@ public function getBreadcrumbsBackLink()
             ->Link;
     }
 
+    /**
+     * Generate a breadcrumb based on the page currently being viewed.
+     * @param bool $unlinked Wheter to include links
+     * @return ArrayList<ArrayData>
+     */
```
ChloeSartorelli: The param and return types are inherited from `LeftAndMain` so no need to add those.
The phpdoc comment is in the wrong place - should be added to `LeftAndMain`.
```suggestion
```

code/Controllers/CMSPagesController.php
ChloeSartorelli: Please reinstate them - those aren't related to the problem you're fixing and since we don't really know what they're doing, removing them now could introduce unexpected regressions. We can remove them in a future refactor after looking into it further.

code/Controllers/CMSMain.php
```diff
@@ -1038,56 +1038,93 @@ public function getBreadcrumbsBackLink()
...
-
-        return $items;
+    /**
+     * Build Breadcrumb for the List view. Each crumb links to the list view for that page.
+     */
+    private function buildListViewBreadcrumb(ArrayList $items, SiteTree $page): void
+    {
+        // Find all ancestors of the provided page
+        $ancestors = $page->getAncestors(true);
+        $ancestors = array_reverse($ancestors->toArray() ?? []);
```
ChloeSartorelli, line 1113: This is very slightly less code than the original, but you're looping over everything twice now instead of just once! No array_unshift though so it maaaaybe balances out? Probably negligible anyway so I'm okay with it.

code/Controllers/CMSMain.php
```diff
@@ -1038,56 +1038,93 @@ public function getBreadcrumbsBackLink()
...
+        foreach ($ancestors as $ancestor) {
+            // Link back to the list view for the current ancestor
+            $params = ['ParentID' => $ancestor->ID];
+            $this->pushCrumb(
+                $items,
+                $ancestor->getMenuTitle(),
```
ChloeSartorelli, line 1126: Original implementation uses `Title` explicitly. Why are you using `getMenuTitle()` here?

code/Controllers/CMSMain.php
```diff
@@ -1038,56 +1038,93 @@ public function getBreadcrumbsBackLink()
...
+    /**
+     * Push the provided an extra breadcrumb crumb at the end of the provided List
+     */
+    private function pushCrumb(ArrayList $items, string $title, string|false $link): void
+    {
+        $items->push(ArrayData::create([
+            'Title' => $title,
+            'Link' => $link
+        ]));
+    }
```
ChloeSartorelli, line 1086: This extra method seems kinda unnecessary but not worth the ping pong to remove it, since it's private anyway.

code/Controllers/CMSMain.php
```diff
@@ -1038,56 +1038,93 @@ public function getBreadcrumbsBackLink()
...
-                'Link' => ($unlinked) ? false : $this->LinkPages()
-            ]));
+        $items = ArrayList::create();
 
-            $this->extend('updateBreadcrumbs', $items);
+        if (($this->getAction() !== 'index') && ($record = $this->currentPage())) {
+            // The page is being edited
+            $this->buildEditFormBreadcrumb($items, $record, $unlinked);
+        } else {
+            // All other scenarios include a "Pages" first crumb
```
ChloeSartorelli:
```suggestion
            // Ensure we always have the "Pages" crumb first
```

code/Controllers/CMSMain.php
ChloeSartorelli: I'm not sure what you mean, can you please elaborate?

There used to be a condition like this:
```php
$record = $this->currentPage();
if (!$record) {
```

Is this new condition performing that same job? If so why are we doing it differently in this PR than we were before?

GuySartorelli (changes requested):
While I appreciate the initiative with all of the refactoring, it made it a lot more time consuming to review since I needed to compare the original code with the new code and parse out the changes and see what moved where, etc etc.

In the future can you please try to keep refactoring separate from fixing the bug? At the very least having separate commit (one dedicated to just what's necessary to fix the bug, and a separate one for the refactoring) would make it a lot easier to keep track of.

GuySartorelli:
Please also change the commit prefix to `FIX` as per the commit message guidelines linked in the PR description.

code/Controllers/CMSMain.php
maxime-rainville: Yes, this scenario is meant to cover what was covered by
```php
$record = $this->currentPage(); 
if (!$record) {
```

## Scenarios covered by this if statement
`$this->currentPage()` will return a truthy value in the following condition:
- when accessing the page for editing (/admin/pages/edit/show/1)
- when accessing a page's setting (/admin/pages/settings/show/1)
- when accessing a page's history (/admin/pages/history/show/1)

In all tho situation we want the breadcrumb to look like this: 
image

- The breadcrumb on the right doesn't have a "Pages" root crumb, because that's covered by the "Pages" link at the top of the site tree.
- The crumbs link to the edit form for their page.

## Scenario not covered by this IF
When viewing the listview and you drill down to see the children of the page, that triggers a PJAX request to `/admin/pages?ParentID=123` where `123` is the ID of the page for which you want to view the children.

For that request, `$this->currentPage()` will return an instance of Page ID 123. However, the breadcrumb needs to be different for that use case. Hence why we are adding a `$this->getAction() !== 'index'` check.

That was effectively the root of the original bug:
- First the logic in CMSPagesController::Breadcrumb would be executed, creating a initial breadcrumb.
- Then the logic in this `if` statement would be executed again. Doubling up the breadcrumb.

code/Controllers/CMSMain.php
maxime-rainville: I honestly didn't notice that MenuTitle difference until reading your comment.

Basically, the original implementation had two different approaches to building the breadcrumb ancestry:
- `CMSPagesController::Breadcrumbs` manually loops through a Page's parent and use the page Title for the crumb.
- `CMSMain::Breadcrumbs` uses `SiteTree::getAncestors` and uses `MenuTitle` for the crumb.

I thought having two different approaches to do the same thing was stupid. The `CMSMain` approaches seemed better since it was using a dedicated method, so I copied that one without noticing the different title.

The ListTree view entries, the SiteTree view entries and all the other breadcrumbs use MenuTitle. I think MenuTitle is more appropriate here.

code/Controllers/CMSPagesController.php
maxime-rainville: Fair enough. I'll reinstate that bit of logic.

code/Controllers/CMSMain.php
maxime-rainville: Done

code/Controllers/CMSMain.php
maxime-rainville: Done

maxime-rainville (commented):
> While I appreciate the initiative with all of the refactoring, it made it a lot more time consuming to review since I needed to compare the original code with the new code and parse out the changes and see what moved where, etc etc.
> 
> In the future can you please try to keep refactoring separate from fixing the bug? At the very least having separate commit (one dedicated to just what's necessary to fix the bug, and a separate one for the refactoring) would make it a lot easier to keep track of.

Sorry for the unduly big change. But the more I looked at this thing, the more it seemed like the bugs were fundamentally caused by the bad implementation.

Refactoring the code while replicating the bugs would have probably made this PR even more complex. And I don't think there would have been a way to fix the bugs without addressing the underlying poor implementation.

> Please also change the commit prefix to FIX as per the commit message guidelines linked in the PR description.

Done.

GuySartorelli (approved):
LGTM, works well locally.

I found a new related bug that when I'm in list view and drill down to a specific page, if I then toggle to "tree view" it'll retain the `?ParentID=9` in the URL and keeps the breadcrumbs there. But I tested that without this PR and it was still happening so it's an existing bug.

Issue #2767, opened 2022-08-24
Scaffold fields for SiteTree


maxime-rainville:
When you define DB fields for a regular DataObject, `getCMSFields` will scaffold them for you.

SiteTree doesn't do that. So you have to manually created all the fields. We should remedy this.

Maybe we provide an option to disable that to minimise upgrade pain.

## Note
- We made some DataObjects auto-scaffoldable with more sensible form fields in https://github.com/silverstripe/.github/issues/262. Some of that might need to be adjusted or taken into account when actioning this card.

## Acceptance Criteria
- `SiteTree::getCMSFields()` calls `parent::getCMSFields()` (which will autoscaffold fields)
- Custom logic in `SiteTree::getCMSFields()` is wrapped in `$this->beforeUpdateCMSFields()` so extension hooks are called appropriately
- The changelog makes it very clear that this change is happening, and that developers should check their implementations of `getCMSFields()` in their page classes to ensure things are working the way they expect (e.g. autscaffolded `has_many` and `many_many` relations may be giving them new tabs that weren't there before)
- All native Page type are revisited to validate that:
  - the FormFields are still displayed in the same tab as they were before
  - to remove redundant scaffolding logic
  - no new form fields are added to existing pages.

## CMS 5 PRs

- https://github.com/silverstripe/silverstripe-framework/pull/11328
- https://github.com/silverstripe/silverstripe-framework/pull/11329
- https://github.com/silverstripe/developer-docs/pull/549

## CMS 6 PRs

**[kitchen sink CI run](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/10360831559)**
recipe-reporting failure is expected because we're forking framework for that CI run.
Compare with [normal CMS 6 run](https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/10345378787) for other failures.

- https://github.com/silverstripe/silverstripe-framework/pull/11327
- https://github.com/silverstripe/silverstripe-cms/pull/2983
- https://github.com/silverstripe/silverstripe-userforms/pull/1315
- https://github.com/silverstripe/silverstripe-blog/pull/775
- https://github.com/silverstripe/silverstripe-subsites/pull/589
- https://github.com/silverstripe/silverstripe-registry/pull/103
- https://github.com/silverstripe/silverstripe-iframe/pull/95
- https://github.com/silverstripe/silverstripe-sharedraftcontent/pull/248
[...]

PR #2983, opened 2024-08-08
NEW Use autoscaffolding for SiteTree CMS fields

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2767

### DIFF
--- a/code/Model/RedirectorPage.php
+++ b/code/Model/RedirectorPage.php
@@ -3,13 +3,10 @@
 namespace SilverStripe\CMS\Model;
 
 use Page;
-use SilverStripe\AssetAdmin\Forms\UploadField;
 use SilverStripe\Assets\File;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\HeaderField;
 use SilverStripe\Forms\OptionsetField;
-use SilverStripe\Forms\TextField;
-use SilverStripe\Forms\TreeDropdownField;
 use SilverStripe\Versioned\Versioned;
 
 /**
@@ -43,6 +40,13 @@ class RedirectorPage extends Page
     private static $has_one = [
         "LinkTo" => SiteTree::class,
         "LinkToFile" => File::class,
+    ];
+
+    private static array $scaffold_cms_fields_settings = [
+        'ignoreFields' => [
+            'RedirectionType',
+            'Content',
+        ],
     ];
 
     private static $table_name = 'RedirectorPage';
@@ -194,18 +198,19 @@ class RedirectorPage extends Page
     public function getCMSFields()
     {
         $this->beforeUpdateCMSFields(function (FieldList $fields) {
-            $fields->removeByName('Content', true);
-
             // Remove all metadata fields, does not apply for redirector pages
             $fields->removeByName('Metadata');
 
             $fields->addFieldsToTab(
                 'Root.Main',
                 [
-                    new HeaderField('RedirectorDescHeader', _t(__CLASS__.'.HEADER', "This page will redirect users to another page")),
-                    new OptionsetField(
-                        "RedirectionType",
-                        _t(__CLASS__.'.REDIRECTTO', "Redirect to"),
+                    HeaderField::create(
+                        'RedirectorDescHeader',
+                        _t(__CLASS__.'.HEADER', "This page will redirect users to another page")
+                    ),
+                    OptionsetField::create(
+                        'RedirectionType',
+                        $this->fieldLabel('RedirectionType'),
                         [
                             "Internal" => _t(__CLASS__.'.REDIRECTTOPAGE', "A page on your website"),
                             "External" => _t(__CLASS__.'.REDIRECTTOEXTERNAL', "Another website"),
@@ -213,14 +218,8 @@ class RedirectorPage extends Page
                         ],
                         "Internal"
                     ),
-                    new TreeDropdownField(
-                        "LinkToID",
-                        _t(__CLASS__.'.YOURPAGE', "Page on your website"),
-                        SiteTree::class
-                    ),
-                    new UploadField('LinkToFile', _t(__CLASS__.'.FILE', "File")),
-                    new TextField("ExternalURL", _t(__CLASS__.'.OTHERURL', "Other website URL"))
-                ]
+                ],
+                'ExternalURL'
             );
         });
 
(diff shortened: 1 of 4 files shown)


code/Model/VirtualPage.php
```diff
@@ -77,8 +79,23 @@ class VirtualPage extends Page
 
     private static $db = [
         "VersionID" => "Int",
+        'CustomMetaDescription' => 'Text',
+        'CustomExtraMeta' => 'HTMLText'
```
ChloeSartorelli, line 83: These come from `SubsitesVirtualPage` (see https://github.com/silverstripe/silverstripe-subsites/pull/589). They make a lot of sense to add here, and not much sense to have explicitly only for the subsites version of this page.

code/Model/VirtualPage.php
```diff
@@ -77,8 +79,23 @@ class VirtualPage extends Page
...
+            'VersionID',
+            'CustomMetaDescription',
+            'CustomExtraMeta',
+        ],
+    ];
+
+    /**
+     * Whether to allow overriding the meta description and extra meta tags.
+     */
+    private static bool $allow_meta_overrides = true;
```
ChloeSartorelli, line 97: Minimises upgrades pain for people who don't want those new fields

lang/en.yml
ChloeSartorelli, line 1: None of these translations were being used, and instead explicit keys were being used. We're relying on scaffolding which relies on `fieldLabel()` now, and that method relies on the keys here. So I've updated their values to match what we expect.

code/Model/RedirectorPage.php
```diff
@@ -194,33 +198,28 @@ protected function onBeforeWrite()
     public function getCMSFields()
     {
         $this->beforeUpdateCMSFields(function (FieldList $fields) {
-            $fields->removeByName('Content', true);
-
             // Remove all metadata fields, does not apply for redirector pages
             $fields->removeByName('Metadata');
```
ChloeSartorelli, line 202: This is a composite field that gets manually added, so we can't use `ignoreFields` for it.

code/Model/SiteTree.php
```diff
@@ -2105,147 +2124,135 @@ public function VirtualPages()
      */
     public function getCMSFields()
     {
-        $dependentNote = '';
-        $dependentTable = new LiteralField('DependentNote', '<p></p>');
+        $this->beforeUpdateCMSFields(function (FieldList $fields) {
+            $dependentNote = '';
+            $dependentTable = new LiteralField('DependentNote', '<p></p>');
```
```suggestion
            $dependentTable = LiteralField::create('DependentNote', '<p></p>');
```

code/Model/SiteTree.php
```diff
@@ -2105,147 +2124,135 @@ public function VirtualPages()
...
-        $dependentPages = $this->DependentPages();
-        $dependentPagesCount = $dependentPages->count();
-        if ($dependentPagesCount) {
-            $dependentColumns = [
-                'Title' => $this->fieldLabel('Title'),
-                'DependentLinkType' => _t(__CLASS__.'.DependtPageColumnLinkType', 'Link type'),
-            ];
-            if (class_exists(Subsite::class)) {
-                $dependentColumns['Subsite.Title'] = Subsite::singleton()->i18n_singular_name();
+                $dependentNote = new LiteralField('DependentNote', '<p>' . _t(__CLASS__.'.DEPENDENT_NOTE', 'The following pages depend on this page. This includes virtual pages, redirector pages, and pages with content links.') . '</p>');
```
```suggestion
                $dependentNote = LiteralField::create('DependentNote', '<p>' . _t(__CLASS__.'.DEPENDENT_NOTE', 'The following pages depend on this page. This includes virtual pages, redirector pages, and pages with content links.') . '</p>');
```

code/Model/SiteTree.php
```diff
@@ -2105,147 +2124,135 @@ public function VirtualPages()
...
+            $fields->dataFieldByName('Content')?->addExtraClass('stacked');
+
+            // Metadata fields
+            $fields->addFieldsToTab('Root.Main', [
+                ToggleCompositeField::create(
+                    'Metadata',
+                    _t(__CLASS__.'.MetadataToggle', 'Metadata'),
+                    [
+                        $metaFieldDesc = new TextareaField("MetaDescription", $this->fieldLabel('MetaDescription')),
+                        $metaFieldExtra = new TextareaField("ExtraMeta", $this->fieldLabel('ExtraMeta'))
```
```suggestion
                        $metaFieldDesc = TextareaField::create("MetaDescription", $this->fieldLabel('MetaDescription')),
                        $metaFieldExtra = TextareaField::create("ExtraMeta", $this->fieldLabel('ExtraMeta'))
```

lang/en.yml
```diff
@@ -163,10 +163,10 @@ en:
     REDIRECTTOPAGE: 'A page on your website'
     SINGULARNAME: 'Redirector Page'
     YOURPAGE: 'Page on your website'
-    db_ExternalURL: 'External URL'
-    db_RedirectionType: 'Redirection type'
-    has_one_LinkTo: 'Link to'
-    has_one_LinkToFile: 'Link to file'
+    db_ExternalURL: 'Other website URL'
```
emteknetnz, line 166: We shouldn't change translation source values, as when we tx push it'll delete any existing translations

Create new keys if you need to change values

emteknetnz (changes requested):

lang/en.yml
ChloeSartorelli: See https://github.com/silverstripe/silverstripe-cms/pull/2983#discussion_r1713062292

I've linked to that comment from other PRs that you've approved, so presumably you broadly agree with it?

Note that `db` keys and `has_one` keys are directly used in `fieldLabels()` in a dynamic way. We can't really substitute them with other keys in a particularly clean way or I'd just use the existing ones that I'm intentionally swapping away from (again see the comment linked above).

code/Model/SiteTree.php
ChloeSartorelli: I didn't add this code - and broadly changing from `new` to `create` should be done as part of https://github.com/silverstripe/silverstripe-framework/issues/10269.

I'll make the changes here just to avoid ping pong.

ChloeSartorelli: Done

code/Model/SiteTree.php
ChloeSartorelli: Done

code/Model/SiteTree.php
ChloeSartorelli: Done

emteknetnz (approved):

PR #2985, opened 2024-08-13
ENH Don't emit deprecation warnings for unavoidable API calls

GuySartorelli:
`Versioned::canArchive()` is deprecated in https://github.com/silverstripe/silverstripe-versioned/pull/461

We can't stop calling it directly until CMS 6 because people may have implemented extensions that update the result of this permission check.

## Issue
- https://github.com/silverstripe/silverstripe-versioned/issues/447

### DIFF
--- a/code/BatchActions/CMSBatchAction_Archive.php
+++ b/code/BatchActions/CMSBatchAction_Archive.php
@@ -5,6 +5,7 @@ namespace SilverStripe\CMS\BatchActions;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\Admin\CMSBatchAction;
 use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Dev\Deprecation;
 
 /**
  * Delete items batch action.
@@ -27,6 +28,7 @@ class CMSBatchAction_Archive extends CMSBatchAction
 
     public function applicablePages($ids)
     {
-        return $this->applicablePagesHelper($ids, 'canArchive');
+        // canArchive() is deprecated, not $this->applicablePagesHelper()
+        return Deprecation::withNoReplacement(fn() => $this->applicablePagesHelper($ids, 'canArchive'));
     }
 }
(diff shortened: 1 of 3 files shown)


code/BatchActions/CMSBatchAction_Archive.php
```diff
@@ -27,6 +28,6 @@ public function run(SS_List $pages): HTTPResponse
 
     public function applicablePages($ids)
     {
-        return $this->applicablePagesHelper($ids, 'canArchive');
+        return Deprecation::withNoReplacement(fn() => $this->applicablePagesHelper($ids, 'canArchive'));
```
emteknetnz, line 31:
```suggestion
        return Deprecation::withNoReplacement(function () {
            // canArchive() is deprecated, not $this->applicablePagesHelper()
            return $this->applicablePagesHelper($ids, 'canArchive'));
        }
```

code/Controllers/CMSMain.php
```diff
@@ -1997,7 +1998,7 @@ public function archive(array $data, Form $form): HTTPResponse
         if (!$record || !$record->exists()) {
             throw new HTTPResponse_Exception("Bad record ID #$id", 404);
         }
-        if (!$record->canArchive()) {
+        if (!Deprecation::withNoReplacement(fn() => $record->canArchive())) {
```
emteknetnz:
```suggestion
        $canArchive = $record->canArchive();
        if (!canArchive)) {
```
Also update other instance in this PR

emteknetnz (changes requested):

code/BatchActions/CMSBatchAction_Archive.php
emteknetnz, line 32:
```suggestion
        // canArchive() is deprecated, not $this->applicablePagesHelper()
        return Deprecation::withNoReplacement(fn() => $this->applicablePagesHelper($ids, 'canArchive'));
```

emteknetnz (changes requested):

code/BatchActions/CMSBatchAction_Archive.php
ChloeSartorelli: I don't see the point of the comment but I added it to avoid ping pong.

code/Controllers/CMSMain.php
ChloeSartorelli: Think I got them all.

emteknetnz (approved):

Issue #2439, opened 2019-05-23
Cannot edit a link if markup is in the link text


ScopeyNZ:
## Description

Currently if there's a link in a TinyMCE (`HtmlEditorField`) that has markup within the `<a>` tag then you cannot edit it again. You don't get the little popover with edit/remove options. It's a little more nuanced, there doesn't seem to be a click handler on children on the `<a>` tag. You can still trigger the popover if you are clicking the `a` tag and not a child element of the `a` tag.

## Steps to reproduce

- Add some text to the TinyMCE field and highlight it
- Bold the text
- Add a link to that highlighted text
- Note that you can't edit the text

## Notes

- Tested with CWP 2.3.0-rc1 and silverstripe/installer@4.4.x-dev
- Tested on TinyMCE's site with their demos - it works a little differently - there's no popover, but it does work.
- The workaround is to either edit the source and remove inner HTML of the anchor tag or you can use the buttons (like the bold button) to remove the markup.
- Un-bolding anrd re-bolding the text fixes the issue.

## PRs

- https://github.com/silverstripe/silverstripe-cms/pull/2993
- https://github.com/silverstripe/silverstripe-admin/pull/1814

obj63mc:
We have run into this issue on a site that is utilizing a span tag inside links to handle button animation.  Is there any chance this gets fixed in SS5.x or even v6?

GuySartorelli:
PR merged - will be automatically tagged by GitHub actions

PR #2993, opened 2024-08-20
Fix TinyMCE edit link when link contains html

satrun77:
## Description

If you have a link in tinymce like `<a href=""><span>Title</span></a>`, then you can edit the link. The HTML inside the link tag can be any tag. 

The issue is that the selected node is the inside HTML tag instead of the link tag.

I'm not 100% keen on the solution using `while` but this is legacy code, so might be ok! Alternative code is welcomed :) 

This PR is needed for https://github.com/silverstripe/silverstripe-admin/pull/1814

## Manual testing steps

1. Install fresh CMS
2. Edit a page
3. Add a link (internal or external)
4. Edit HTML code in TinyMCE, add any HTML tags inside the link
5. Close the HTML code view in TInyMCE
6. Try to edit the link, it won't show the popup form.

## Issues

- https://github.com/silverstripe/silverstripe-cms/issues/2439

## Pull request checklist

    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [ ] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] CI is green


client/src/legacy/TinyMCE_sslink-internal.js
```diff
@@ -119,7 +119,17 @@ jQuery.entwine('ss', ($) => {
...
+
+      // Find "a" node, issue https://github.com/silverstripe/silverstripe-cms/issues/2439
+      let linkNode = editor.getSelectedNode();
+      if (linkNode.nodeName !== 'A') {
+        let count = 0;
+        while (count < 10 && linkNode.parentNode && linkNode.nodeName !== 'A') {
+          linkNode = linkNode.parentNode;
+          count +=1;
+        }
+      }
```
ChloeSartorelli, line 131: Please don't use a while loop - especially with a completely arbitrary number of iterations. If there's a function available to recursively check the ancestors of the element, use that. Otherwise, this might need to go back to the drawing board.

client/src/legacy/TinyMCE_sslink-internal.js
```diff
@@ -119,7 +119,17 @@ jQuery.entwine('ss', ($) => {
 
     getOriginalAttributes() {
       const editor = this.getElement().getEditor();
-      const node = $(editor.getSelectedNode());
+
+      // Find "a" node, issue https://github.com/silverstripe/silverstripe-cms/issues/2439
```
ChloeSartorelli, line 123:
```suggestion
      // Try to find a parent <a> tag in case the user clicks on some child node inside a link
```

GuySartorelli (changes requested):
Thanks for tackling this - unfortunately there's some problems with the approach.

satrun77:
No longer needed! The fix in https://github.com/silverstripe/silverstripe-admin/pull/1814 fixes the issue

PR #3002, opened 2024-09-11
FIX: Remove legacy fields which prevent page publish (fixes #2455)

### LINKED ISSUE #2455: SiteTree won't write new versions if only a field named 'Status' is changed
Please don’t ask how much time I wasted because of this bug 😭😂

```php
<?php

class MyPage extends Page
{
    private static $db = [
        'Status' => 'Varchar(255)'
    ];

    {
            $fields->addFieldToTab('Root.Main', TextField::create('Status'));
        });

        return parent::getCMSFields();
    }

    public function onAfterPublish()
    {
        die('you will never get this, a-la-la-la-la');
    }
}
```

Make changes to the value of status, `onAfterPublish()` won’t be called. The only time it’ll be called is if you also make a change to another field’s value as well. There are other issues this bug causes, but this is the easiest to reproduce.

The cause is this line: https://github.com/silverstripe/silverstripe-cms/blob/8a44071ec63c33abcec0d2901081114b30ae1907/code/Model/SiteTree.php#L1588.

I’m not sure if we should just remove that whole code block, but at the very least we should remove `Status` from the array as it’s no longer a field on `SiteTree` (it hasn’t been since 2.4). We should also remove `ToDo`, `VersionID` and `SaveCount`. 

## PRs
- https://github.com/silverstripe/silverstripe-cms/pull/3002

lozcalver:
## Description

This updates the list of fields which are ignored when deciding whether a new version (and therefore a new live version) should be created. Also made it configurable while I was there.

## Manual testing steps

Create a page type with a field `Status`, save and publish. Then change _only_ the status field and publish again: a new version won’t be written, and the live site won’t update. If you change status and something else, it will.

## Issues

- #2455

## Pull request checklist

    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] CI is green

### DIFF
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -300,6 +300,15 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         "Stage",  "Live"
     ];
 
+    /**
+     * Fields which, if changed on their own, won't cause a new version/live record to be created
+     * @var string[]
+     */
+    private static array $fields_ignored_by_versioning = [
+        'HasBrokenFile',
+        'HasBrokenLink',
+    ];
+
     private static $default_sort = "\"Sort\"";
 
     /**
@@ -1692,7 +1701,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         }
 
         // Check to see if we've only altered fields that shouldn't affect versioning
-        $fieldsIgnoredByVersioning = ['HasBrokenLink', 'Status', 'HasBrokenFile', 'ToDo', 'VersionID', 'SaveCount'];
+        $fieldsIgnoredByVersioning = $this->config()->get('fields_ignored_by_versioning') ?? [];
         $changedFields = array_keys($this->getChangedFields(true, 2) ?? []);
 
         // This more rigorous check is inline with the test that write() does to decide whether or not to write to the


code/Model/SiteTree.php
```diff
@@ -1693,7 +1703,7 @@ protected function onBeforeWrite()
         }
 
         // Check to see if we've only altered fields that shouldn't affect versioning
-        $fieldsIgnoredByVersioning = ['HasBrokenLink', 'Status', 'HasBrokenFile', 'ToDo', 'VersionID', 'SaveCount'];
```
lozcalver, line 1695: `HasBrokenFile` and `HasBrokenLink` are the only two fields that actually exist anymore.

I thought about changing `VersionID` to `Version` (which is what the field is now named), but that would be a behavioural change. It would also be a bit weird: changing the version number in the `SiteTree` table without writing a version would mean that `SiteTree` and `SiteTree_Versions` could end up out of sync.

code/Model/SiteTree.php
```diff
@@ -301,6 +301,16 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
         "Stage",  "Live"
     ];
 
+    /**
+     * Fields which, if changed on their own, won't cause a new version/live record to be created
+     * @var string[]
+     * @config
```
```suggestion
```

GuySartorelli (changes requested):
Overall looks good, thanks!

Can you please retarget this to `5.2`? That branch is still supported for bug fixes so it'd be good to get the fix in there instead of having to wait for 5.3.0 stable to release.

GuySartorelli (approved):
LGTM, works well locally.

PR #3009, opened 2024-09-24
ENH Use symfony/validation logic

GuySartorelli:
Relies on changes in https://github.com/silverstripe/silverstripe-framework/pull/11399

## Issue
- https://github.com/silverstripe/.github/issues/193

### DIFF
--- a/code/Model/RedirectorPage.php
+++ b/code/Model/RedirectorPage.php
@@ -7,6 +7,7 @@ use SilverStripe\Assets\File;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\HeaderField;
 use SilverStripe\Forms\OptionsetField;
+use SilverStripe\Forms\UrlField;
 use SilverStripe\Versioned\Versioned;
 
 /**
@@ -46,6 +47,9 @@ class RedirectorPage extends Page
         'ignoreFields' => [
             'RedirectionType',
             'Content',
+        ],
+        'fieldClasses' => [
+            'ExternalURL' => UrlField::class,
         ],
     ];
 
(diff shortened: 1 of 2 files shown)


code/Model/RedirectorPage.php
```diff
@@ -171,35 +175,12 @@ public function syncLinkTracking()
...
-                ])) {
-                    // we only allow http(s) urls
-                    $this->ExternalURL = '';
-                }
-            } else {
-                // malformed URL to reject
-                $this->ExternalURL = '';
-            }
-        }
-    }
```
ChloeSartorelli, line 196: No need to set false if it's invalid, because the `URLField` validation won't allow an invalid value in the first place.
The trade off is we're not assuming `http` by default anymore, users will have to add that themselves.
The `URLField` validation has been updated to make it clear when that needs to be added.

code/Model/RedirectorPage.php
```diff
@@ -171,35 +175,12 @@ public function syncLinkTracking()
...
-
     public function getCMSFields()
     {
         $this->beforeUpdateCMSFields(function (FieldList $fields) {
             // Remove all metadata fields, does not apply for redirector pages
             $fields->removeByName('Metadata');
+            $fields->dataFieldByName('ExternalURL')?->setAllowRelativeProtocol(true);
```
ChloeSartorelli, line 183: As per the first condition in the now-removed `onBeforeWrite()`.

code/Model/RedirectorPage.php
```diff
@@ -171,35 +175,12 @@ public function syncLinkTracking()
         }
     }
 
-    protected function onBeforeWrite()
```
emteknetnz, line 174: Need to keep this is as you can programatically save things without it going via a UrlField

emteknetnz (changes requested):

code/Model/RedirectorPage.php
ChloeSartorelli: That felt very much like it was being done _instead_ of validation. I feel like validating the value in the better option.

I can add the URL validation in the `validate()` method for this class if you like? But I don't think it makes sense to alter the value that's being set, especially programatic values.

emteknetnz: OK makes sense, yes agree throwing a validation exception is better than setting the value to blank and writing

emteknetnz (approved):

emteknetnz (changes requested):
Behat failure

GuySartorelli:
Oops, I somehow missed a reference to the old `PasswordValidator`.
Should be fixed by https://github.com/silverstripe/silverstripe-behat-extension/pull/286

emteknetnz (approved):

PR #3012, opened 2024-10-10
API Deprecations for template layer

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11322

### DIFF
--- a/code/Controllers/ContentController.php
+++ b/code/Controllers/ContentController.php
@@ -297,9 +297,11 @@ class ContentController extends Controller
 
     /**
      * @return ArrayList<SiteTree>
+     * @deprecated 5.4.0 Use getMenu() instead. You can continue to use $Menu in templates.
      */
     public function Menu($level)
     {
+        Deprecation::noticeWithNoReplacment('5.4.0', 'Use getMenu() instead. You can continue to use $Menu in templates.');
         return $this->getMenu($level);
     }
 


tests/php/Controllers/ContentControllerTest.php
```diff
@@ -204,7 +205,7 @@ public function testGetViewer()
                 __DIR__
                 . '/themes/controllertest/templates/SilverStripe/CMS/Tests/Controllers/'
                 . 'ContentControllerTestPage_test.ss',
-                $viewer->templates()['main']
+                Deprecation::withSuppressedNotice(fn() => $viewer->templates())['main']
```
ChloeSartorelli: Deprecation added in https://github.com/silverstripe/silverstripe-framework/pull/11420

code/Controllers/ContentController.php
```diff
@@ -297,9 +297,13 @@ public function getMenu($level = 1)
...
      * @return ArrayList<SiteTree>
+     * @deprecated 5.4.0 Use getMenu() instead. You can continue to use $Menu in templates.
      */
     public function Menu($level)
     {
+        Deprecation::withSuppressedNotice(
+            fn() => Deprecation::notice('5.4.0', 'Use getMenu() instead. You can continue to use $Menu in templates.')
+        );
         return $this->getMenu($level);
     }
```
ChloeSartorelli: This method existed because of a limitation in the templating engine where if you use `$Menu(2)` for example, it would call `getMenu()` but not pass the args.

I've resolved that limitation, so it'll now pass the args through to the getter. So now we can get rid of this pointless method.

emteknetnz (approved):

PR #3015, opened 2024-10-14
API Update method signature to match parent class

GuySartorelli:
Two commits:

## Commit 1:

Quick tidyup to use the `ignore_menuitem` configuration property instead of removing the menu items at runtime.

## Commit 2:

Reflects changes in https://github.com/silverstripe/silverstripe-admin/pull/1836

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1761

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -1,10 +1,5 @@
 <?php
 
-use SilverStripe\Admin\CMSMenu;
-use SilverStripe\CMS\Controllers\CMSMain;
-use SilverStripe\CMS\Controllers\CMSPageAddController;
-use SilverStripe\CMS\Controllers\CMSPageEditController;
-use SilverStripe\CMS\Controllers\CMSPageSettingsController;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Core\Manifest\ModuleLoader;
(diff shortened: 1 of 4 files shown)


_config.php
```diff
@@ -31,8 +26,3 @@
     'sitetree_link',
     [SiteTree::class, 'link_shortcode_handler']
 );
-
-CMSMenu::remove_menu_class(CMSMain::class);
```
ChloeSartorelli, line 35: Note that `CMSMain` itself is a special case and should _not_ get `ignore_menuitem` set to true.

_config.php
```diff
@@ -1,10 +1,5 @@
 <?php
 
-use SilverStripe\Admin\CMSMenu;
-use SilverStripe\CMS\Controllers\CMSMain;
-use SilverStripe\CMS\Controllers\CMSPageAddController;
-use SilverStripe\CMS\Controllers\CMSPageEditController;
-use SilverStripe\CMS\Controllers\CMSPageSettingsController;
```
ChloeSartorelli, line 7: None of these are used anymore

emteknetnz (approved):

PR #3022, opened 2024-11-10
API Move some code to framework to be reusable

GuySartorelli:
This logic is used by `CMSMain` but needs to be callable on any hierarchical model.

See [the corresponding framework PR](https://github.com/silverstripe/silverstripe-framework/pull/11460) to see where this code is going.

See [the overall refactoring PR](https://github.com/silverstripe/silverstripe-cms/pull/3017) (or just look in CMSMain) to see where it's being used and why it needs to be moved. Basically, anything `CMSMain` calls needs to be available on a `DataObject` record which has the `Hierarchy` extension applied.

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2947

Issue #2947, opened 2024-05-30
Don't hardcode `SiteTree` in `CMSMain`


GuySartorelli:
`CMSMain` has a `tree_class` configuration property which is sometimes referenced - but sometimes `SiteTree` is explicitly hardcoded.
The hardcoded references and the assumptions about using `SiteTree` should be removed.

## Benefits
- opens a path for further refactoring to allow other hierarchical data to be displayed in a hierarchical way
- opens a path to 'demistify' `SiteTree` so it becomes just another subclass of `DataObject` rather than having its own code paths in admin
- opens a path to reduce code/feature duplication across admin and cms modules (e.g. the search UI, sending toasts, handling tabs/breadcurmbs and more)

## Related
- https://github.com/silverstripe/silverstripe-cms/issues/2933
- [original POC PR](https://github.com/silverstripe/silverstripe-cms/pull/2939) - this PR includes many additional notes and implementation details

## Notes
- There are many ways to make calls to `SiteTree` methods/properties more generic, such as:
    - Implementing a method on `CMSMain` that calls that method if it exists, but falls back on other things - e.g. get `MenuTitle` if there is one, but fall back on `Title`
    - Moving the main implementation of the method from `SiteTree` into `CMSMain`, and invoking an extension hook on the record (probably with `invokeWithExtensions()`) to allow updating the value
- Don't forget there's both a tree view _and_ a list view! Both of these need to work and have their strings updated.

## Acceptance criteria
- [x] A `getModelClass()` method is implemented, which returns the value stored in the `tree_class` configuration property
    - This is to conform with [the LeftAndMain refactor](https://github.com/silverstripe/silverstripe-admin/issues/1763)
- [x] All hardcoded references to `SiteTree` or `Page` use the `getModelClass()` method instead (including subclasses of `CMSMain`)
- [x] Nothing assumes the model has the `Versioned` extension
- [x] Whenever `CMSMain`, one of its subclasses, or a template tries to call a method or fetch a property that is explicitly implemented on `SiteTree`, this is handled in a generic way
    - If any method/property is explicitly required which isn't provided by the `DataObject` class or the `Hierarchy` extension, throw clear exceptions if they're missing. One example (and probably the only case) is `CMSEditLink`.
[...]

PR #3023, opened 2024-11-11
API Deprecate and update code in preparation for CMS 6

GuySartorelli:
Needs https://github.com/silverstripe/silverstripe-framework/pull/11461 for CI to go green - please merge that PR first.

This PR deprecates code that will be removed in https://github.com/silverstripe/silverstripe-cms/pull/3022, and adds the new `class_description` config.

Note that there are a lot of methods in https://github.com/silverstripe/silverstripe-cms/pull/3022 which are being removed from `SiteTree` but are being added to a superclass or to the `Hierarchy` extension in https://github.com/silverstripe/silverstripe-framework/pull/11460. The `Hierarchy` extension is _required_ to be on `SiteTree` for that class and CMSMain to work correctly, so the effect is that the methods are not removed from a downstream developer POV.
Downstream developers can still override the methods in subclasses _and_ call `parent::whatever()` as though the method was directly implemented on `SiteTree`.

## Issue

### DIFF
--- a/code/Model/RedirectorPage.php
+++ b/code/Model/RedirectorPage.php
@@ -23,7 +23,12 @@ use SilverStripe\Versioned\Versioned;
  */
 class RedirectorPage extends Page
 {
+    /**
+     * @deprecated 5.4.0 use class_description instead.
+     */
     private static $description = 'Redirects requests to another location';
+
+    private static string $class_description = 'Redirects requests to another location';
 
     private static $icon_class = 'font-icon-p-redirect';
 
--- a/code/Model/SiteTree.php
+++ b/code/Model/SiteTree.php
@@ -136,6 +136,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      * Used as a cache for `SiteTree::allowedChildren()`
      * Drastically reduces admin page load when there are a lot of page types
      * @var array
+     * @deprecated 5.4.0 will be moved to SilverStripe\ORM\Hierarchy\Hierarchy->cache_allowedChildren
      */
     protected static $_allowedChildren = [];
 
@@ -419,6 +420,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      *
      * @config
      * @var string
+     * @deprecated 5.4.0 use class_description instead.
      */
     private static $description = null;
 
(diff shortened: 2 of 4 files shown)


code/Model/SiteTree.php
```diff
@@ -136,6 +136,7 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
      * Used as a cache for `SiteTree::allowedChildren()`
      * Drastically reduces admin page load when there are a lot of page types
      * @var array
+     * @deprecated 5.4.0 will be moved to SilverStripe\ORM\Hierarchy\Hierarchy->cache_allowedChildren
      */
     protected static $_allowedChildren = [];
```
ChloeSartorelli, line 141: We don't really have a process for deprecating non-config properties - and this one is very unlikely to be used by downstream devs anyway. I won't mention it in the changelog unless asked to.

code/Model/RedirectorPage.php
```diff
@@ -23,8 +23,13 @@
  */
 class RedirectorPage extends Page
 {
+    /**
+     * @deprecated 5.4.0 use class_description instead.
+     */
     private static $description = 'Redirects requests to another location';
```
ChloeSartorelli, line 29: Note that while I have added the deprecation notice to subclasses to ensure notices are emitted appropriately, I will only mention the config on `SiteTree` being deprecated in the changelog.

code/Model/SiteTree.php
```diff
@@ -3248,25 +3256,24 @@ public function getPageIconURL()
...
         }
-        return $this->config()->get('description');
-    }
-
-    /**
-     * Get localised description for this page
-     *
-     * @return string|null
-     */
-    public function i18n_classDescription()
```
ChloeSartorelli, line 3263: `i18n_classDescription()` is removed here but will now exist on a parent class, so no need to deprecate.

code/Model/SiteTree.php
```diff
@@ -3248,25 +3256,24 @@ public function getPageIconURL()
      */
     public function classDescription()
     {
+        // Ensure base class has an appropriate description if not explicitly set,
+        // since we can't set that config for projects.
         $base = in_array(static::class, [Page::class, SiteTree::class]);
         if ($base) {
-            return $this->config()->get('base_description');
+            $baseDescription = static::config()->get('base_class_description');
```
emteknetnz, line 3263: Move logic in this method to a new getClassDescription() method and update classDescription() to call $this->getClassDescription().

Do the same with i18n_classDescription() /  getI18nClassDescipcription()

emteknetnz (changes requested):

code/Model/SiteTree.php
ChloeSartorelli: Conversation about this will be held in https://github.com/silverstripe/silverstripe-framework/pull/11461#discussion_r1839344166

emteknetnz (approved):

PR #3029, opened 2024-11-25
API Use dynamic config for ModalController forms

GuySartorelli:
Remove redundant extension in favour of new functionality in ModalController

Relies on changes in https://github.com/silverstripe/silverstripe-admin/pull/1860

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1765


client/src/legacy/TinyMCE_sslink-anchor.js
```diff
@@ -44,7 +44,7 @@ const plugin = {
 };
 
 const modalId = 'insert-link__dialog-wrapper--anchor';
-const sectionConfigKey = 'SilverStripe\\CMS\\Controllers\\CMSPageEditController';
+const sectionConfigKey = 'SilverStripe\\Admin\\LeftAndMain';
```
ChloeSartorelli, line 47: `LeftAndMain` now dynamically lists all link forms registered with `ModalController.link_modal_form_factories`

These do get inherited, so `CMSPageEditController` would still work, but it seems sensible to use the more generic controller here (especially since `CMSPageEditController` is not long for this world)

code/Controllers/CMSPageEditController.php
```diff
@@ -37,18 +36,11 @@ class CMSPageEditController extends CMSMain
...
             'form' => [
                 'AddToCampaignForm' => [
                     'schemaUrl' => $this->Link('schema/AddToCampaignForm'),
                 ],
-                'editorInternalLink' => [
-                    'schemaUrl' => $modalController->Link('schema/editorInternalLink'),
-                ],
-                'editorAnchorLink' => [
-                    'schemaUrl' => $modalController->Link('schema/editorAnchorLink/:pageid'),
-                ],
```
ChloeSartorelli, line 51: These get included in `LeftAndMain::getClientConfig()` now instead.

code/Forms/AnchorLinkFormFactory.php
```diff
@@ -10,7 +10,7 @@ protected function getFormFields($controller, $name, $context)
 
         // Ensure current page is selected
         $pageIDField = $fields->dataFieldByName('PageID');
-        $pageIDField->setValue((int)$context['PageID']);
+        $pageIDField->setValue((int)$context['ItemID']);
```
ChloeSartorelli, line 13: This was changed to be more generic so we can reuse it in other link forms if needed.

emteknetnz (approved):

PR #3017, opened 2024-10-16
API Make CMSMain more generic

GuySartorelli:
Remove hardcoded references to pages and SiteTree
Remove assumption that records are versioned
Remove or validate assumptions about methods on the model class
Improve general architecture of CMSMain

Relies on changes in https://github.com/silverstripe/silverstripe-admin/pull/1837

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2947

Issue #2951, opened 2024-05-30
Make an abstract class for managing hierarchical data in a tree view


GuySartorelli:
`SiteTree` is not the only model with hierarchical data. It would be good to be able to manage other hierarchical data in the same way we currently manage `SiteTree` records.

## Related
- https://github.com/silverstripe/silverstripe-cms/issues/2933

## Notes
- The exact name of the class isn't important - if you think of a better name, feel free. Though note this class does go hand-in-hand with the `Hierarchy` extension.
- There's a `RecordTypes()` method and a `getRecordTypes()` method which seem to have different purposes. Maybe rename one of those to something different.

## Acceptance Criteria
- Make a new abstract `HierarchyModelAdmin` class and move all of the code from `CMSMain` into it. This class should live in the admin module
- `getCMSTreeTitle()` in `HierarchyModelAdmin` returns something sensible instead of the site name (this is used as the "root" of the tree)
- Make `CMSMain` subclass `HierarchyModelAdmin` and only give it what it needs as a unique admin section (e.g. url segment, section title, model class)
- Delete the subclasses of `CMSMain` and instead make them action methods implemented in `HierarchyModelAdmin`
    - Make sure the correct (renamed) templates are used for these actions
- Rename all of the `CMSMain` templates so that they are named after the new class instead, and move them to the admin module
- Any JS and CSS which applied to `HierarchyModelAdmin` is moved to the admin module
- The `HierarchyModelAdmin` works without needing anything from the cms module
- Double check the `extends()` and `invokeWithExtensions()` that are being called on records and see if they would be better as extension hooks on the new class instead.

PR #3035, opened 2024-12-08
API Make CMSMain more generic

GuySartorelli:
Replaces https://github.com/silverstripe/silverstripe-cms/pull/3017

- Remove hardcoded references to pages and SiteTree
- Remove assumption that records are versioned
- Remove or validate assumptions about methods on the model class
- Improve general architecture of CMSMain
- Update localisation keys to be more standardised and to live in more appropriate places

Relies on changes in https://github.com/silverstripe/silverstripe-admin/pull/1837

## Issue


code/Controllers/CMSMain.php
```diff
@@ -150,35 +148,42 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
...
 
     protected function init()
     {
+        $modelClass = $this->getModelClass();
+        if (!$modelClass::has_extension(Hierarchy::class)) {
+            throw new LogicException($modelClass . ' must have the Hierarchy extension');
+        }
+        if (!DataObject::singleton($modelClass)->getSortField()) {
+            throw new LogicException($modelClass . ' must define a sort field');
+        }
```
ChloeSartorelli, line 185: This is debatable - I've got it here mostly because I don't want to rework the UI elements that assume records can be sorted. Realistically there probably isn't much value in using this UI if your record isn't sortable anyway.

code/Controllers/CMSMain.php
```diff
@@ -289,14 +277,18 @@ public function Link($action = null)
...
     }
 
-    public function LinkPages()
+    public function LinkRecords()
     {
-        return CMSPagesController::singleton()->Link();
+        $controller = $this;
+        if (static::class !== CMSMain::class) {
+            $controller = CMSMain::singleton();
+        }
```
ChloeSartorelli, line 285: This is a little untidy but necessary for now. Will be cleared up as part of https://github.com/silverstripe/silverstripe-cms/issues/2951 which removes the `CMSMain` subclasses.

code/Controllers/CMSMain.php
```diff
@@ -800,121 +796,120 @@ public function savetreenode(HTTPRequest $request): HTTPResponse
...
-    public function CanOrganiseSitetree()
+    public function canCreateTopLevel(): bool
     {
-        return Permission::check('SITETREE_REORGANISE');
+        // Hardcoded check for SiteTree
+        // Ideally CMSMain will get an abstract parent class with all of the generic logic,
+        // and then this conditional check can be removed in CMSMain itself.
+        if (is_a($this->getModelClass(), SiteTree::class, true)) {
+            return SiteConfig::current_site_config()->canCreateTopLevel();
+        }
```
ChloeSartorelli, line 860: Another thing that will be tidied up as part of https://github.com/silverstripe/silverstripe-cms/issues/2951

code/Controllers/CMSMain.php
```diff
@@ -800,121 +796,120 @@ public function savetreenode(HTTPRequest $request): HTTPResponse
...
+     * Note that a canEdit() check is performed separately.
+     */
+    public function canOrganiseTree(): bool
+    {
+        // Hardcoded check for SiteTree
+        // Ideally CMSMain will get an abstract parent class with all of the generic logic,
+        // and then this conditional check can be removed in CMSMain itself.
+        if (is_a($this->getModelClass(), SiteTree::class, true)) {
+            return (bool) Permission::check('SITETREE_REORGANISE');
+        }
```
ChloeSartorelli, line 876: Another thing that will be tidied up as part of https://github.com/silverstripe/silverstripe-cms/issues/2951

code/Controllers/CMSPageAddController.php
ChloeSartorelli, line 1: This whole class was just a couple of templates and a form. It's been moved into `CMSMainAddForm`, the associated template, and an action on `CMSMain` to get the form.

code/Controllers/CMSPagesController.php
ChloeSartorelli, line 1: This whole class was effectively just a placeholder to render some templates. Those templates have been distilled into `CMSMain_LeftPanel.ss` which is included in the appropriate places in `CMSMain_Content.ss` and `CMSMain_Tools.ss`

Note I'm not removing `CMSPageEditController` in this PR because it wasn't removable when the bulk of this work was done. That can be removed in https://github.com/silverstripe/silverstripe-cms/issues/2951 which has an explicit AC for that purpose.

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_Tools.ss
ChloeSartorelli, line 1: There was quite a lot of duplication between this and the old `CMSPagesController` templates. That duplication is now distilled into the `CMSMain_LeftPanel.ss` template.

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_PageList_Sidebar.ss
ChloeSartorelli, line 1: This just had a pointless include in it

code/Controllers/CMSMain.php
```diff
@@ -913,132 +794,120 @@ public function savetreenode(HTTPRequest $request): HTTPResponse
...
-     */
-    public function getSearchContext()
-    {
-        Deprecation::notice('5.4.0', 'Will be replaced with ' . SiteTree::class . '::getDefaultSearchContext().');
-        $context = SiteTree::singleton()->getDefaultSearchContext();
-
-        $this->extend('updateSearchContext', $context);
-
-        return $context;
-    }
```
ChloeSartorelli, line 1023: Not needed - this is just basically adding an extra extension hook on top of getDefaultSearchContext().
If we think that's valuable, it should be added to `DataObject::getDefaultSearchContext()` instead - though the fact that method doesn't already have an extension hook implies one isn't needed there. People can also override `getDefaultSearchContext()` in their `Page` or other classes to return different search contexts.

code/Controllers/CMSPageAddController.php
```diff
@@ -1,255 +0,0 @@
...
-    {
-        $className = isset($data['PageType']) ? $data['PageType'] : "Page";
-        $parentID = isset($data['ParentID']) ? (int)$data['ParentID'] : 0;
-
-        if (!$parentID && isset($data['Parent'])) {
-            $page = SiteTree::get_by_link($data['Parent']);
-            if ($page) {
-                $parentID = $page->ID;
-            }
-        }
```
ChloeSartorelli, line 217: This specific piece of code isn't replicated in the new form or action.
It was originally introduced in https://github.com/silverstripe/silverstripe-cms/commit/c247db51465973385fe2a678293d2f4a28f3b1af which doesn't really make it clear what the intended use case of it is. It doesn't seem to be needed at all, nothing that I can see sets the `Parent` key in the data array.

emteknetnz (approved):
Great work!

emteknetnz (changes requested):
Except now there's a merge conflict! :-)

emteknetnz (approved):

PR #3036, opened 2024-12-08
API Deprecate API that's being removed in CMS 6

GuySartorelli:
## Issue

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -87,7 +87,13 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
     /**
      * Unique ID for page icons CSS block
      */
-    const PAGE_ICONS_ID = 'PageIcons';
+    public const CMS_RECORD_ICONS_ID = 'PageIcons';
+
+    /**
+     * Unique ID for page icons CSS block
+     * @deprecated 5.4.0 use CMS_RECORD_ICONS_ID instead
+     */
+    const PAGE_ICONS_ID = CMSMain::CMS_RECORD_ICONS_ID;
 
     private static $url_segment = 'pages';
 
@@ -143,6 +149,7 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         'schema',
         'SearchForm',
         'SiteTreeAsUL',
+        'TreeAsUL',
         'getshowdeletedsubtree',
         'savetreenode',
         'getsubtree',
@@ -162,14 +169,17 @@ class CMSMain extends LeftAndMain implements CurrentPageIdentifier, PermissionPr
         'TreeIsFiltered' => 'Boolean',
         'AddForm' => 'HTMLFragment',
         'LinkPages' => 'Text',
+        'LinkRecords' => 'Text',
         'Link' => 'Text',
         'ListViewForm' => 'HTMLFragment',
         'ExtraTreeTools' => 'HTMLFragment',
         'PageList' => 'HTMLFragment',
         'PageListSidebar' => 'HTMLFragment',
         'SiteTreeHints' => 'HTMLFragment',
+        'TreeHints' => 'HTMLFragment',
         'SecurityID' => 'Text',
         'SiteTreeAsUL' => 'HTMLFragment',
+        'TreeAsUL' => 'HTMLFragment',
     ];
 
     private static $dependencies = [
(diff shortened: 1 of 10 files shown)


code/Controllers/CMSMain.php
```diff
@@ -223,19 +233,23 @@ public function getResponseNegotiator(): PjaxResponseNegotiator
...
     /**
      * Page list view for edit-form
      *
      * @return DBHTMLText
+     * @deprecated 5.4.0 Will be removed without equivalent functionality
      */
     public function PageListSidebar()
     {
+        Deprecation::noticeWithNoReplacment('5.4.0');
         return $this->renderWith($this->getTemplatesWithSuffix('_PageList_Sidebar'));
```
ChloeSartorelli, line 253: It's just a template that redirects to a different template. Completely pointless.

code/Controllers/CMSMain.php
```diff
@@ -223,19 +233,23 @@ public function getResponseNegotiator(): PjaxResponseNegotiator
      * Get pages listing area
      *
      * @return DBHTMLText
+     * @deprecated 5.4.0 Will be renamed to RecordList
      */
     public function PageList()
     {
+        Deprecation::noticeWithNoReplacment('5.4.0', 'Will be renamed to RecordList');
         return $this->renderWith($this->getTemplatesWithSuffix('_PageList'));
```
ChloeSartorelli, line 241: Note I didn't add an implementation for `RecordList()` because that will be using a renamed template, so people would update their calls to the method but not the template in 5 and then still have to make an update in 6 anyway.

It's extremely unlikely anyone will be calling this anyway so kinda moot.

code/Model/SiteTree.php
ChloeSartorelli, line 1: Note there's some API being removed from `SiteTree` in CMS 6 but _doesn't_ need deprecating because it's moved to the `Hierarchy` extension in framework, which `SiteTree` needs in order to function at all - so those methods will still be callable and overridable in subclasses anyway.

code/Controllers/CMSMain.php
```diff
@@ -1761,6 +1918,13 @@ public function currentPageID()
         return $id;
     }
 
+    public function currentRecordID(): ?int
+    {
+        $id = parent::currentRecordID();
+        $this->extend('updateCurrentPageID', $id);
+        return $id;
+    }
```
ChloeSartorelli, line 1927: Mirrors `currentPageID` above, but uses the same extension hook.
The CMS 6 PR will update the hook name and move it to `LeftAndMain::currentRecordID()` directly instead.

code/Controllers/CMSSiteTreeFilter.php
```diff
@@ -113,8 +114,12 @@ public function getNumChildrenMethod()
         return $this->numChildrenMethod;
     }
 
+    /**
+     * @deprecated 5.4.0 Will be renamed to getRecordClasses()
+     */
     public function getPageClasses($page)
     {
+        Deprecation::noticeWithNoReplacment('5.4.0', 'Will be renamed to getRecordClasses()');
```
ChloeSartorelli, line 122: Didn't add the new implementations of these two methods because they're defined by the interface.

code/Controllers/CMSMain.php
```diff
@@ -891,9 +1013,11 @@ public function ExtraTreeTools()
      * and can be modified on extensions through updateSearchContext
      *
      * @return \SilverStripe\ORM\Search\SearchContext
+     * @deprecated 5.4.0 Use SilverStripe\CMS\Model\SiteTree::getDefaultSearchContext() instead.
      */
     public function getSearchContext()
     {
+        Deprecation::notice('5.4.0', 'Use ' . SiteTree::class . '::getDefaultSearchContext() instead.');
```
emteknetnz: getDefaultSearchContext() is not quite equivalent as it does not call $this->extend('updateSearchContext', $context);

emteknetnz (changes requested):

code/Controllers/CMSMain.php
ChloeSartorelli: What would you like the message to be?
"Will be removed without equivalent functionality" isn't correct because `getDefaultSearchContext()` is _effectively_ equivalent, just not 100% identical.
Maybe "Will be replaced with SiteTree::getDefaultSearchContext()"?

emteknetnz: `Will be replaced with SiteTree::getDefaultSearchContext()` makes sense

You should also ensure that the/an extension hook makes it way into the CMS 6 version

ChloeSartorelli: > You should also ensure that the/an extension hook makes it way into the CMS 6 version

I don't see any value in that - devs can just override `getDefaultSearchContext()` in their subclasses of `SiteTree`. The fact that method doesn't already have an extension hook implies one isn't needed.

ChloeSartorelli: Updated the deprecation message.

emteknetnz (approved):

PR #3040, opened 2024-12-16
DEP Upgrade redux dependencies

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/333

Note that yarn build was run, though there are no changes to bundle files


GuySartorelli:
> Note that `yarn build` was run, though there are no changes to bundle files

This is expected, as redux is included in the webpack externals so the version provided by silverstripe/admin is ultimately used.

GuySartorelli (approved):
LGTM

PR #3050, opened 2025-01-15
DEP Upgrade bootstrap and reactstrap

GuySartorelli:
See https://github.com/silverstripe/silverstripe-admin/pull/1889 which links to the specific upgrade guide entries for most changes made.

## Issue
- https://github.com/silverstripe/.github/issues/332


client/src/styles/legacy/_CMSMain.scss
```diff
@@ -146,20 +146,6 @@
...
-    &:last-of-type {
-      z-index: 1;
-    }
-
-    .btn {
-      margin-top: 0;
-      padding-top: $input-padding-y;
-      padding-bottom: $input-padding-y;
-    }
-  }
```
ChloeSartorelli, line 161: https://getbootstrap.com/docs/5.0/migration/#forms

> Dropped `.input-group-append` and `.input-group-prepend`. You can now just add buttons and .input-group-text as direct children of the input groups.

Turns out the styling isn't needed anymore in any case, from what I can tell, so just removed this block outright.

emteknetnz (approved):

Issue #2949, opened 2024-05-30
Refactor tree filtering in `CMSMain`/`HierarchyModelAdmin`


GuySartorelli:
Form fields for filtering in `CMSMain` are hardcoded - and some of the functionality is duplicated or else needlessly different between this filter functionality and `GridFieldFilterHeader`.

## Relevant
- https://github.com/silverstripe/silverstripe-cms/issues/2933
- [original POC PR](https://github.com/silverstripe/silverstripe-cms/pull/2939) - this PR includes many additional notes and implementation details

## Acceptance Criteria
- [x] The fields to use for filtering are pulled from the model in the same way `GridFieldFilterHeader` does (i.e. it uses `searchableFields()`)
- [x] `SearchContext` is used for filtering in `CMSMain` the same way it is in `GridFieldFilterHeader`
- [x] Any `CMSSiteTreeFilter` logic is either made generic if possible, or else used by the `SearchContext` for `SiteTree` wherever filtering pages is used
    - i.e. filtering pages in a gridfield uses the same logic as filtering pages in `CMSMain`
- [x] A shared template and shared javascript/css are used for the filter in `GridFieldFilter` and `CMSMain` (with the ability to use different custom templates as desired - e.g. by trying to get specific templates by class name first, and falling back to the shared one)
- [x] As much as feasible, PHP logic for filtering in `GridFieldFilterHeader` and `CMSMain` is shared

### Enable `WithinRangeFilter` in `searchable_fields`

This set of PRs is required to match the ranged filtering against `LastEdited` which is currently only possible for `SiteTree` inside `CMSMain`.

- https://github.com/silverstripe/silverstripe-framework/pull/11609 (Reassign to Guy after merging this one)
- https://github.com/silverstripe/silverstripe-framework/pull/11602
- https://github.com/silverstripe/developer-docs/pull/687

### Avoid `filterByCallback()` in `CMSSiteTreeFilter` subclasses

This set of PRs is required to migrate the `SiteTree` filtering functionality to primarily use `SearchContext`, which requires filtering be done by the ORM not in PHP.

[**Kitchen sink CI run**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/13424395823)

- https://github.com/silverstripe/silverstripe-versioned/pull/489
- https://github.com/silverstripe/silverstripe-cms/pull/3057
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/388
[...]

PR #3052, opened 2025-01-30
API Refactor SiteTree filtering to work like GridField filters

GuySartorelli:
CMSMain uses an entirely separate code path for building the filter form and for the actual filter functionality compared with the gridfield filter used literally everywhere else in the CMS.

This aims to unify the two as much as possible, so that filtering pages in CMSMain is the same as filtering everywhere else.
It also generalises it which will help to unlock https://github.com/silverstripe/silverstripe-cms/issues/2951

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2949

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -2,7 +2,6 @@
 
 namespace SilverStripe\CMS\Controllers;
 
-use InvalidArgumentException;
 use LogicException;
 use Psr\SimpleCache\CacheInterface;
 use SilverStripe\Admin\AdminRootController;
@@ -13,11 +12,9 @@ use SilverStripe\Admin\Navigator\SilverStripeNavigator;
 use SilverStripe\CMS\BatchActions\CMSBatchAction_Archive;
 use SilverStripe\CMS\BatchActions\CMSBatchAction_Publish;
 use SilverStripe\CMS\BatchActions\CMSBatchAction_Unpublish;
-use SilverStripe\CMS\Controllers\CMSSiteTreeFilter_Search;
 use SilverStripe\CMS\Forms\CMSMainAddForm;
 use SilverStripe\CMS\Model\CurrentRecordIdentifier;
 use SilverStripe\CMS\Model\SiteTree;
-use SilverStripe\CMS\Search\SearchForm;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
@@ -33,10 +30,6 @@ use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Manifest\ModuleResource;
 use SilverStripe\Core\Manifest\ModuleResourceLoader;
-use SilverStripe\Dev\Deprecation;
-use SilverStripe\Forms\DateField;
-use SilverStripe\Forms\DropdownField;
-use SilverStripe\Forms\FieldGroup;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\Form;
 use SilverStripe\Forms\FormAction;
@@ -67,11 +60,10 @@ use SilverStripe\Security\PermissionProvider;
 use SilverStripe\Security\Security;
 use SilverStripe\Security\SecurityToken;
 use SilverStripe\SiteConfig\SiteConfig;
-use SilverStripe\Versioned\ChangeSet;
-use SilverStripe\Versioned\ChangeSetItem;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\VersionedAdmin\Controllers\CMSPageHistoryViewerController;
 use SilverStripe\Model\ArrayData;
+use SilverStripe\ORM\Search\SearchContextForm;
 use SilverStripe\Security\Permission;
 use SilverStripe\Security\PermissionCheckable;
 use SilverStripe\Versioned\RecursivePublishable;
@@ -136,7 +128,7 @@ class CMSMain extends LeftAndMain implements CurrentRecordIdentifier, Permission
         'submit',
         'EditForm',
         'schema',
-        'SearchForm',
+        'getSearchForm',
         'TreeAsUL',
         'getshowdeletedsubtree',
         'savetreenode',
@@ -151,6 +143,7 @@ class CMSMain extends LeftAndMain implements CurrentRecordIdentifier, Permission
 
     private static array $url_handlers = [
         'EditForm/$ID' => 'EditForm',
+        'GET SearchForm' => 'getSearchForm',
     ];
 
     private static array $casting = [
@@ -497,11 +490,10 @@ class CMSMain extends LeftAndMain implements CurrentRecordIdentifier, Permission
     public function TreeAsUL()
     {
         $modelClass = $this->getModelClass();
-        $filter = $this->getSearchFilter();
 
         DataObject::singleton($modelClass)->prepopulateTreeDataCache(null, [
-            'childrenMethod' => $filter ? $filter->getChildrenMethod() : 'AllChildrenIncludingDeleted',
-            'numChildrenMethod' => $filter ? $filter->getNumChildrenMethod() : 'numChildren',
+            'childrenMethod' => 'AllChildrenIncludingDeleted',
+            'numChildrenMethod' => 'numChildren',
         ]);
 
         $html = $this->getTreeFor($modelClass);
@@ -531,21 +523,6 @@ class CMSMain extends LeftAndMain implements CurrentRecordIdentifier, Permission
         $nodeCountThreshold = null
     ) {
         $nodeCountThreshold = is_null($nodeCountThreshold) ? Config::inst()->get($className, 'node_threshold_total') : $nodeCountThreshold;
-        // Provide better defaults from filter
-        $filter = $this->getSearchFilter();
-        if ($filter) {
-            if (!$childrenMethod) {
-                $childrenMethod = $filter->getChildrenMethod();
-            }
-            if (!$numChildrenMethod) {
-                $numChildrenMethod = $filter->getNumChildrenMethod();
-            }
-            if (!$filterFunction) {
-                $filterFunction = function ($node) use ($filter) {
-                    return $filter->isRecordIncluded($node);
-                };
-            }
-        }
 
         // Build set from node and begin marking
         $record = ($rootID) ? $this->getRecord($rootID) : null;
@@ -626,15 +603,6 @@ class CMSMain extends LeftAndMain implements CurrentRecordIdentifier, Permission
             foreach ($statuses as $s) {
                 $classes .= ' status-' . $s;
             }
-        }
-
-        // Get additional filter classes
-        $filter = $this->getSearchFilter();
-        if ($filter && ($filterClasses = $filter->getRecordClasses($node))) {
-            if (is_array($filterClasses)) {
-                $filterClasses = implode(' ', $filterClasses);
-            }
-            $classes .= ' ' . $filterClasses;
         }
 
         return trim($classes ?? '');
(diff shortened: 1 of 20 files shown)


templates/SilverStripe/CMS/Controllers/Includes/CMSMain_LeftPanel.ss
```diff
@@ -12,13 +12,11 @@
             <%-- Full breadcrumbs (useful for tree view which isn't available when viewing an edit form) --%>
             <% include SilverStripe\\Admin\\CMSBreadcrumbs %>
         <% end_if %>
-        <% include SilverStripe\\CMS\\Controllers\\CMSMain_Filter %>
+        $SearchForm.FilterButton($TreeIsFiltered)
```
ChloeSartorelli, line 15: Love that we can pass variables into function calls from templates now 👌

code/Controllers/CMSSiteTreeFilter_ChangedPages.php
```diff
@@ -3,28 +3,26 @@
...
 {
-
-    public static function title()
+    public static function title(): string
     {
         return _t(__CLASS__ . '.Title', "Modified pages");
     }
 
-    public function getFilteredPages()
+    public function getFilteredPages(DataList $list): DataList
```
ChloeSartorelli, line 19: These `CMSSiteTreeFilter` classes used to do _all_ of the filtering, basically duplicating the work `SearchContext` is meant to do.

They've now been updated to only update a list to filter by status instead of creating a new filtered list. See `SiteTreeSearchContext` for the usage.

code/Controllers/CMSSiteTreeFilter_PublishedPages.php
```diff
@@ -14,41 +14,22 @@
...
-        $pages = $this->applyDefaultFilters($pages);
-
-        return $pages;
+        $list = Versioned::updateListToAlsoIncludeDeleted($list);
+        $baseTable = SiteTree::singleton()->baseTable();
+        $liveTable = SiteTree::singleton()->stageTable($baseTable, Versioned::LIVE);
+        return $list->innerJoin(
+            $liveTable,
+            "\"{$baseTable}_Versions\".\"RecordID\" = \"$liveTable\".\"ID\""
+        );
```
ChloeSartorelli, line 33: Unrelated refactoring to not use hardcoded table names

code/Controllers/CMSMain.php
```diff
@@ -496,11 +490,10 @@ public function AddForm(): CMSMainAddForm
     public function TreeAsUL()
     {
         $modelClass = $this->getModelClass();
-        $filter = $this->getSearchFilter();
```
ChloeSartorelli, line 500: The search filter can't be used in the tree view anyway, so this wasn't doing anything

code/Controllers/CMSMain.php
```diff
@@ -530,21 +523,6 @@ public function getTreeFor(
...
-            }
-            if (!$numChildrenMethod) {
-                $numChildrenMethod = $filter->getNumChildrenMethod();
-            }
-            if (!$filterFunction) {
-                $filterFunction = function ($node) use ($filter) {
-                    return $filter->isRecordIncluded($node);
-                };
-            }
-        }
```
ChloeSartorelli, line 548: The search filter can't be used in the tree view anyway, so this wasn't doing anything

code/Controllers/CMSMain.php
```diff
@@ -627,15 +605,6 @@ public function getTreeNodeClasses(DataObject $node): string
...
         }
 
-        // Get additional filter classes
-        $filter = $this->getSearchFilter();
-        if ($filter && ($filterClasses = $filter->getRecordClasses($node))) {
-            if (is_array($filterClasses)) {
-                $filterClasses = implode(' ', $filterClasses);
-            }
-            $classes .= ' ' . $filterClasses;
-        }
```
ChloeSartorelli, line 638: The search filter can't be used in the tree view anyway, so this wasn't doing anything

code/Controllers/CMSMain.php
```diff
@@ -894,110 +863,17 @@ public function ExtraTreeTools(): string
         return $html;
     }
 
-    /**
-     * Returns the search form schema for the current model
-     */
-    public function getSearchFieldSchema(): string
```
ChloeSartorelli, line 904: Moved to framework to be shared with gridfield

code/Controllers/CMSMain.php
```diff
@@ -1658,58 +1534,15 @@ public function childfilter(HTTPRequest $request): HTTPResponse
...
     }
 
-    /**
-     * Safely reconstruct a selected filter from a given set of query parameters
-     *
-     * @param array $params Query parameters to use, or null if none present
-     * @return CMSSiteTreeFilter The filter class
-     * @throws InvalidArgumentException if invalid filter class is passed.
-     */
-    protected function getQueryFilter($params)
```
ChloeSartorelli, line 1677: No longer needed - let the search context handle it

code/Controllers/CMSMain.php
```diff
@@ -1658,58 +1534,15 @@ public function childfilter(HTTPRequest $request): HTTPResponse
...
-     * Returns the records meet a certain criteria as {@see CMSSiteTreeFilter} or the subrecords of a parent record
-     * defaulting to no filter and show all records in first level.
-     * Doubles as search results, if any search parameters are set through {@link SearchForm()}.
-     *
-     * @param array $params Search filter criteria
-     * @param int $parentID Optional parent node to filter on (can't be combined with other search criteria)
-     * @return SS_List
-     * @throws InvalidArgumentException if invalid filter class is passed.
-     */
-    public function getList($params = [], $parentID = 0)
```
ChloeSartorelli, line 1701: Was only used in `ListViewForm` which now just invokes search context isntead

tests/behat/features/search-for-a-page.feature
```diff
@@ -11,47 +11,47 @@ Feature: Search for a page
...
-  #     But I should not see "Old Page" in the cms list
+  Scenario: I can search for a page by its oldest last edited date
+    Given a "page" "Recent Page" was last edited "08-08-2007"
+      And a "page" "Old Page" was last edited "03-03-2007"
+    And I take a screenshot after every step
+    When I press the "Advanced" button
+      And I fill in "Search__LastEdited_SearchFrom" with "06-06-2007"
+      And I press the "Search" button
+    Then I should see "Recent Page" in the cms list
+      But I should not see "Old Page" in the cms list
```
ChloeSartorelli, line 38: This and the below are our only behat tests for the ranged filtering (though we do have good unit test coverage) - so I'm reviving these.

tests/behat/features/manage-page-permissions.feature
```diff
@@ -14,13 +14,12 @@ Feature: Manage page permissions
...
-  #   And I press the "Save" button
-  #   When I am not logged in
-  #   And I go to the homepage
-  #   Then I should see "Welcome"
+  Scenario: I can open view permissions to everyone
+    Given I select "Anyone" from "Who can view this page?" input group
+    And I press the "Save" button
+    When I am not logged in
+    And I go to the homepage
+    Then I should see "Welcome"
```
ChloeSartorelli, line 22: Unrelated - spotted this commented out test, ran it, and it passes so I'm uncommenting it.

code/Controllers/CMSSiteTreeFilter.php
```diff
@@ -96,169 +42,13 @@ public static function get_all_filters()
...
-        return null;
-    }
-
-    /**
-     * Gets the list of filtered pages
-     *
-     * @see {@link ModelData::getStatusFlags()}
-     * @return SS_List
-     */
-    abstract public function getFilteredPages();
```
ChloeSartorelli, line 151: See bottom of the diff for this class - this one stays.

code/Controllers/CMSMain.php
```diff
@@ -1666,62 +1534,21 @@ public function childfilter(HTTPRequest $request): HTTPResponse
...
      * @return Form
      */
     public function ListViewForm()
     {
-        $params = $this->getRequest()->requestVar('q');
+        $modelClass = $this->getModelClass();
         $parentID = $this->getRequest()->requestVar('ParentID');
-        // Set default filter if other params are set
-        if ($params && empty($params['FilterClass'])) {
-            $params['FilterClass'] = CMSSiteTreeFilter_Search::class;
```
ChloeSartorelli, line 1722: Let the search context handle this

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_Filter.ss
ChloeSartorelli, line 1: Replaced with reusable markup in framework

emteknetnz (commented):
Seems good, won't approve just yet (need to fully get my head around the other PRs) though also not requesting any changes at this stage either

GuySartorelli:
If the reason to not approve is because of the elemental PR, please just approve. This PR does not rely on elemental, that's the opposite direction any coupling should have.

emteknetnz:
It's not the elemental PR, I'm referring to admin + framework PRs

code/Model/SiteTree.php
```diff
@@ -2844,6 +2871,52 @@ public function isHomePage(): bool
...
+                'title' => _t(__CLASS__ . '.Filter_Status', 'Page status'),
+                'field' => DropdownField::class,
+            ]
+        ], $fields);
+        return $fields;
+    }
+
+    public function scaffoldSearchFields($_params = null)
+    {
+        $fields = parent::scaffoldSearchFields($_params);
```
emteknetnz, line 2900:
```suggestion
    public function scaffoldSearchFields($params = null)
    {
        $fields = parent::scaffoldSearchFields($params);
```
Any reason for the `_` in the variable name?

code/Model/SiteTree.php
```diff
@@ -2844,6 +2871,52 @@ public function isHomePage(): bool
...
+
+        // Update "Status" sources
+        $filters = CMSSiteTreeFilter::get_all_filters();
+        // Remove 'All records' as we set that to empty/default value
+        unset($filters[CMSSiteTreeFilter_Search::class]);
+        $fields->dataFieldByName('FilterClass')?->setSource($filters)->setEmptyString(CMSSiteTreeFilter_Search::singleton()->title());
```
```suggestion
        $filterClass = $fields->dataFieldByName('FilterClass');
        if ($filterClass) {
            $title = CMSSiteTreeFilter_Search::singleton()->title();
            $filterClass->setSource($filters)->setEmptyString($title);
        }
```
I don't like everything on one line, it's hard to read.

code/Model/SiteTree.php
```diff
@@ -2844,6 +2871,52 @@ public function isHomePage(): bool
...
+
+        // Update "Page type" sources
+        $classNames = [];
+        $validClasses = ClassInfo::getValidSubClasses(SiteTree::class);
+        $this->invokeWithExtensions('updateAllowedSubClasses', $validClasses);
+        foreach ($validClasses as $class) {
+            $classNames[$class] = SiteTree::singleton($class)->i18n_singular_name();
+        }
+        $fields->dataFieldByName('ClassName')?->setSource($classNames)->setEmptyString(_t(__CLASS__ . '.Filter_ClassName_Any', 'Any'));
```
emteknetnz: Split this out across multiple lines like recommended above

code/Controllers/CMSMain.php
```diff
@@ -1666,62 +1534,21 @@ public function childfilter(HTTPRequest $request): HTTPResponse
...
+        $params = $this->getRequest()->requestVar('q') ?? [];
+        if ($params) {
+            $form = $this->getSearchForm();
+            $params = $form->prepareValuesForSearchContext($params);
+            $list = DataObject::singleton($modelClass)->getDefaultSearchContext()->getQuery($params);
+        } else {
+            $list = DataObject::singleton($modelClass)::get()->filter('ParentID', is_numeric($parentID) ? $parentID : 0);
```
emteknetnz, line 1550: We should be instead throwing an exception if $parentID isn't numeric? There's no reason for it to be non-numeric?

code/Model/SiteTree.php
```diff
@@ -289,8 +295,29 @@ class SiteTree extends DataObject implements PermissionProvider, i18nEntityProvi
...
+        ],
+        'LastEdited' => [
+            'general' => false,
+            'filter' => WithinRangeFilter::class,
+            'field' => DateField::class,
+        ],
+        // The below fields are excluded from the filter form but
+        // are included in the general search
+        'Title' => [
+            'field' => HiddenField::class,
```
emteknetnz, line 310: Would it be better to have these as LiteralField's so that no HTML renders on the frontend, rather than have a bunch of hidden fields render?

emteknetnz (changes requested):

code/Model/SiteTree.php
ChloeSartorelli: It matches the name in the parent class - since named arguments are a thing in PHP it's best to keep the names consistent across subclasses.

code/Model/SiteTree.php
ChloeSartorelli: Done though I disagree

code/Model/SiteTree.php
ChloeSartorelli: Done

code/Controllers/CMSMain.php
ChloeSartorelli: I'm just using the same logic that was already in place. https://github.com/silverstripe/silverstripe-cms/blob/c989e4f39bc5fad3527875ae3fce1502dfdcec1b/code/Controllers/CMSMain.php#L1812
Changing that isn't in scope - and given the short time til the beta I'd prefer not to be making changes we don't need to.

code/Model/SiteTree.php
ChloeSartorelli: `HiddenField` is a "data" field, but `LiteralField` isn't, which means there are different methods for getting the form fields. Changing this could result in some functionality not working as expected.

It would also require special handling to avoid the title (second argument to the constructor when scaffolding the form fields) from being displayed as the "content" of the `LiteralField`.

emteknetnz (approved):

GuySartorelli:
Sorry, looks like I forgot to push some of those requested changes til now.

emteknetnz (approved):

PR #3059, opened 2025-02-18
API Remove code which is being migrated to a new TinyMCE module

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-htmleditor-tinymce/issues/2


tests/behat/src/AnchorContext.php
ChloeSartorelli, line 1: This is moved out of `FixtureContext` so it can be used in the new htmleditor behat without clashing with the asset-admin context that has the same parent class as `FixtureContext`.

emteknetnz (approved):

PR #3027, opened 2024-11-20
API Remove silverstripe/campaign-admin integration support

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/342

PR #3062, opened 2025-03-04
ENH Add campaign-admin support back in

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/895

This contains some of the code that was originally reverted https://github.com/silverstripe/silverstripe-cms/pull/3027 - the rest of it has been moved into campaign-admin

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -69,6 +69,8 @@ use SilverStripe\Security\PermissionCheckable;
 use SilverStripe\Versioned\RecursivePublishable;
 use SilverStripe\View\Requirements;
 use SilverStripe\View\ThemeResourceLoader;
+use SilverStripe\Versioned\ChangeSetItem;
+use SilverStripe\Versioned\ChangeSet;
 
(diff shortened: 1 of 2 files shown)


code/Controllers/CMSMain.php
ChloeSartorelli, line 1: Can these changes not be made by adding new extension hooks, so the code can live in the campaign admin module?

GuySartorelli (changes requested):

code/Controllers/CMSMain.php
emteknetnz: Yes, have moved to campaign-admin

emteknetnz:
Closing PR as it's no longer needed. The code in this PR was moved to an extension in the campaign-admin PR, and the unit tests are essentially duplicates of what's already in https://github.com/silverstripe/silverstripe-campaign-admin/blob/2/tests/php/SiteTreeExtensionTest.php

A new PR has been opened that adds an extension hook https://github.com/silverstripe/silverstripe-cms/pull/3065

PR #3064, opened 2025-03-07
API Deprecate API that will be removed in CMS 6

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-htmleditor-tinymce/issues/2


tests/behat/src/FixtureContext.php
ChloeSartorelli, line 1: This is a bit of a weird one to deprecate but:
a) it does no harm
b) some people _do_ use behat and we've had at least one complain in the past with changes like this which weren't announced

emteknetnz: Extra context - moved because https://github.com/silverstripe/silverstripe-cms/pull/3059/files#r1984185238

emteknetnz (approved):

PR #3072, opened 2025-04-09
ENH Ensure report links are escaped

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/897

If $siteTree->URLSegment somehow got XSS on to it this would protect against it from showing in the report

We do something similar already on the external links report https://github.com/silverstripe/silverstripe-externallinks/blob/3/src/Reports/BrokenExternalLinksReport.php#L41


GuySartorelli (approved):
LGTM, matches what was approved privately.

PR #3077, opened 2025-05-06
ENH More efficient hierarchy querying

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11703

Requires https://github.com/silverstripe/silverstripe-framework/pull/11720

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -490,15 +490,25 @@ class CMSMain extends LeftAndMain implements CurrentRecordIdentifier, Permission
     public function TreeAsUL()
     {
         $modelClass = $this->getModelClass();
-
-        DataObject::singleton($modelClass)->prepopulateTreeDataCache(null, [
-            'childrenMethod' => 'AllChildrenIncludingDeleted',
-            'numChildrenMethod' => 'numChildren',
-        ]);
-
+        $options = $this->getTreeAsULPrepopulateOptions($modelClass);
+        DataObject::singleton($modelClass)->prepopulateTreeDataCache(null, $options);
         $html = $this->getTreeFor($modelClass);
         $this->extend('updateTreeAsUL', $html);
         return $html;
+    }
+
+    /**
+     * Get the options used for the call to Hierarchy::prepopulateTreeDataCache()
+     */
+    private function getTreeAsULPrepopulateOptions(string $modelClass)
+    {
+        /** @var DataObject&Hierarchy $obj */
+        $obj = DataObject::singleton($modelClass);
+        $baseClass = $obj->getHierarchyBaseClass();
+        return [
+            'childrenMethod' => $baseClass::config()->get('tree_children_method'),
+            'numChildrenMethod' => 'numChildren',
+        ];
     }
 
     /**
--- a/templates/SilverStripe/CMS/Controllers/Includes/CMSMain_TreeView.ss
+++ b/templates/SilverStripe/CMS/Controllers/Includes/CMSMain_TreeView.ss
@@ -1,9 +1,5 @@
 <% include SilverStripe\\CMS\\Controllers\\CMSMain_ContentToolActions View='Tree' %>
 
-<div class="ss-dialog cms-page-add-form-dialog cms-dialog-content" id="cms-page-add-form" title="<%t SilverStripe\Admin\LeftAndMain.AddNew 'Add new {name}' name=$getRecord('singleton').i18n_singular_name().lowercase %>">
-	$AddForm
-</div>
-
 $ExtraTreeTools
 
 <% if $TreeIsFiltered %>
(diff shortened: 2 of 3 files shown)


templates/SilverStripe/CMS/Controllers/Includes/CMSMain_TreeView.ss
```diff
@@ -1,9 +1,5 @@
 <% include SilverStripe\\CMS\\Controllers\\CMSMain_ContentToolActions View='Tree' %>
 
-<div class="ss-dialog cms-page-add-form-dialog cms-dialog-content" id="cms-page-add-form" title="<%t SilverStripe\Admin\LeftAndMain.AddNew 'Add new {name}' name=$getRecord('singleton').i18n_singular_name().lowercase %>">
```
emteknetnz, line 3: This was adding a hidden and completely redundant CMSMainAddForm inbetween the row with the 'Add new page' button on it and the site tree - however because it contained a TreeDropdownField, which by default does an XHR request to /tree, it was spamming off pointless and expensive requests to fetch the site tree for no reason.

The actual place this CMSMainAddForm is used is on /admin/pages/add. I suspect people just forgot to delete the old one from /admin/pages.

Assuming we're in a non-blocking environment this one wouldn't be too bad from a user point of view as it just cutting down on an expensive background XHR request, though it was obviously a waste of server resources

SimulatedPanda:
I approve the inclusion of this work in CMS v6.0 when all necessary reviews have been completed and any issues have been addressed.

code/Controllers/CMSMain.php
```diff
@@ -492,7 +492,7 @@ public function TreeAsUL()
         $modelClass = $this->getModelClass();
 
         DataObject::singleton($modelClass)->prepopulateTreeDataCache(null, [
-            'childrenMethod' => 'AllChildrenIncludingDeleted',
+            'childrenMethod' => Hierarchy::config()->get('default_children_method'),
```
ChloeSartorelli: Given all the work we did to generalise `CMSMain` to pave the way for using it as a base for a general tree admin for any `Hierarchical` model, I think we should allow the base model to set the configuration.

This allows using different methods for different hierarchical models.
It also means we don't have to have `Configurable` on the extension which is a bit of a weird pattern.

`$baseClass = DataObject::getSchema()->baseDataClass($modelClass);`

Or better, use Hierarchy's `getBaseHierarchyClass` method.

```suggestion
            'childrenMethod' => $baseClass::config()->get('default_children_method'),
```

GuySartorelli (changes requested):

code/Controllers/CMSMain.php
ChloeSartorelli: Please also add tests that validate this respects config from the correct place.

emteknetnz: I think this would just be `$modelClass::config()->get('default_children_method')` - because a subclass may override the config on base class?

I think we'd still want to have `private static $default_children_method` defined on Hierarchy otherwise we have config which isn't defined anywhere, which is a pretty confusing pattern

> Please also add tests that validate this respects config from the correct place.

Testing we're fetching the correct config? That's not something we normally test?  That also sounds like a pretty weird test - calling a method and then testing that inside that method `DataObject::singleton($modelClass)->prepopulateTreeDataCache()` was called with a specific array value in one of the params

ChloeSartorelli: Agree define the config there. Just doesn't need to use the `Configurable` trait.

> I think this would just be `$modelClass::config()->get('default_children_method')` - because a subclass may override the config on base class?

Right, but the queries aren't checking each class, they're querying based on the base class. So the base class is the only one that really matters. The PHPDoc for the config should make it clear that only the base class values will be respected, and then if someone sets it on the base class they're just not following what is documented and should expect failure.

emteknetnz: OK base class makes sense

Though we'll still need a keep the config on Hierarchy as there needs to be a default defined somewhere because we cannot assume that all classes that apply the Hierarchy extension will have the config defined

ChloeSartorelli: Again:

> Agree define the config there. Just doesn't need to use the `Configurable` trait.

You don't need that trait on an extension to provide default configuration

code/Controllers/CMSMain.php
```diff
@@ -490,17 +491,27 @@ public function AddForm(): CMSMainAddForm
...
+        DataObject::singleton($modelClass)->prepopulateTreeDataCache(null, $options);
         $html = $this->getTreeFor($modelClass);
         $this->extend('updateTreeAsUL', $html);
         return $html;
     }
 
+    /**
+     * Get the options used for the call to Hierarchy::prepopulateTreeDataCache()
+     */
+    private function getTreeAsULPrepopulateOptions(string $modelClass)
```
emteknetnz, line 503: Split off into its own method so can unit test the config far easier

GuySartorelli (approved):

PR #3080, opened 2025-05-16
TLN Update translations

GuySartorelli:
Automated translations update generated using [silverstripe/tx-translator](https://github.com/silverstripe/silverstripe-tx-translator)

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -1,4 +1,21 @@
 en:
+  SilverStripe\Admin\LeftAndMain:
+    ARCHIVED_RECORD: 'Archived {name} "{title}"'
+    AddNew: 'Add new {name}'
+    ArchiveWarning: 'Warning: This record will be unpublished before being sent to the archive.\n\nAre you sure you want to proceed?'
+    ArchiveWarningWithChildren: 'Warning: This record and all of its child records will be unpublished before being sent to the archive.\n\nAre you sure you want to proceed?'
+    CREATED_RECORD: 'Created {name} "{title}"'
+    DELETED_RECORD: 'Deleted {name} "{title}"'
+    DUPLICATED_RECORD: 'Duplicated {name} "{title}"'
+    DUPLICATED_RECORD_WITH_CHILDREN: 'Duplicated {name} "{title}" and children'
+    NEW_RECORD: 'New {recordtype}'
+    NewRecord: 'New {name}'
+    PUBLISHED_RECORD: 'Published {name} "{title}"'
+    RESTORED_RECORD: 'Restored {name} "{title}"'
+    ROLLEDBACK_PUBLISHED: 'Rolled back to published version.'
+    ROLLEDBACK_VERSION: 'Rolled back to version #{version}.'
+    SAVED_RECORD: 'Saved {name} "{title}"'
+    UNPUBLISHED_RECORD: 'Unpublished {name} "{title}"'
   SilverStripe\CMS\BatchActions\CMSBatchAction_Archive:
     RESULT: 'Deleted %d pages from draft and live, and sent them to the archive'
     TITLE: 'Unpublish and archive'
@@ -24,6 +41,8 @@ en:
     ArchiveWarningWithChildren: 'Warning: This page and all of its child pages will be unpublished before being sent to the archive.\n\nAre you sure you want to proceed?'
     ArchiveWarningWithChildrenAndCampaigns: 'Warning: This page and all of its child pages will be unpublished and automatically removed from their associated {NumCampaigns} before being sent to the archive.\n\nAre you sure you want to proceed?'
     CANT_REORGANISE: 'You do not have permission to alter Top level pages. Your change was not saved.'
+    CANT_REORGANISE2: 'You do not have permission to rearange the tree. Your change was not saved.'
+    CANT_REORGANISE_TOPLEVEL: 'You do not have permission to alter Top level records. Your change was not saved.'
     Cancel: Cancel
     ChoosePageParentMode: 'Choose where to create this page'
     ChoosePageType: 'Choose page type'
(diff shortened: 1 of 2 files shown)


lang/en.yml
```diff
@@ -75,7 +107,7 @@ en:
   SilverStripe\CMS\Controllers\CMSPageEditController:
     ErrorItemPermissionDenied: "It seems you don't have the necessary permissions to add {ObjectTitle} to a campaign"
     ErrorNotFound: "That {Type} couldn't be found"
-    MENUTITLE: 'Edit Page'
+    MENUTITLE: Pages
```
ChloeSartorelli, line 110: This new value is correct. I have confirmed that in CMS 5 this and the other `MENUTITLE` change don't affect anything (or at least if they do, I wasn't able to identify it)

lang/en.yml
```diff
@@ -93,7 +125,7 @@ en:
     VIEWINGVERSION: 'Currently viewing version {version}.'
     WHEN: When
   SilverStripe\CMS\Controllers\CMSPageSettingsController:
-    MENUTITLE: 'Edit Page'
+    MENUTITLE: Pages
```
emteknetnz, line 128: Is this supposed to change?

lang/en.yml
```diff
@@ -149,7 +187,7 @@ en:
     HelpChars: ' Special characters are automatically converted or removed.'
     OK: OK
   SilverStripe\CMS\Model\RedirectorPage:
-    CLASS_DESCRIPTION: 'Redirects to an internal page or an external URL'
+    CLASS_DESCRIPTION: 'Redirects requests to another location'
```
emteknetnz, line 190: Why is this changing?

lang/en.yml
```diff
@@ -164,10 +202,10 @@ en:
     REDIRECTTOPAGE: 'A page on your website'
     SINGULARNAME: 'Redirector Page'
     YOURPAGE: 'Page on your website'
-    db_ExternalURL: 'Other website URL'
-    db_RedirectionType: 'Redirect to'
-    has_one_LinkTo: 'Page on your website'
-    has_one_LinkToFile: 'File'
+    db_ExternalURL: 'External URL'
```
emteknetnz, line 205: Why are these changing?

lang/en.yml
```diff
@@ -341,16 +381,17 @@ en:
...
+    OverrideNote: 'Overrides inherited value from the source'
     PLURALNAME: 'Virtual Pages'
     PLURALS:
       one: 'A Virtual Page'
       other: '{count} Virtual Pages'
     PageTypNotAllowedOnRoot: 'Original page type "{type}" is not allowed on the root level for this virtual page'
     SINGULARNAME: 'Virtual Page'
-    db_CustomMetaDescription: 'Custom Meta Description'
-    db_CustomExtraMeta: 'Custom Meta Tags'
+    db_CustomExtraMeta: 'Custom extra meta'
```
emteknetnz, line 391: Why are these changing?

emteknetnz (changes requested):

lang/en.yml
ChloeSartorelli: Yes, same deal as https://github.com/silverstripe/silverstripe-cms/pull/3080#discussion_r2092124290

lang/en.yml
ChloeSartorelli: Beats me. The PHP had that value in CMS 5 as well, so the question really is "why wasn't this in the yaml before?"
The text collection is correctly setting this to the value in PHP-land.

If you want I can change the PHP to match this value, but either way the PHP and YAML need to match, and both have been at these values for quite a while from what I can see.

lang/en.yml
ChloeSartorelli: https://github.com/silverstripe/silverstripe-cms/pull/3080#discussion_r2094630679

lang/en.yml
ChloeSartorelli: https://github.com/silverstripe/silverstripe-cms/pull/3080#discussion_r2094630679

lang/en.yml
ChloeSartorelli: Will hold off updating the failing behat test pending your decision

emteknetnz: OK keep as is in this PR

lang/en.yml
emteknetnz: `db_ExternalURL` is changing because in CMS 5 it has `new TextField("ExternalURL", _t(__CLASS__.'.OTHERURL', "Other website URL"))` in `RedirectorPage::getCMSFields()` i.e. it's the default value in PHP

In CMS 6 we removed that line, presumably due to improvements in scaffolding

I'll assume all the other changes in this file are for similar reasons

What's kind of annoying is that transifex has no concept of CMS majors, we pretty much have to pick one here which should be CMS 6, so I think it's best to leave things as they are in the PR

lang/en.yml
emteknetnz: Leave as is per https://github.com/silverstripe/silverstripe-cms/pull/3080/files#r2094642942

lang/en.yml
emteknetnz: Leave as is per https://github.com/silverstripe/silverstripe-cms/pull/3080/files#r2094642942

emteknetnz (approved):

emteknetnz (changes requested):
Oh behat test needs fixed

lang/en.yml
ChloeSartorelli: > we pretty much have to pick one here which should be CMS 6

Why should it be the CMS 6 one though? I didn't notice that line in CMS 5 - I was just looking for `provideI18nEntities` and `fieldLabels` which are where these should ideally be set.

Probably when we removed that line, we didn't realise there would be this change as a result. If that's the case, it would mean that this is an unintended change that we should revert.

If you agree with that assessment I'll add the relevant code back into PHP-land to revert this unexpected change.
If you think we should stick with this PR I'm happy for that too, let me know either way.

I'll hold off updating behat tests pending your decision.

emteknetnz: I think we just need to let CMS 5 break here to some degree here (and it's a very small degree) otherwise we're going to have this basically redundant code sitting around in CMS 6 purely to retain the CMS 5 era source translation values, which we'll then need to retain in CMS 7 and beyond so as not to break CMS 6, etc

I guess this is just one of the unintended side-effects of having automatically generated translation keys

GuySartorelli:
Behat updated.
Still not 100% sure about this but we can always change back if someone complains.

emteknetnz (approved):

PR #3083, opened 2025-05-23
MNT Update test to suppress deleted version history

GuySartorelli:
Follows option 1 in https://github.com/silverstripe/silverstripe-versioned/pull/511#issuecomment-2902905039 - any discussion about whether that is the correct option should be held in that PR, not this one.

Needs that PR to be merged in order for CI to go green.

## Issue

- https://github.com/silverstripe/silverstripe-versioned-admin/issues/402

### DIFF
--- a/tests/php/Controllers/CMSMainTest.php
+++ b/tests/php/Controllers/CMSMainTest.php
@@ -611,7 +611,7 @@ class CMSMainTest extends FunctionalTest
         $page12 = $this->objFromFixture(SiteTree::class, 'page12');
         // Deleted
         $page1->doUnpublish();
-        $page1->delete();
+        $page1->suppressDeletedVersion(fn () => $page1->delete());
         // Live and draft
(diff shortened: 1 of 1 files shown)


tests/php/Controllers/CMSMainTest.php
```diff
@@ -611,7 +611,7 @@ public function testListViewForm(array $params, ?int $limit, int $expectedPreLim
         $page12 = $this->objFromFixture(SiteTree::class, 'page12');
         // Deleted
         $page1->doUnpublish();
-        $page1->delete();
+        $page1->suppressDeletedVersion(fn () => $page1->delete());
```
ChloeSartorelli, line 614: We should probably also add a properly archived page here, which is what I think `CMSSiteTreeFilter_StatusDeletedPages` is meant to show. Lemme know if you want that (I'd do it in `6.0` or `5.4` since it's just testing existing stuff, not directly related to the changes at hand)

GuySartorelli:
No longer needed

PR #3084, opened 2025-05-26
API Deprecate constructor which will be removed.

GuySartorelli:
Constructors with args should be properly deprecated with a warning because otherwise developers may continue to parse arguments in, expecting it them to be used.

## Issue

- https://github.com/silverstripe/.github/issues/409

### DIFF
--- a/code/Controllers/CMSSiteTreeFilter.php
+++ b/code/Controllers/CMSSiteTreeFilter.php
@@ -103,9 +103,13 @@ abstract class CMSSiteTreeFilter implements LeftAndMain_SearchFilter
         return $filterMap;
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


emteknetnz (changes requested):
Shouldn't we only be raising the deprecation message only if `$params !== null`?  Seems like if someone called this with no args they'd get the default `$params = null`, which is OK?

GuySartorelli:
Yeah I guess that's sensible. It's a weird scenario where it's debatable if the constructor method is deprecated, or if it's just the params that are deprecated since there's an implicit constructor.

Updated.
Please also merge the other PRs - they have nothing to do with this one and can be merged independently.

emteknetnz:
Forgot to push?

emteknetnz (approved):

Issue #3088, opened 2025-06-15
CMS 6 REGRESSION: Navigation labels with `&` displaying as `&amp;` on CMS page tree navigation


prij:
### Module version(s) affected

6

### Description

On Silverstripe 6, pages with '&' in their navigation labels are displaying with `&amp;` in the CMS Page sitetree.

![Image](https://github.com/user-attachments/assets/b4e860e4-2759-41d0-ba5e-4598d75df380)

### How to reproduce

Create a new page with '&' in the title - eg., 'Terms & Conditions'

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-cms/pull/3090

GuySartorelli:
Can reproduce, and confirmed this is a CMS 6 regression.

prij:
Just adding that this issue affects all special characters, not just &
- Apostrophes (') show as `&#039;`
- Quotes (") show as `&quot;`

GuySartorelli:
PR merged. This will be automatically tagged by GitHub Actions

PR #3090, opened 2025-06-22
FIX Do not double escape page titles

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/3088


code/Controllers/CMSMain.php
```diff
@@ -1668,9 +1678,11 @@ protected function getCreatableSubClasses(DataObject $record): array
         // Build the list of candidate children
         $cache = $this->getCreatableChildrenCache();
         $cacheKey = $this->generateChildrenCacheKey();
-        $children = $cache->get($cacheKey, []);
+        $children = [];
```
emteknetnz, line 1681: Semi-related fix required for unit-tests, getCreatableChildrenCache() can be null (`private ?CacheInterface $creatableChildrenCache = null;`), so we need to account for that in this method.

GuySartorelli (approved):
LGTM

PR #3092, opened 2025-07-01
MNT Run module-standardiser

emteknetnz:
This pull-request was created automatically by [module-standardiser](https://github.com/silverstripe/module-standardiser)


GuySartorelli (approved):
LGTM.
Note that both the scheduled and dispatched workflows run on the default branch, which then checks out the correct branch in the "Checkout code" step - so it's fine that this isn't in the CMS 5 branch.

Issue #3106, opened 2025-08-20
Performance improvements for `SiteTree`


GuySartorelli:
I've identified a couple of small performance improvements that can be made for `SiteTree`:

1. Use `getParent()` instead of `Parent()` since the former is cached
2. Avoid calculating the default url segment if we don't need to

## PRs

- https://github.com/silverstripe/silverstripe-cms/pull/3107

PR #3107, opened 2025-08-20
ENH Improve performance of SiteTree

GuySartorelli:
1. Use `getParent()` (which is cached) instead of `Parent()` (it's safe to do and likely to have already been cached by another usage of that method)
2. Avoid calculating default URL segment when it's not needed (saves ~10ms on page write though results seem to vary across attempts)

## Issues

- https://github.com/silverstripe/silverstripe-cms/issues/3106

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -2184,9 +2184,10 @@ class CMSMain extends LeftAndMain implements CurrentRecordIdentifier, Permission
         }
 
         if (($id = $this->urlParams['ID']) && is_numeric($id)) {
+            $modelClass = $this->getModelClass();
             /** @var DataObject&Hierarchy $record */
-            $record = DataObject::get($this->getModelClass())->byID($id);
-            if ($record && !$record->canCreate(null, ['Parent' => $record->Parent()])) {
+            $record = DataObject::get($modelClass)->byID($id);
+            if ($record && !$record->canCreate(null, ['Parent' => $record->getParent() ?? $modelClass::create()])) {
                 return Security::permissionFailure($this);
             }
             if (!$record || !$record->ID) {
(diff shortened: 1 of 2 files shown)


code/Model/SiteTree.php
```diff
@@ -1915,11 +1919,7 @@ public function getCMSFields()
             $urlsegment = SiteTreeURLSegmentField::create("URLSegment", $this->fieldLabel('URLSegment'))
                 ->setURLPrefix($baseLink)
                 ->setURLSuffix('?stage=Stage')
-                ->setDefaultURL($this->generateURLSegment(_t(
-                    'SilverStripe\\CMS\\Controllers\\CMSMain.NEWPAGE',
-                    'New {pagetype}',
-                    ['pagetype' => $this->i18n_singular_name()]
-                )))
+                ->setDefaultURL($this->getDefaultUrlSegment())
```
ChloeSartorelli, line 1923: This is just a small tidy-up since I had to make `getDefaultUrlSegment()` into a method to only call it when we need to in `onBeforeWrite()`.

code/Controllers/CMSMain.php
```diff
@@ -2184,9 +2184,10 @@ public function duplicate(HTTPRequest $request): HTTPResponse
         }
 
         if (($id = $this->urlParams['ID']) && is_numeric($id)) {
+            $modelClass = $this->getModelClass();
             /** @var DataObject&Hierarchy $record */
-            $record = DataObject::get($this->getModelClass())->byID($id);
-            if ($record && !$record->canCreate(null, ['Parent' => $record->Parent()])) {
+            $record = DataObject::get($modelClass)->byID($id);
+            if ($record && !$record->canCreate(null, ['Parent' => $record->getParent() ?? $modelClass::create()])) {
```
ChloeSartorelli, line 2190: The magic `Parent()` method returns an empty record if it failed to fetch one from the db, but `getParent()` returns `null` - so we have to account for that change.

emteknetnz (approved):

PR #3109, opened 2025-08-21
ENH Disable sorting on queries where sort order doesn't matter

GuySartorelli:
Sorting queries can be slow, especially on large datasets and especially if indexes aren't optimised. We should avoid sorting when the sort order doesn't matter.

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11833

### DIFF
--- a/code/Controllers/CMSMain.php
+++ b/code/Controllers/CMSMain.php
@@ -1466,8 +1466,8 @@ class CMSMain extends LeftAndMain implements CurrentRecordIdentifier, Permission
      */
     protected function collateDescendants($recordIDs, &$collator)
     {
-
-        $children = DataObject::get($this->getModelClass())->filter(['ParentID' => $recordIDs])->column();
+        // Disabling sort improves performance
+        $children = DataObject::get($this->getModelClass())->filter(['ParentID' => $recordIDs])->sort(null)->column('ID');
         if ($children) {
             foreach ($children as $item) {
                 $collator[] = $item;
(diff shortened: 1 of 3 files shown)


code/Model/SiteTree.php
```diff
@@ -1606,7 +1606,8 @@ protected function onBeforeDelete()
 
         // If deleting this page, delete all its children.
         if ($this->isInDB() && SiteTree::config()->get('enforce_strict_hierarchy')) {
-            foreach ($this->AllChildren() as $child) {
+            # Disabling sort improves performance
+            foreach ($this->AllChildren()->sort(null) as $child) {
                 $child->delete();
             }
```
ChloeSartorelli, line 1617: Doesn't matter what order they're deleted in

code/Controllers/CMSMain.php
```diff
@@ -1470,7 +1470,7 @@ protected function getArchiveWarningMessage(DataObject $record): string
     protected function collateDescendants($recordIDs, &$collator)
     {
 
-        $children = DataObject::get($this->getModelClass())->filter(['ParentID' => $recordIDs])->column();
+        $children = DataObject::get($this->getModelClass())->filter(['ParentID' => $recordIDs])->getIDList();
```
ChloeSartorelli: Order doesn't matter, we ultimately just use it for aggregate data.

emteknetnz (approved):

Issue #3100, opened 2025-08-11
Accessibility: Screen readers don't know the purpose of the `URLSegment` link


GuySartorelli:
When editing a page, there's a `<a>` link for the `URLSegment` which is actually a link to the draft page.

Especially for screen readers (but maybe for sighted users as well), it's not clear what this actually represents.

## Notes

- The `title` attribute is usually not read by screen readers unless there is no other text to read, so just adding `title` on its own isn't sufficient.
- Using `aria-described-by` will usually result in screen readers reading the full link text (in this case the URL) as well as the description text. This may be useful (the URL segment is available this way) but may also be hard to aurally parse.
- Using `aria-label` will result in screen readers only reading the label, not the link text. It may be appropriate to have something like "Link to draft page for URL segment $URLSegment" (where "$URLSegment" is the actual URL segment)

## Acceptance Criteria

- [ ] Add either a `aria-described-by` attribute (referencing the formfield label) or `aria-label` attribute (with some useful text) to the URL segment `<a>` tag.
- [ ] If appropriate, also add a `title` attribute with text similar to "link to draft page" for sighted users.

## PRs
- https://github.com/silverstripe/silverstripe-cms/pull/3110

GuySartorelli:
PR merged

PR #3110, opened 2025-08-26
ENH Include title and aria-label attrs on URL segment field link

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/3100

For the 'About us' url segment, will render as

```html
<a
  class="URL-link"
  href="https://mysite.test/about-us?stage=Stage"
  target="_blank"
  aria-label="View the draft page for URL segment about-us"
  title="View the draft page"
  >https://mysite.test/about-us</a>
```


client/src/legacy/SiteTreeURLSegmentField.js
```diff
@@ -34,7 +36,15 @@ $.entwine('ss', function($) {
...
 			// Transfer current value to holder
-			this.find('.URL-link').attr('href', encodeURI(url + field.data('suffix'))).text(previewUrl);
+			var $link = this.find('.URL-link');
+			$link.attr('href', encodeURI(url + field.data('suffix'))).text(previewUrl);
+			$link.attr('aria-label', i18n.inject(
+				i18n._t(
+					'SilverStripe\\CMS\\Forms\\SiteTreeURLSegmentField.ViewDraftFor',
+					'View the draft page for URL segment {segment}',
+				),
+				{ segment: encodeURI(val) }
```
ChloeSartorelli: I don't think we should be using `encodeURI` here. Currently I'm pretty sure it won't do anything (I don't think any encodable characters are even allowed) - but if we changed that, the screen reader would read out an encoded value but the actual value they get when editing is not encoded.

templates/SilverStripe/CMS/Forms/SiteTreeURLSegmentField.ss
```diff
@@ -1,7 +1,12 @@
 <div class="preview-holder">
-    <a class="URL-link" href="$URL" target="_blank">
+    <a class="URL-link"
+        href="$URL"
+        target="_blank"
+        aria-label="<%t SilverStripe\\CMS\\Forms\\SiteTreeURLSegmentField.ViewDraftFor 'View the draft page for URL segment {segment}' segment='' %>"
+        title="<%t SilverStripe\\CMS\\Forms\\SiteTreeURLSegmentField.ViewDraft 'View the draft page' %>">
```
ChloeSartorelli: Why do we have this both in the template _and_ in JavaScript? These localisation strings could get out of sync.

GuySartorelli (changes requested):

client/src/legacy/SiteTreeURLSegmentField.js
emteknetnz: Updated

templates/SilverStripe/CMS/Forms/SiteTreeURLSegmentField.ss
emteknetnz: Removed aria-label from template, kept title in template

ChloeSartorelli: Why have them in two separate places? I don't understand why they wouldn't either both be in the template, or both be in the JS. There's probably just something I'm missing about how this field works?

emteknetnz: The title doesn't dynamically change, aria-label does

ChloeSartorelli: Fair enough - though please add a comment to the template that says an aria-label attribute is added through JS. Otherwise it may be tempting when looking at the template in the future to think there's no aria-label, and just adding one which may not match the JS one.

emteknetnz: Updated

GuySartorelli (approved):
LGTM

PR #3120, opened 2025-10-09
ENH Add aria attrs to site tree

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2000

Requires https://github.com/silverstripe/silverstripe-framework/pull/11863 for tests to go green


lang/en.yml
```diff
@@ -88,6 +88,8 @@ en:
     SAVEDRAFT: Save
     SEARCHRESULTS: 'Search results'
     SHOW_AS_LIST: 'show as list'
+    TOGGLE_CHILD_PAGES: 'Toggle child pages of {title}'
+    TOGGLE_CHILD_PAGES_ROOT: 'Toggle child pages of top level'
```
ChloeSartorelli:
```suggestion
    TOGGLE_CHILD_PAGES_ROOT: 'Toggle site tree'
```
This probably makes more sense. The "child pages of top level" are really just the root level pages in the site tree - so you're toggling the whole tree.

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_SubTree.ss
```diff
@@ -1,15 +1,32 @@
...
+    <ul id="subtree-0" role="group">
+        <li id="record-0"
+            data-id="0"
+            class="Root nodelete"
+            role="treeitem"
+            aria-level="1"
+            <%-- aria-expanded value is handled via JS --%>
+            aria-expanded="false"
+        ><ins class="jstree-icon font-icon-right-dir"
+              role="button"
```
ChloeSartorelli: We should just use a semantic `button` element instead of an [inserted text element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ins) which is just completely symantically incorrect.

While screen readers _should_ use the `role` attribute, some won't (unfortunately orca which I'm using to test is one that won't) so we should use semantically correct HTML elements wherever possible.

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_TreeView.ss
```diff
@@ -10,6 +10,7 @@ $ExtraTreeTools
 		</a>
 
 		<div class="cms-tree <% if $TreeIsFiltered %>filtered-list<% end_if %>"
+			role="tree"
```
ChloeSartorelli: I think the `tree` role should be on the first `ul` (i.e. where the list starts). This is how they do it in the [W3C example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) and is also consistent with what elements screen readers usually pay attention to - they don't announce a `div` usually but they do usually announce a `ul` as being a list.

By adding the `tree` role to the `ul` element you're saying "This isn't a normal list, it's a tree". Without the role it's just "a list inside a tree"

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_TreeNode.ss
```diff
@@ -1,5 +1,31 @@
-<li id="record-{$node.ID}" data-id="{$node.ID}" data-recordtype="{$node.ClassName}" class="$markingClasses $extraClass"><ins class="jstree-icon font-icon-right-dir" tabindex="0">&nbsp;</ins>
-    <a href="{$Controller.LinkRecordEdit($node.ID).ATT}" title="{$Title.ATT}"><ins class="jstree-icon font-icon-drag-handle">&nbsp;</ins>
+<li id="record-{$node.ID}"
+    data-id="{$node.ID}"
+    data-recordtype="{$node.ClassName}"
+    class="$markingClasses $extraClass"
+    role="treeitem"
```
ChloeSartorelli: In [the W3C example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) the `<li>` elements have `role="none"` and the `<a>` elements have `role="treeitem"`. We should probably follow that example unless there's a good reason not to.

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_SubTree.ss
ChloeSartorelli: That said - in [the W3C example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) the arrows aren't actually separately tabbable - instead the treeitem intercepts key presses for left/right arrows for expanding/collapsing, and the arrow is just a `span` with a click handler for mouse interaction.

This is more consistent with the keyboard interaction recommendations W3C make (it makes no mention of tab or enter for expanding/collapsing child nodes). Perhaps we should follow their example here as well?

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_SubTree.ss
```diff
@@ -1,15 +1,32 @@
 <% if not $node.IsInDB %><%-- Only render root node if it's the true root --%>
-    <ul><li id="record-0" data-id="0" class="Root nodelete"><ins class="jstree-icon font-icon-right-dir" tabindex="0">&nbsp;</ins><strong>$rootTitle</strong>
+    <ul id="subtree-0" role="group">
```
ChloeSartorelli: In [the W3C example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) the `ul[role="group"]` elements also have an aria-label which matches the parent node's name. Unsure what the ultimate effect of this is so unsure if that's important or not.

GuySartorelli (changes requested):
There are various recommendations in [the W3C example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) which probably apply to our usage, since that example is for a tree used as navigation which is what we're doing here.

I recommend we follow that example, including add a `nav` parent to the tree list (instead of the current `div`), and additional recommendations such as using the `aria-current` attribute etc.
Note that not everything in the live example has a corresponding textual recommendation - but IMO we should treat the example's aria attributes as best practice anyway. I've marked some (but not all) of the differences between their example and this implementation in separate PR comments.

We can however ignore some of the recommendations from that example that are for regular websites, such as the use of the `banner` landmark.

I will note that Orca (the screen reader I'm using for testing) works really well both in terms of correctly announcing tree items and the tree itself _and_ in terms of respecting correct key bindings with the W3C example than with our site tree. In our one orca didn't seem to recognise it was in a tree at all, and wouldn't respect the left/right arrow keybindings.

lang/en.yml
emteknetnz: Updated

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_SubTree.ss
emteknetnz: I'm not going to change HTML tags (will affect styling) or tab order (will affect behat tests) in this PR - I've created a [follow up issue](https://github.com/silverstripe/silverstripe-cms/issues/3122) for this

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_SubTree.ss
emteknetnz: This is the root most `<ul>`, I've set this to role="tree" using JS

Later on in this file I've updated non-root `<ul>` nodes so that they have an aria-label that is the same as the parent page name

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_TreeNode.ss
emteknetnz: Updated

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_TreeView.ss
emteknetnz: Updated, had to do this via JS otherwise jstree will just override whatever we set on the root `<ul>`

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_SubTree.ss
ChloeSartorelli: It feels like you might have not seen the second comment there, about making these not tabbable and not using the button role? If we do that, it doesn't really matter what the HTML tag is because assistive technologies will ignore them anyway - they'll only be for mouse interaction for sighted users.

GuySartorelli (changes requested):
The experience with a screen reader still isn't the same or as good as the W3C example. Are you using a screen reader to test this?
One clear example is when I use left/right arrow keys to expand/collapse a node, with the W3C example the screen reader announces "collapsed" or "expanded" as appropriate. With this PR I get no feedback at all.

I again recommend we follow the W3C example as closely as we can.

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_SubTree.ss
emteknetnz: As previously mentioned, changing either the html tags or the tab order will affect styling / css and behat and I do not want to that in this PR as this one is focused on aria attrs. I have created a follow up issue to implement those.

emteknetnz:
Have moved all the aria stuff off the `<li>` on to the `<a>`, seems to be closer to the w3c example now

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_SubTree.ss
ChloeSartorelli: This is nearly the only problem now with what the screen reader is saying. Whenever I tab to one of these button-not-buttons, it _leaves_ the context of the tree. The screen reader changes to "browse mode". When I tab back off it, it _enters_ the context of the tree again, and I get told all the tree information (level 2, etc).

What interactive elements live where very much impacts the screen reader experience - it's hard to say "yes this is done the screen reader experience is good now" when this interactive element is interfering with that experience. These two things inform one another.

I guess I have to call it done and then when [the follow up issue](https://github.com/silverstripe/silverstripe-cms/issues/3122) is implemented, I have to check it again and see how it impacts the screen reader experience? And then if more changes are needed to the aria attributes, we'll open _another_ separate card to address it separately again?

GuySartorelli (changes requested):
There's a merge conflict.

The only other thing remaining that bothers me (other than the `<ins>` stuff) is the level starts at level 2? Feels like the root level pages should all be level 1.

client/src/legacy/CMSMain.Tree.js
```diff
@@ -183,6 +183,14 @@ $.entwine('ss.tree', function($) {
     }
   });
 
+  // Add role="tree" to the root level <ul> for accessibility
+  // This needs to be done with JS as jstree will override any attributes set server-side
+  $('.cms-tree > ul').entwine({
+    onmatch: function() {
+      this.attr('role', 'tree');
+    }
+  });
```
ChloeSartorelli, line 194: We should probably use one of the jstree events, such as `loaded.jstree` (similar to other events already in https://github.com/silverstripe/silverstripe-admin/pull/2016)

composer.json
```diff
@@ -21,7 +21,7 @@
     "require": {
         "php": "^8.3",
         "silverstripe/admin": "^3.2",
-        "silverstripe/framework": "^6.2",
+        "silverstripe/framework": "^6.3",
```
emteknetnz, line 24: Requires the change in [src/ORM/Hierarchy/MarkedSet.php](https://github.com/silverstripe/silverstripe-framework/pull/11863/files#diff-f394bad8974650c0f4ffcc769b47c9d6a22f630d526be20daf4ac7363b02f0e9)

PR #3126, opened 2025-10-23
FIX Use application/json for Content-Type

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11866

### DIFF
--- a/code/BatchActions/CMSBatchAction_Archive.php
+++ b/code/BatchActions/CMSBatchAction_Archive.php
@@ -23,7 +23,8 @@ class CMSBatchAction_Archive extends CMSBatchAction
             $pages,
             'doArchive',
             _t(__CLASS__ . '.RESULT', 'Deleted %d pages from draft and live, and sent them to the archive')
-        );
+        )
+            ->addHeader('Content-Type', 'application/json');
     }
 
     public function applicablePages($ids)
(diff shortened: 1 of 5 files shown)


code/BatchActions/CMSBatchAction_Archive.php
```diff
@@ -23,7 +23,8 @@ public function run(SS_List $pages): HTTPResponse
             $pages,
             'doArchive',
             _t(__CLASS__ . '.RESULT', 'Deleted %d pages from draft and live, and sent them to the archive')
-        );
+        )
+            ->addHeader('Content-Type', 'application/json');
```
emteknetnz, line 27: This is essentially re-adding the same header added in the [admin PR](https://github.com/silverstripe/silverstripe-admin/pull/2031/files#diff-d9091c39d47fdfe4f05392bdd1ada56cc04b05009a1c7005c8a0215abb33f023R72) - though we need to add it here also otherwise it'll fail the --prefer-lowest build

This will override the existing header, i.e. there will not be duplicate 'Content-Type' headers sent

GuySartorelli (commented):
I'd prefer not to add redundant code like this - can we bump the dep to the specific patch instead? Or put the test in admin where the code change was?

emteknetnz:
> I'd prefer not to add redundant code like this - can we bump the dep to the specific patch instead?

In silverstripe/cms, composer.json has a constraint of `"silverstripe/admin": "^2.2",` so we cannot bump that to 2.4.x without a minor release, which we aren't going to do for CMS 5

> Or put the test in admin where the code change was?

I could make a TestOnly batch action and test that if you want and then just close this PR if you want?

GuySartorelli:
> I could make a TestOnly batch action and test that if you want and then just close this PR if you want?

Yes please. That resolves the redundant code problem as well as giving us faster feedback if we do accidentally break batch actions through changing something in admin.

Issue #3132, opened 2025-11-21
i18n: .lowercase modifier breaks German and Luxembourgish noun capitalization in "Add new {name}" button


lerni:
### Module version(s) affected

CMS: 6.x

### Description

The "Add new page" button and similar buttons use `.lowercase` (LowerCase from DBString) on singular name, which breaks German and Luxembourgish grammar rules where nouns must be capitalized. 

**Affected templates:**
- `cms/templates/SilverStripe/CMS/Controllers/Includes/CMSMain_ContentToolActions.ss`
- `cms/templates/SilverStripe/CMS/Controllers/Includes/CMSMain_ListView.ss`

**Current implementation:**
```ss
<%t SilverStripe\Admin\LeftAndMain.AddNew 'Add new {name}' name=$getRecord('singleton').i18n_singular_name().lowercase %>
```

### How to reproduce

Install Silverstripe CMS 6.x
Switch to German language (de_DE)
Navigate to the CMS pages section
Observe the "Neue seite hinzufügen" button (incorrect lowercase "seite")

### Possible Solution

May `de_*` & `lb_LU` just need to handle translation separately, but on the other hand assuming lowercase is a general solution is probably flawed.

### Additional Context

Current (incorrect):
<img width="303" height="136" alt="Image" src="https://github.com/user-attachments/assets/2023ff17-c24f-48cd-9b67-5f185d903677" />

Expected:
"Neue Seite hinzufügen" (with capital "S" in "Seite")

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## Kitchen Sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/19653616727

## PRs
- https://github.com/silverstripe/silverstripe-cms/pull/3133

GuySartorelli:
PR merged

lerni:
Impressive turnaround time - thank you!

PR #3133, opened 2025-11-25
FIX Do not lowercase singular name

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/3132

Reason this bug occurs is because [en.yml](https://github.com/silverstripe/silverstripe-cms/blob/6/lang/en.yml#L4) has ` AddNew: 'Add new {name}'`

While  [de.yml](https://github.com/silverstripe/silverstripe-cms/blob/6/lang/de.yml#L4) has `AddNew: '{name} neu hinzufügen'`

So the assumption that we should lowercase the singular name because it's not the first word is a bad assumption.

We don't lowercase the singular name on [GridFieldAddNewButton](https://github.com/silverstripe/silverstripe-framework/blob/6/src/Forms/GridField/GridFieldAddNewButton.php#L65) - so it makes sense that we shouldn't lowercase in these 2x templates

I've also had a look around for any other examples like these where we use the `AddNew` string - I couldn't find any others where we are lowercasing `{name}`


GuySartorelli (approved):
LGTM. Displays like this in English:

image

Discussed offline if we should capitalise the N in New ("Add New Page"), or if we should lowercase everything except for the first letter ("Add new page" - or "Seite neu hinzufügen" for German) but opted to leave that alone for now.

PR #3134, opened 2025-11-26
MNT Update unit tests for PHP 8.5

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/455

The following changes were made based on the [PHP 8.5 changelog](https://www.php.net/ChangeLog-8.php):

- The setAccessible() methods of various Reflection objects have been deprecated, as those no longer have an effect.


GuySartorelli (approved):
LGTM

Issue #3135, opened 2025-12-02
Accessibility: Make site tree a tab island


emteknetnz:
Currently every element in the site tree is a tab element. This should be updated to make the whole site tree a tab-island and then use arrow keys to navigate within the site tree.

Doing this will make it much easier to tab to other parts of the CMS

Note that updating the html elements is probably out of scope - [this issue exists](https://github.com/silverstripe/silverstripe-cms/issues/3122) to update the HTML, ~though I'm not convinced we should while this is still a non-react component~ (tree will remain non-react until at least CMS 7).

## Acceptance criteria
- [ ]  The site tree has a single tab-able element
- [ ]  Arrow keys can be used to navigate through the site tree
- [ ]  site tree uses an appropriate [HTML landmark](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/) (e.g. `nav`)
- [ ]  Behat tests pass

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/21654732767
^ cms and linkfield behat failures are [existing](https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/21577954127)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2068
- https://github.com/silverstripe/silverstripe-cms/pull/3136

GuySartorelli:
PR(s) merged. This will be included in the next minor release.

Issue #3122, opened 2025-10-14
Accessibility: Update site tree html


emteknetnz:
Follow on from https://github.com/silverstripe/silverstripe-cms/pull/3120/files#r2427353622

The expander icons in the site tree should change from `<ins role="button">` to either `<button>` (to keep them tabbable) or `<span>` (to make them not tabbable)

## Acceptance criteria
- [ ] `<ins>` is no longer used for expander icons
- [ ] The left/right arrows still collapse/expand nested pages
- [ ] Behat tests in silverstripe/admin still pass
- [ ] Changelog is updated to note the change in HTML

## Kitchen sink CI
https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/21934273548
(asset-admin and linkfield issues are [existing](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/21846601524))

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2111
- https://github.com/silverstripe/silverstripe-cms/pull/3156
- https://github.com/silverstripe/developer-docs/pull/874

emteknetnz:
There's some ancient and fragile JS in jquery.jstree.js in silverstripe/admin (and elsewhere?) that expects there to be `ins` elements.

I'm very inclined to just leave the site tree semantically wrong for now and do things properly later with with a rebuild in react.

I've created https://github.com/silverstripe/silverstripe-cms/issues/3135 for now to update the html to make the site tree a tab island

emteknetnz:
~I've attempted this though I'm finding it too hard so giving up. The issue is that jstree uses the `<ins>` tags for its logic, and attempting to swap them out to `<span>` is fragile and is breaking random things for me e.g. adding pages using the right hand context menu is now adding them to the root instead of as a child page.~

Update: this is actually an existing bug - https://github.com/silverstripe/silverstripe-cms/issues/3154

GuySartorelli:
@emteknetnz Before I review this, can you please tell me if this is the stage at which the screen reader experience should be correct? Or is there another step after this which will touch the screen reader experience?

emteknetnz:
Screen reader has not been implemented in this issue - it will be done in this [issue instead](https://github.com/silverstripe/silverstripe-admin/issues/2000)

GuySartorelli:
PR(s) merged. This will be included in the next minor release.

Issue #3115, opened 2025-10-01
Accessibility: Move pages with keyboard


emteknetnz:
Split from https://github.com/silverstripe/silverstripe-cms/issues/3101

You can drag and drop pages with the mouse, but there is no equivalent to move pages with just the keyboard

Because pages can only move up or down by one page sort per keystroke, it's important to not update the database until after the page has reached its target destination, others there will be a bunch of database queries and version created that don't need to exist.

I first attempt this on https://github.com/silverstripe/silverstripe-admin/pull/2001, however I was able to get pages to move between siblings, I was simply unable to pages to move between nesting levels i.e. so that the current parent then becomes a sibling. Code I attempted to use is below where I tried to bind Ctrl-Up to move the page up.

Note, should probably be bound to space to activate a move mode, as elemental uses space to move elements

```js
          .on('keydown', function(e) {
            // ..
            } else if (key === 'ArrowUp') {
              // if ctrl or command (iOS) modifier pressed, move the node up
              if (e.ctrlKey || e.metaKey) {
                var $prev = $li.prev('li');
                if ($prev.length) {
                  $(this).jstree('move_node', $li, $prev, 'before');
                } else {
                  // if no previous sibling, move before parent if possible
                  // >> This simply does not work, I have no idea why
                  // >> I've spent a long time debugging it and trying different variations and nothing seems to work
                  var $parent = $li.closest('ul').closest('li');
                  if ($parent.length) {
                    $(this).jstree('move_node', $li, $parent, 'before');
                  }
                }
              } else {
                // otherwise move focus up
                // ...
```

## Related
- https://github.com/silverstripe/silverstripe-cms/issues/3101

## Acceptance criteria
- [ ] Pages can be moved within the site tree using just the keyboard
- [ ] Page moves are not persisted to the database until they have reached their final destination

PR #3136, opened 2025-12-02
ENH Make site tree a tab-island

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/3135

Needs https://github.com/silverstripe/silverstripe-admin/pull/2068 merged to pass


GuySartorelli (changes requested):
Works locally, but:

1. Please use an appropriate HTML landmark element as the parent of the island (most likely `nav`)
2. If we're following wcag patterns, when we tab to the tree and a page is open, we should be at that page in the tree (see https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/)

emteknetnz:
> If we're following wcag patterns, when we tab to the tree and a page is open, we should be at that page in the tree 

The underlying library (jstree) we're using is veeery old, brittle, and makes development hard. This recommendation would involve adding state management to the site tree

If we wish to do this, at this point I'm thinking we're best off pausing accessibility on the site tree specifically and just rebuilding the site tree in react, which needs to happen at some point.

Thoughts on that?

GuySartorelli:
The state should already be managed with a `aria-current="page"` attribute, assuming the state you mean is knowing which item to focus on when tabbing to the tree.

emteknetnz:
aria-current is being used on the tree right now

There's still a lot of work to do to get the site tree accessible
- dynamic tab index (this card)
- [aria attributes](https://github.com/silverstripe/silverstripe-admin/issues/2000)
- [use sematic html](https://github.com/silverstripe/silverstripe-cms/issues/3122)
- [move pages with keyboard](https://github.com/silverstripe/silverstripe-cms/issues/3115) < this is complex

Given the work remaining, I think we should just rebuild the site tree in react at this point

GuySartorelli:
I think that may be more difficult than you expect - the markup is at least partially done through explicit PHP methods which expect to return markup that updates the tree. Anyone customising those methods should expect their customisations to affect the tree markup.

Maybe lets discuss offline and report back here with a decision.

emteknetnz:
Chatted offline, agreed that while it's not ideal, it's acceptable from a backwards compatibility perspective if rebuild in react and make that default, and we retain the old API and functionality which can be used instead of the new react version via config.

code/Controllers/CMSMain.php
```diff
@@ -144,6 +144,7 @@ class CMSMain extends LeftAndMain implements CurrentRecordIdentifier, Permission
     private static array $url_handlers = [
         'EditForm/$ID' => 'EditForm',
         'GET SearchForm' => 'getSearchForm',
+        'treeview/$ID' => 'treeview',
```
emteknetnz, line 147: The new `$ID` param is optional, so this isn't really an API break.

client/src/legacy/CMSMain.js
```diff
@@ -55,6 +55,15 @@ $.entwine('ss', function ($) {
...
 
       this.data('deferredNoCache', (clearFiltered || viewType === VIEW_TYPE_LIST));
+      
+      // Append current page ID to treeview URL if in page edit mode
+      if (viewType === VIEW_TYPE_TREE) {
+        const pageID = $('.cms-edit-form input[name=ID]').val();
+        if (pageID) {
+          url = `${url}/${pageID}`;
+        }
+      }
```
ChloeSartorelli, line 65: What does this do? When does this happen?

code/Controllers/CMSMain.php
```diff
@@ -579,9 +580,25 @@ public function getTreeFor(
      */
     protected function getTreeNodeCustomisations()
     {
-        return function (DataObject $node) {
+        // $isFirstPageAssigned acts like a non-static private class variable
+        // because we're returning a callable that captures it by reference
```
ChloeSartorelli: This comment is more confusing than just not having the comment at all IMO :sweat_smile: 

If we want a class variable/property we should add one. If we don't, then referencing that specific API type for a variable that isn't that is confusing.

If the intention is just to explain how `&` works in PHP, we typically don't add code comments to explain PHP syntax.

```suggestion
```

code/Controllers/CMSMain.php
```diff
@@ -579,9 +580,25 @@ public function getTreeFor(
...
+            if (!$isFirstPageAssigned) {
+                $isFirstPage = true;
+                $isFirstPageAssigned = true;
+            }
+
             return [
                 'Controller' => $this,
+                'isCurrentPage' => $isCurrentPage,
+                'isFirstPage' => $isFirstPage,
+                'hasCurrentPage' => $hasCurrentPage,
```
ChloeSartorelli, line 599: Why are we assigning `$hasCurrentPage` outside the callback, but other things like `isCurrentPage` inside the callback?

Seems we should just assign it inside the callback. We don't need intermediate variables, either.

```suggestion
        $currentRecordID = $this->currentRecordID();

        return function (DataObject $node) use ($currentRecordID, &$isFirstPageAssigned) {
            $isFirstPage = false;
            if (!$isFirstPageAssigned) {
                $isFirstPage = true;
                $isFirstPageAssigned = true;
            }

            return [
                'Controller' => $this,
                'isCurrentPage' => $node->ID === $currentRecordID,
                'isFirstPage' => $isFirstPage,
                'hasCurrentPage' => $currentRecordID !== null,
```

I'd probably assign `$hasCurrentPage` inside the callback too (anonymous functions are automatically bound to the object that creates them, so `$currentRecordID = $this->currentRecordID()` will work in the callback) but I'll leave this to you to decide, since I don't know I don't know if the old or new value should be used if the current record ID changes between the time the callback is created and called.

code/Controllers/CMSMain.php
```diff
@@ -1486,6 +1503,10 @@ protected function collateDescendants($recordIDs, &$collator)
      */
     public function treeview()
     {
+        $id = $this->getRequest()->param('ID');
+        if ($id) {
+            $this->setCurrentRecordID($id);
+        }
```
ChloeSartorelli, line 1507: Why is this change being made? Was there a bug this is fixing? Seems like it's probably a change that goes beyond just setting up roving tabindex stuff

GuySartorelli (changes requested):
I've only given the code a cursory glance for now

code/Controllers/CMSMain.php
emteknetnz: It's not fixing a bug, this is to support the new functionality

`treeview` is called via an ajax request so the server doesn't know which the current page is. 

The current page is used in CMSMain_TreeNode.ss to set tabindex="0" and aria-current="page"

client/src/legacy/CMSMain.js
emteknetnz: It's to send the current pageID to the treeview endpoint - see https://github.com/silverstripe/silverstripe-cms/pull/3136/changes#r2739477617

code/Controllers/CMSMain.php
emteknetnz: Removed

code/Controllers/CMSMain.php
emteknetnz: I prefer it the way it is, it's assigning the stuff that's not dependent on `$node` outside the function,

GuySartorelli (approved):
LGTM. Will merge after other PR(s)

GuySartorelli:
(oop, needs rebase. Didn't see that)

emteknetnz:
Rebased

GuySartorelli (approved):
LGTM. Works as expected locally.

PR #3137, opened 2025-12-11
NEW React site tree

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2070

Experimental, do not merge


tests/php/Controllers/CMSSiteTreeControllerTest.php
```diff
@@ -0,0 +1,245 @@
+<?php
+
+namespace SilverStripe\CMS\Tests\Controllers;
```
emteknetnz, line 3: This unit test file is in the wrong module - it should be in admin instead since it's testing an admin component (ComplexTreeView)

GuySartorelli:
Closing as we chose not to proceed with this for now

PR #3145, opened 2026-01-28
FIX Respect value of PageType optionsetfield in AddForm

GuySartorelli:
There's an extension hook that allows changing the default page type for new pages, though the value in the form field is not respected. This change ensures that the internal "selected" state in the JavaScript logic gets initially set to whatever value is set in the form field.

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11944


client/src/legacy/CMSMain.AddForm.js
```diff
@@ -150,11 +153,11 @@ $.entwine('ss', function($){
 
       // Keep current selection if we can, or set default child selection, or fall back to first available option
       if (keepSelection) {
-        var selectedEl = $(currentSelection).parents('li:first');
+        var selectedEl = $(currentSelection);
       } else if (defaultChildClass) {
         var selectedEl = this
           .find('#Form_AddForm_PageType div.radio input[value=' + defaultChildClass + ']')
-          .parents('li:first');
+          .parents('div.radio');
```
ChloeSartorelli, line 160: There's no `li` in the markup (maybe there was in ancient times) - the item that gets the `selected` class and is therefore the relevant element is the `div`.

emteknetnz (approved):

PR #3149, opened 2026-02-02
FIX Show validation errors for add record form.

GuySartorelli:
This ensures validation errors are displayed in the form for both regular form submission as well as the context menu's form submission.

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11943

### DIFF
--- a/code/Controllers/CMSPageAddController.php
+++ b/code/Controllers/CMSPageAddController.php
@@ -184,14 +184,17 @@ class CMSPageAddController extends CMSPageEditController
         )->setHTMLID('Form_AddForm')->setStrictFormMethodCheck(false);
         $form->setAttribute('data-hints', $this->SiteTreeHints());
         $form->setAttribute('data-childfilter', $this->Link('childfilter'));
+        $form->setAttribute('data-pjax-fragment', 'CurrentForm');
         $form->setValidationResponseCallback(function (ValidationResult $errors) use ($negotiator, $form) {
             $request = $this->getRequest();
             if ($request->isAjax() && $negotiator) {
-                $result = $form->forTemplate();
                 return $negotiator->respond($request, [
-                    'CurrentForm' => function () use ($result) {
-                        return $result;
-                    }
+                    'Content' => function () use ($form) {
+                        return $this->renderWith($this->getTemplatesWithSuffix('_Content'), ['AddForm' => $form]);
+                    },
+                    'CurrentForm' => function () use ($form) {
+                        return $form->forTemplate();
+                    },
                 ]);
             }
             return null;
--- a/tests/behat/features/create-a-page.feature
+++ b/tests/behat/features/create-a-page.feature
@@ -61,3 +61,33 @@ Feature: Create a page
     And I press the "Add new" button
     Then I should see "Virtual Page" in the "#Form_AddForm_PageType div.radio.selected" element
     Then the "Virtual Page" checkbox should be checked
+
+  Scenario: I can create a page using the context menu
+    Given I am logged in as a member of "EDITOR" group
+    When I go to "/admin/pages"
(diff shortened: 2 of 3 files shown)


code/Controllers/CMSPageAddController.php
```diff
@@ -184,14 +184,17 @@ public function AddForm()
         )->setHTMLID('Form_AddForm')->setStrictFormMethodCheck(false);
         $form->setAttribute('data-hints', $this->SiteTreeHints());
         $form->setAttribute('data-childfilter', $this->Link('childfilter'));
+        $form->setAttribute('data-pjax-fragment', 'CurrentForm');
```
ChloeSartorelli, line 187: This change ensures validation errors display correctly on normal form submission (i.e. when you load the form and submit it manually)

code/Controllers/CMSPageAddController.php
```diff
@@ -184,14 +184,17 @@ public function AddForm()
...
             $request = $this->getRequest();
             if ($request->isAjax() && $negotiator) {
-                $result = $form->forTemplate();
                 return $negotiator->respond($request, [
-                    'CurrentForm' => function () use ($result) {
-                        return $result;
-                    }
+                    'Content' => function () use ($form) {
+                        return $this->renderWith($this->getTemplatesWithSuffix('_Content'), ['AddForm' => $form]);
+                    },
```
ChloeSartorelli, line 194: Adding this ensures validation errors are displayed when the ajax form submission from the context menu fails validation.

This is required because the entire controller template gets rendered, and usually that results in a whole new form being instantiated which _hasn't_ been validated against. By passing it the form that explicitly already failed validation, we get the messages rendered.

Note that this only happens on validation error because it's passed into setValidationResponseCallback(). Loading the form without submitting it won't go through this path.

code/Controllers/CMSPageAddController.php
```diff
@@ -184,14 +184,17 @@ public function AddForm()
         )->setHTMLID('Form_AddForm')->setStrictFormMethodCheck(false);
         $form->setAttribute('data-hints', $this->SiteTreeHints());
         $form->setAttribute('data-childfilter', $this->Link('childfilter'));
+        $form->setAttribute('data-pjax-fragment', 'CurrentForm');
         $form->setValidationResponseCallback(function (ValidationResult $errors) use ($negotiator, $form) {
             $request = $this->getRequest();
             if ($request->isAjax() && $negotiator) {
-                $result = $form->forTemplate();
```
ChloeSartorelli, line 190: Moved this to be inside the 'CurrentForm' pjax callback since we don't need it outside of that and rendering is expensive.

tests/behat/features/create-a-page.feature
```diff
@@ -61,3 +61,33 @@ Feature: Create a page
...
     Then the "Virtual Page" checkbox should be checked
+
+  Scenario: I can create a page using the context menu
+    Given I am logged in as a member of "EDITOR" group
+    When I go to "/admin/pages"
+		And I right click on "MyPage" in the tree
+		And I hover on "Add new page here" in the context menu
+		And I click on "Page" in the context menu
+    Then I should see "New Page" in the tree
+    And I should see an edit page form
```
ChloeSartorelli, line 72: This is technically not in scope but we didn't have any tests for adding a page with the context menu. I figured it makes sense to add while I'm in here. This also acts as a control against false negatives if the below scenario fails.

ChloeSartorelli: (oops, and fixed tabs to spaces after making that comment)

emteknetnz (approved):

mooror:
Looks great. I like how you break down a number of the changes you made. Nicely done :+1:

Issue #3154, opened 2026-02-05
CMSMain_TreeView right click "Add new page here" error


jackkemmish:
### Module version(s) affected

6

### Description

Hi there, I've just spotted a potential bug since this change to the `CMSMain_TreeView.ss` file was made
https://github.com/silverstripe/silverstripe-cms/commit/28612d0b3a30e58365961599e18f8a4ba92e90f1#diff-ed696030b6f3af3c09efb5d0a64649c6c9a7b48d9bd52fe2aea80bae04231c85

When right clicking on a page in the CMS tree and selecting a page within "Add new page here", clicking an option shows this error in dev tools:
```
[Unknown Error]
GET /admin/pages/edit/AddForm/?action_doAdd=1&SecurityID=1e9b966df05523c937ff2887da2ea8ed45bea906

Line in
Trace
```

I have manually tried reverting back to how it was and you can successfully trigger a page to be added.

Not sure if this helps, but the value for the `data-url-addpage` data attribute is different when inspecting between `$Link` and `$LinkRecordAdd` as you can see below. There are two places this data attribute can be found in the code when inspecting:
```
Old version (`$LinkRecordAdd`)
data-url-addpage="admin/pages/edit/add/?ParentID=%s"
data-url-addpage="admin/pages/edit/AddForm/?action_doAdd=1"

New version (`$Link`)
data-url-addpage="admin/pages/edit/add/?ParentID=%s"
data-url-addpage="admin/pages/edit/add/AddForm/?action_doAdd=1&ParentID=%s&RecordType=%s"
```

Look forward to your response.

Thanks,
Jack

### How to reproduce

1. Go into the CMS
2. Right click on a page and click on a page underneath "Add new page here"
3. The page will attempt to load, then an `error` toast appears.

### Possible Solution

Revert back to the `$LinkRecordAdd` function?

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-cms/pull/3158

emteknetnz:
Probably broke due to this work https://github.com/silverstripe/silverstripe-cms/issues/3135

emteknetnz:
[...]

PR #3158, opened 2026-02-10
FIX Restore adding child page via tree context menu

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/3154

Also seems that the RecordType get var wasn't be respected - presumably this has been an issue for a while, so I've fixed that up in this PR

This PR is a bit strange in that it makes the UX of the 6.1 worse that 5.4 and adding a new page via the context menu means you go to the "add a page" screen rather than the page being instantly added as a child of the page it's being added to, however the existing functionality is currently so broken I think this is worthwhile.

I think the intention of the [prior work](https://github.com/silverstripe/silverstripe-cms/pull/3149) on 5.4 was that it would be merged up to 6.1... however that doesn't seem to have gone to plan? If we want re-implement that work it will need to be done as a proper PR separate from this work which is more urgent.


tests/behat/features/create-a-page.feature
```diff
@@ -62,15 +62,6 @@ Feature: Create a page
     Then I should see "Virtual Page" in the "#Form_AddForm_RecordType div.radio.selected" element
     Then the "Virtual Page" checkbox should be checked
 
-  Scenario: I can create a page using the context menu
```
emteknetnz, line 65: This was a poor test that passed even though the functionality was broken. The page was being added to the root node instead of as as a child of "MyPage"

tests/php/Forms/CMSMainAddFormTest.php
```diff
@@ -0,0 +1,95 @@
...
+            $this->assertMatchesRegularExpression(
+                '/<input[^>]*name="ParentModeField"[^>]*value="child"[^>]*checked/i',
+                $html
+            );
+            $this->assertStringContainsString((string)$page->ID, $html);
+        } finally {
+            // This prevents bleed over into other tests. Sending a request with ParentID in the querystring
+            // seems to set a persistent state somewhere that causes other tests to fail where expected
+            // urls end up with ?ParentID=1 appended to them
+            $this->get('admin/pages/add');
```
emteknetnz, line 87: This is a workaround for an existing issue i.e. running the test above will [trigger this issue](https://github.com/silverstripe/silverstripe-cms/actions/runs/21891661080/job/63198662800#step:12:117) even without the changes in CMSMainAddForm.php and CMSMain_TreeView.ss

tests/behat/features/create-a-page.feature
```diff
@@ -62,28 +62,10 @@ Feature: Create a page
...
-    And I click on "Page" in the context menu
-    Then I should see "New Page" in the tree
-    And I should see an edit page form
-
   Scenario: Failed validation during page creation shows the validation error
     Given I add an extension "SilverStripe\CMS\Tests\Behaviour\ValidationFailedAddPageExtension" to the "SilverStripe\CMS\Forms\CMSMainAddForm" class
     And I am logged in as a member of "EDITOR" group
     When I go to "/admin/pages"
-    # First use the context menu approach
-    And I right click on "MyPage" in the tree
```
emteknetnz, line 79: This functionality is being removed in this PR (it never worked in 6.1), it will need to be re-added in a different PR

code/Forms/CMSMainAddForm.php
```diff
@@ -110,6 +110,12 @@ protected function createFields(): FieldList
             ['type' => mb_strtolower($singleton->i18n_singular_name())]
         );
 
+        $selectedModelClass = $controller->getDefaultModelClass();
+        $recordType = $controller->getRequest()->getVar('RecordType');
+        if ($recordType && class_exists($recordType) && is_subclass_of($recordType, $modelClass)) {
+            $selectedModelClass = $recordType;
+        }
```
ChloeSartorelli, line 117: Why is this being added? I don't think the `RecordType` get variable is meant to populate the form like this (but of course that is because the context menu is meant to submit the form, not go to the filled-in form for manual submission)

tests/php/Forms/CMSMainAddFormTest.php
```diff
@@ -0,0 +1,95 @@
...
+            ],
+            'VirtualPage subclass' => [
+                'recordType' => VirtualPage::class,
+                'expectedClass' => 'VirtualPage',
+                'shouldBeChecked' => true,
+            ],
+            'invalid class defaults to Page' => [
+                'recordType' => 'InvalidClass',
+                'expectedClass' => 'Page',
+                'shouldBeChecked' => true,
```
ChloeSartorelli: In all of these scenarios, `shouldBeChecked` is true. Were you going to add scenarios where that should be false?

GuySartorelli (commented):
Is the intention of this PR to simply revert the offending merge-up that broken the context menu, and require that work to be done separately?

If so, either the original issue(s) for that work needs to be reopened, or a new issue needs to be created.

tests/php/Forms/CMSMainAddFormTest.php
emteknetnz: Removed shouldBeChecked

code/Forms/CMSMainAddForm.php
emteknetnz: It's being added to populate the form. Otherwise it look weird that you can select a page type and it doesn't pre-select it i.e. you need to select it a second time.

Even if we get the right click context menu to submit properly so it doesn't reach this page, it still seems fine to keep this code

emteknetnz:
> Is the intention of this PR to simply revert the offending merge-up that broken the context menu, and require that work to be done separately?

Intention is to make the flow of selecting a page from the right context menu behave as I'd expect it to behave like it more or less always has i.e. I go to the 'add new page with options preselected', rather than it's current very broken state - the 'context menu bypasses this page' functionality is effectively new functionality so should not be released in a patch.

I've raised a [new issue](https://github.com/silverstripe/silverstripe-cms/issues/3159)

GuySartorelli:
I'm going to ignore the "make the flow of selecting a page from the right context menu behave as I'd expect it to behave like it more or less always has" part, because before 6.0 it created the page, not bring you to the form. Bringing you to the form is a regression in 6.0.

> the 'context menu bypasses this page' functionality is effectively new functionality so should not be released in a patch.

It isn't. The fix for 5.4 wasn't to add that functionality (which already worked in the base case), it was to fix some edge cases around it.

I'll update the issue you opened.

GuySartorelli (approved):
Works as expected locally

Issue #3159, opened 2026-02-12
Bypass middle screen when create new page via right-click context menu


emteknetnz:
Follow on from https://github.com/silverstripe/silverstripe-cms/pull/3158

In CMS 4 and 5, in the context menu on the site tree, selecting "add new page here" would automatically submit the add page form, creating a page directly under the item you right clicked.

Work in 5.4 in https://github.com/silverstripe/silverstripe-framework/issues/11944 and https://github.com/silverstripe/silverstripe-framework/issues/11943 fixed some edge cases around this functionality and added behat tests.

A [merge-up was attempted to bring these two fixes into 6.1](https://github.com/silverstripe/silverstripe-cms/commit/28612d0b3a30e58365961599e18f8a4ba92e90f1#diff-ed696030b6f3af3c09efb5d0a64649c6c9a7b48d9bd52fe2aea80bae04231c85) but this caused a regression: https://github.com/silverstripe/silverstripe-cms/issues/3154

We should update CMS 6 so that the form is submitted when creating a page via the context menu.

This is a bug fix and can be done in a patch, so long as doing so doesn't break semver.
The two linked issues should also be rechecked to make sure they're resolved in CMS 6.1 as part of fixing this.

## PRs

- https://github.com/silverstripe/silverstripe-cms/pull/3161

GuySartorelli:
I've updated the issue - the main functionality broke in 6.0 so this is a bug fix not a new feature.

PR #3161, opened 2026-02-19
FIX Submit add record form with sitetree context menu

GuySartorelli:
The "add new page here" option in the site tree context menu was never meant to go to the form - it was meant to submit it. We broke this in CMS 6.0, so this PR brings that functionality back.

## Issue

- https://github.com/silverstripe/silverstripe-cms/issues/3159


yarn.lock
```diff
@@ -7534,7 +7534,7 @@ table@^6.9.0:
     string-width "^4.2.3"
     strip-ansi "^6.0.1"
 
-tapable@^2.2.1, tapable@^2.3.0:
+tapable@^2.2.0, tapable@^2.2.1, tapable@^2.3.0:
```
ChloeSartorelli, line 7537: This happened when I ran `yarn install`. Assuming the JS CI job goes green, I'm inclined to keep it, or else it'll just pop up again next time `yarn install` gets run.

tests/behat/features/create-a-page.feature
ChloeSartorelli, line 1: See https://github.com/silverstripe/silverstripe-cms/pull/3158/ - these were originally added in 5.4, but removed when the associated template change was reverted.
Adding them back in here, but more robust (i.e. addressing the issues noted in PR comments in the above linked PR)

templates/SilverStripe/CMS/Controllers/Includes/CMSMain_TreeView.ss
ChloeSartorelli, line 1: Main changes in this template are:

1. `LinkRecordAdd` change to `Link` - This is required so that the request can be forwarded to the form. The URL returned from `LinkRecordAdd()` leads to a string of the rendered form. The URL returned from `Link('AddForm')` leads to the form itself.
2. Bring the `ParentID` and `RecordType` parts of the query string directly into the query string. This is required because `Link()` doesn't accept two arguments like `LinkRecordAdd()` does.
3. Add `ParentModeField` to the query string. This is required for if validation fails - otherwise the radio button at the top of the form is unset. The behat tests validate this.

emteknetnz (approved):

